Issue #39, opened 2019-08-21
Reconciliation: failed retry notification


csviri:

mycrEEpy:
0.3.9 currently has this issue right? Any updates on when a new version of the merged fix will get a release?

csviri:
Hi, this is a feature, that just currently on level of discussion.
(Will make better descriptions for issues) And its just about how to better notify (other than logs), but have also in case some metrics on failed execution. The retries for fails itself are already merged. 

We plan some changes in the the close future, but we hope we can release a version 1.0.0 at most in february.

PR #42, opened 2019-09-04
Reconciliation retry

charlottemach:
Solution for issue #39, turning the EventScheduler into a queue for retrying failed events.


operator-framework/src/main/java/com/github/containersolutions/operator/EventScheduler.java
```diff
@@ -0,0 +1,82 @@
...
+    private Map<String, Pair<Action, CustomResource>> customEventQueue = Collections.synchronizedMap(new HashMap<>());
+
+    private EventDispatcher eventDispatcher;
+
+    private Integer BACKOFF = 5;
+    private Integer DELAY = 0;
+
+    public <R extends CustomResource> EventScheduler(EventDispatcher<R> eventDispatcher) {
+        this.eventDispatcher = eventDispatcher;
+        startRetryingQueue();
```
adam-sandor: Let's have this called by the class that creates the Scheduler explicitly. Creating threads in a constructor is not nice, because it's somewhat unexpected. At least I only expect simple initialization logic to happen in a constructor.

charlottemach: Added it to the operator instead & adapted the tests. 
I can still squash my many commits if require

PR #73, opened 2020-04-21
Return a parameterized version of the CustomResourceOperationsImpl object

JRial95:
Currently, the `getCustomResourceClients(Class<T> crc)` method returns a raw version of the CustomResourceOperationsImpl object, so it is not possible to call the `getItems()` method when you want a list of all your custom resources:
```java
List<MyCR> crs = operator.getCustomResourceClients(MyCR.class)
    .inNamespace("default")
    .list()
    .getItems();
```
This pull request modifies the `getCustomResourceClients(Class<T> crc)` method to return a parameterized version of the CustomResourceOperationsImpl object to make this possible.


csviri:
Cool, thx for PR!

Issue #84, opened 2020-04-28
Create Nicer User Guide 


csviri:
- its probably enough just to add it to the Readme.md for now

frankkoornstra:
It'd be really nice to have a short section on how to make things work with Spring :pray:

csviri:
@frankkoornstra  Yep, will add a nice user manual soon. For you can take a look on the spring boot auto config example:
https://github.com/ContainerSolutions/java-operator-sdk/tree/master/samples/basic/spring-boot

csviri:
Created a skeleton of this, will continuously update it. I close this issue for now.

PR #88, opened 2020-04-29
Add documentation for Spring Boot in Readme

frankkoornstra:
Just a few lines to get people started if they want to use Spring Boot. I don't have a ton of experience with Spring yet so please feel free to correct me, this is the way I got it working but wouldn't be surprised if there's a better way :slightly_smiling_face: 

Contributes to #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,36 @@ public class WebServerSpec {
 }
 ```
 
+### Spring Boot
+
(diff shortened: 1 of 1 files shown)


csviri:
Thank you!! will, merge it, maybe make some minor adjustments later. 
We will try to prepare also a comprehensive user guide soon..

PR #92, opened 2020-05-03
Generation aware scheduling

csviri:
Will process events only if generation attribute is increased from last execution. (Configurable with flag)


operator-framework/src/test/java/com/github/containersolutions/operator/EventSchedulerTest.java
```diff
@@ -131,16 +173,15 @@ public void processesNewEventIfItIsReceivedAfterExecutionInError() {
...
+                .matches(list -> eventExecutedBefore(0, 1),
                         "Start time of event 2 is after end time of event 1");
 
         assertThat(eventProcessingList.get(0).getException()).isNotNull();
         assertThat(eventProcessingList.get(1).getException()).isNull();
     }
 
     @Test
-    public void numberOfRetriesIsLimited() {
+    public void numberOfRetriesCanBeLimited() {
```
adam-sandor, line 184: What does this test do? It's definitely not self-explanatory...

PR #106, opened 2020-05-11
Mysqlsample improvements

adam-sandor:


samples/mysql-schema/k8s/deployment.yaml
```diff
@@ -21,8 +23,8 @@ spec:
       serviceAccount: mysql-schema-operator
       containers:
       - name: operator
-        image: mysql-schema-operator
-        imagePullPolicy: Never
+        image: eu.gcr.io/adamsandor-test/mysql-schema-operator:1.2.3-SNAPSHOT
```
adam-sandor, line 26: this needs to be more generic

PR #119, opened 2020-06-19
Store mysql credentials in k8s secret

mircea-cosbuc:
This patch addresses https://github.com/ContainerSolutions/java-operator-sdk/issues/113 by creating a predictable MySQL user with granted privileges over the created schema. The credentials are stored as a K8s secret in the namespace where the schema is created.


adam-sandor:
LGTM

PR #132, opened 2020-07-04
Subresource support with context

csviri:
- this PR introduces context object to API, which contains the CR client
- this allows to call updates on subresource, see sampe integration test added


operator-framework/src/test/java/com/github/containersolutions/operator/sample/subresource/SubResourceTestCustomResourceController.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Optional<SubResourceTestCustomResource> createOrUpdateResource(SubResourceTestCustomResource resource, Context<SubResourceTestCustomResource> context) {
+        numberOfExecutions.addAndGet(1);
+        log.info("Value: " + resource.getSpec().getValue());
+
+        ensureStatusExists(resource);
+        resource.getStatus().setState(SubResourceTestCustomResourceStatus.State.SUCCESS);
+        context.getCustomResourceClient().updateStatus(resource);
```
adam-sandor, line 35: I don't understand this context-based approach. The controller has to return the CR to tell the framework to update it, but it also has to update it using the context?

csviri: Yes, the call for the subresource is an another request. You cannot update it by updating the resourc or vice versa if marked as subresource in crd.

We could design a nicer api, but by definition you can have more subresource types ( although not really supported this by fabric8 ). So for now this is a quick and general solution.

csviri: By returning the cr we can update anything except subresource. This is by design of k8s.

csviri: What I could think as an alternative, is just return the CR as we do now, and a flag if we should also call the update status. But still we want in any circumstances the context, so we have the CR client available.

csviri:
discussed further and will make the controller interface update also in a new PR

PR #134, opened 2020-07-05
Improved Retry and Sub Resource Support

csviri:
- retry funtionality
- explicit API for status sub-resource handling
- removed support for generatio awareness


operator-framework/src/main/java/com/github/containersolutions/operator/api/UpdateControl.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Optional;
+
+public class UpdateControl<T extends CustomResource> {
+
+    private final T customResource;
+    private final boolean updateStatusSubResource;
+    private final boolean updateCustomResource;
+
+    private UpdateControl(T customResource, boolean updateStatusSubResource, boolean updateCustomResource) {
```
adam-sandor, line 13: very nicely done with the static methods

operator-framework/src/main/java/com/github/containersolutions/operator/processing/EventDispatcher.java
```diff
@@ -100,14 +116,21 @@ private boolean markedForDeletion(CustomResource resource) {
     }
 
     // created to support unit testing
```
sjmittal, line 118: This comment seems bit misleading as this class in invoked from regular flow too.

operator-framework/src/main/java/com/github/containersolutions/operator/processing/EventDispatcher.java
```diff
@@ -75,14 +91,14 @@ private boolean hasDefaultFinalizer(CustomResource resource) {
     }
 
     private void removeDefaultFinalizer(CustomResource resource) {
+        log.debug("Removing finalizer on {}: {}", resource);
```
sjmittal, line 94: If we log entire resource it becomes very verbose. Is it possible to use `trace` whenever we want to log entire resource.

operator-framework/src/main/java/com/github/containersolutions/operator/processing/EventDispatcher.java
csviri: > This comment seems bit misleading as this class in invoked from regular flow too.

But its still true, and this is the only reason we are created this class. This is intentional.

operator-framework/src/main/java/com/github/containersolutions/operator/processing/EventDispatcher.java
csviri: > If we log entire resource it becomes very verbose. Is it possible to use `trace` whenever we want to log entire resource.

You are right, could you pls create an issue to sanitize logs?

csviri:
https://github.com/ContainerSolutions/java-operator-sdk/issues/136
this would be fixed also by this PR

adam-sandor:
integration tests are failing when I run them agains a GKE cluster

PR #195, opened 2020-09-25
ci: JDK 15

sullis:


csviri:
@adam-sandor  @sullis we should probably remove 14.

The principle should be that from Java 8 we support all LTS releases + (?) the latest stable release. Not even sure about stable release.

sullis:
> @adam-sandor @sullis we should probably remove 14.
> 

Done.

csviri:
thank you @sullis !

PR #202, opened 2020-10-06
Add tomcat operator sample

charlottemach:
Add a sample tomcat operator, let me know if the README needs more work.
from https://github.com/charlottemach/jax-demo/pull/1


adam-sandor:
One small missing thing: it should be added to the list of maven modules in the samples pom.xml

csviri:
Really like this sample but not sure we want to have an additional one in the core repo. Can we define a clear purpose with it? 
If not maybe we should just put it into a sperate repo.

adam-sandor:
I'd say the purpose is:
* watching Kubernetes resources created by the Operator
* having multiple Controllers

What do you think does that justify another sample?

TBH I think it would be nicer to have the samples outside of the main repo - it would look nicer to have a repo called tomcat-operator. But for that we'd have to create our own org in Github to have the samples and the sdk next to each other. Which actually might be the thing to do...

charlottemach:
We don't have to add it. We could move them all out or just link to external ones?

adam-sandor:
I would either:
* Move the project under it's own Github org
* Move all the samples under their own repos
Or
* merge this

Which one would you guys go for? @charlottemach @csviri 

PS. I'm for merging it first, then do the big move. Just to make sure the code doesn't hang in there until we complete the move.

csviri:
> I would either:
> 
> * Move the project under it's own Github org
> * Move all the samples under their own repos
>   Or
> * merge this
> 
> Which one would you guys go for? @charlottemach @csviri
> 
> PS. I'm for merging it first, then do the big move. Just to make sure the code doesn't hang in there until we complete the move.

Why we cant just put it in same org?

csviri:
I'm also fine to createa separate org

adam-sandor:
Good quesion... I think putting it into the CS org isn't a good idea because there are so many repos there. I think it's nice if a project + related repos are nicely grouped not together with a 100 other random repos. But maybe that's just me..

adam-sandor:
@charlottemach can you downgrade the language version to 8 (or just remove that from the pom) so it compiles? after that I'll merge it and start working on moving everything to the new org

PR #205, opened 2020-10-17
SchemaController: User PreparedStatements where possible

### LINKED ISSUE #120: MySQL Schema example - JDBC query parameter handling
The parameters in the JDBC queries are prone to injection attack. Let's make sure someone doesn't copy them this way into a production application.

Solution: use PreparedStatements

mureinik:
Use `PreparedStatement`s where possible in the `SchemaController` in order to avoid SQL Injections.

Note that `PreparedSatement`s can only dynamically bind values and not object names, so this technique could only be applied to the queries, and not the DDL statements. The security around these statements can probably be improved by sanitizing the schema values, but it's out of the scope of this PR.

As a side bonus, this PR also uses the try-with-resource idiom when creating these `PreparedStatement`s and `ResultSet`s so they will be properly closed instead of the current implementation that may leak resources.

Closes #120


adam-sandor:
Nicely done @mureinik ! thanks for the PR!

PR #210, opened 2020-10-20
Make things work with Quarkus

adam-sandor:
* Downgrade Fabric8 to version 4.10.3
* Add option to specify doneable class in Controller annotation to prevent aspectj from kicking in

Quarkus has it's wrapper around Fabric8 and prevents another dependency bringing in it's own version. Also there is some version compatibility problem with AspectJ.


adam-sandor:
Don't merge it pls, just take a look if it looks good.

adam-sandor:
Not needed anymore

Issue #57, opened 2020-02-04
Use Controller name as finalizer


adam-sandor:
Use the name of the controller as the finalizer string, but keep it customizable.

ng29:
Hey @adam-sandor  / @charlottemach  
Can i work on this ?? 
Some more info around this will be highly appreciated 👍

csviri:
Hi @ng29 welcome!
currently if you check the master we read the finalizer name from the annotation, where is always a default if not set. 
The idea was, that if its not configured expicitly we don't use the default but the [packag+"."+classNameOfController] with lowercase.

escher-m:
I've added some code based on my understanding of the info. Please let me know if this covers what was intended @csviri

PR #218, opened 2020-10-27
Use Controller name as finalizer

escher-m:
In reference to Issue #57

### DIFF
--- a/operator-framework/src/main/java/com/github/containersolutions/operator/ControllerUtils.java
+++ b/operator-framework/src/main/java/com/github/containersolutions/operator/ControllerUtils.java
@@ -28,7 +28,7 @@ public class ControllerUtils {
         if(getAnnotation(controller).finalizerName().equals(DEFAULT_FINALIZER))
         {
             //[package+"."+classNameOfController]
-            return getAnnotation(controller).customResourceClass.getPackage().getName() + "." + getAnnotation(controller).customResourceClass.getSimpleName();
+            return getAnnotation(controller).getPackage().getName() + "." + getAnnotation(controller).getClass();
         }
         
         return getAnnotation(controller).finalizerName();


operator-framework/src/main/java/com/github/containersolutions/operator/ControllerUtils.java
```diff
@@ -25,6 +25,12 @@
             doneableClassCache = new HashMap<>();
 
     static String getDefaultFinalizer(ResourceController controller) {
+        if(getAnnotation(controller).finalizerName().equals(DEFAULT_FINALIZER))
+        {
+            //[package+"."+classNameOfController]
+            return getAnnotation(controller).customResourceClass.getPackage().getName() + "." + getAnnotation(controller).customResourceClass.getSimpleName();
```
csviri, line 31: This is more like the name of the customResource. Original idea was to use `controller.getClass()`

escher-m:
@csviri I couldn't find any `package` instantiation by searching in the repository. 
So I used code from ControllerUtils.java :
```public static <T extends CustomResource> Class<? extends CustomResourceDoneable<T>>
    getCustomResourceDoneableClass(ResourceController<T> controller) {
        try {
            Class<? extends CustomResource> customResourceClass = getAnnotation(controller).customResourceClass();
            String className = **customResourceClass.getPackage().getName()** + "." + customResourceClass.getSimpleName() + "CustomResourceDoneable";

            if (doneableClassCache.containsKey(customResourceClass)) {
                return (Class<? extends CustomResourceDoneable<T>>) doneableClassCache.get(customResourceClass);
            }
```

csviri:
@escher-m  you receive the controller as the parameter for this exact method, I ment the name of the class of that controller instance should be used. 

Also I can see that checks are failing on this branch.

PR #221, opened 2020-11-13
Fixed deleting on the wrong name

synox:
When deleting, we have to use the same name as used in creating the service


adam-sandor:
good catch!

PR #223, opened 2020-11-16
Controller name as finalizer by default

### LINKED ISSUE #57: Use Controller name as finalizer

psycho-ir:
closes #57 
closes #218

### DIFF
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -6,7 +6,7 @@ public interface ResourceController<R extends CustomResource> {
 
     /**
      * The implementation should delete the associated component(s). Note that this is method is called when an object
-     * is marked for deletion. After its executed the default finalizer is automatically removed by the framework;
+     * is marked for deletion. After its executed the custom resource finalizer is automatically removed by the framework;
      * unless the return value is false - note that this is almost never the case.
      * Its important to have the implementation also idempotent, in the current implementation to cover all edge cases
      * actually will be executed mostly twice.
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -28,4 +28,10 @@ public interface ResourceController<R extends CustomResource> {
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
 
+    // What about we let developer to set the finalizer Name through overriding
+    // this method rather than setting it as attribute in annotation?
+    default String getDefaultFinalizerName() {
+        return this.getClass().getCanonicalName();
+    }
+
 }
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -27,11 +27,4 @@ public interface ResourceController<R extends CustomResource> {
      * <b>However we will always call an update if there is no finalizer on object and its not marked for deletion.</b>
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
-
-    // What about we let developer to set the finalizer Name through overriding
-    // this method rather than setting it as attribute in annotation?
-    default String getDefaultFinalizerName() {
-        return this.getClass().getCanonicalName();
-    }
-
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Set up Minikube
         uses: manusa/actions-setup-minikube@v2.0.1
         with:
-          minikube version: 'v1.15.0'
+          minikube version: 'v1.15.1'
           kubernetes version: ${{ matrix.kubernetes }}
           driver: 'docker'
       - name: Run integration tests
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
@@ -29,7 +29,7 @@ public class ControllerUtils {
         if (!Controller.NULL.equals(annotationFinalizerName)) {
             return annotationFinalizerName;
         }
-        final String crdName = getAnnotation(controller).crdName();
+        final String crdName = getAnnotation(controller).crdName() + "/finalizer";
         return crdName;
     }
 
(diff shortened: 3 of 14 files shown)


operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -28,4 +28,10 @@
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
 
+    // What about we let developer to set the finalizer Name through overriding
+    // this method rather than setting it as attribute in annotation?
+    default String getDefaultFinalizerName() {
```
psycho-ir, line 33: what do you think about setting the finalizer name by overriding this method rather than `Controller.finalizerName` attribute.

psycho-ir, line 33: Several places in the code finalizer references are named `defaultFinalzer`. Any objection to renaming them all to `finalizer`? I suppose they are indeed referring to the finalizer and not the default one. 
e.g. 
Several spots in this file:
https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java#L24

psycho-ir:
Some of Integration tests are failing due to the long finalized name. i.e ->  ```
name part must be no more than 63 characters.
```

We can either truncate the package name or add validation on the `finalizerName` length and ask the developer to override the name a pick a shorter value.

adam-sandor:
How about not putting the package name in the finalizer at all?

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
adam-sandor: I'm not a big fan of overriding methods when we just want to set a string. I like to keep it in the annotation.

adam-sandor: I agree.. sounds like we're calling everything defaultFinalizer even though we just mean the finalizer. This is actually confusing as there a thing called the default finalizer name. @csviri can you confirm we're seeing this right?

psycho-ir:
> How about not putting the package name in the finalizer at all?

It may result in conflict. What I also found risky about using the class name is that refactoring the code may cause unintended different finalizer name.

What about using the `crdName` which is less likely to be changed and if it changes then I think it's OK to have another finalizer name by default. i.e `crdName`.`finalizer`

adam-sandor:
Good thinking, CRD name is better. If that changes than you really need to think about migration strategy anyway. Do we need to add `.finalizer` to the end?

psycho-ir:
> Good thinking, CRD name is better. If that changes than you really need to think about migration strategy anyway. Do we need to add `.finalizer` to the end?

I had a look into a few finalizers and didn't any de facto in naming conventions. 
e.g. `service.kubernetes.io/load-balancer-cleanup` or even `foregroundDeletion`. 
These options looks good to m e: 
- crdName
- crdName/finalizer
- finalizer.crdName

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
psycho-ir: removed this method

psycho-ir: I renamed the default finalizer to finalizer or similar everywhere in docs, logs and code for now. 
LMK if you have any objection/thoughts about it.

psycho-ir:
> > Good thinking, CRD name is better. If that changes than you really need to think about migration strategy anyway. Do we need to add `.finalizer` to the end?
> 
> I had a look into a few finalizers and didn't any de facto in naming conventions.
> e.g. `service.kubernetes.io/load-balancer-cleanup` or even `foregroundDeletion`.
> These options looks good to m e:
> 
> * crdName
> * crdName/finalizer
> * finalizer.crdName

I picked `crdName` option, easy to change though in case you prefer any other convention.

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
csviri: @psycho-ir looks good, yep it jus tfinalizer in most of the cases not default finalizer. 
I don't think we should use default method in the interface however. Would stick to the annotation to override the finalizer name.

psycho-ir: Right, I have removed this method already.

adam-sandor:
I'll give this a test today and merge if it's alright

metacosm:
> > > Good thinking, CRD name is better. If that changes than you really need to think about migration strategy anyway. Do we need to add `.finalizer` to the end?
> > 
> > 
> > I had a look into a few finalizers and didn't any de facto in naming conventions.
> > e.g. `service.kubernetes.io/load-balancer-cleanup` or even `foregroundDeletion`.
> > These options looks good to m e:
> > 
> > * crdName
> > * crdName/finalizer
> > * finalizer.crdName
> 
> I picked `crdName` option, easy to change though in case you prefer any other convention.

This is actually incorrect according to the spec: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers where a custom finalizer is supposed to be `<domain name>/<finalizer name>`…

psycho-ir:
> > > > Good thinking, CRD name is better. If that changes than you really need to think about migration strategy anyway. Do we need to add `.finalizer` to the end?
> > > 
> > > 
> > > I had a look into a few finalizers and didn't any de facto in naming conventions.
> > > e.g. `service.kubernetes.io/load-balancer-cleanup` or even `foregroundDeletion`.
> > > These options looks good to m e:
> > > 
> > > * crdName
> > > * crdName/finalizer
> > > * finalizer.crdName
> > 
> > 
> > I picked `crdName` option, easy to change though in case you prefer any other convention.
> 
> This is actually incorrect according to the spec: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers where a custom finalizer is supposed to be `<domain name>/<finalizer name>`…
> 
> Note also that #229 uses the `addFinalizer` method added on `HasMetadata` in 5.0 which validates that part (and is why the tests are currently failing on that PR).

Good catch! so I'll go with `crdName/finalizer` option as the default name.

psycho-ir:
Added the `/finalizer` suffix to be compliant with the specification.

kirek007:
LGTM!

metacosm:
@psycho-ir do you recall which test failed if the name was too long? The problem with using the CRD name is that, while the group should usually be a valid domain name, it doesn't appear to be mandatory.

Issue #199, opened 2020-09-30
Improve handling of Custom Resource Client


adam-sandor:
There are two problems with controller that want to modify their custom resource objects using the special client Fabric8 provides for these operations (MixedOperation interface).

1. The client gets created in the Operator using a somewhat complicated mechanism, and there is no good way to get hold of the client object in the controller.

2. The type of the client is horrific abomination of generics, so if I do manage to get it into the controller it looks like this: `private MixedOperation<Tomcat, CustomResourceList<Tomcat>, CustomResourceDoneable<Tomcat>, Resource<Tomcat, CustomResourceDoneable<Tomcat>>> tomcatOperations;`

One possible solution is to pass the client to the controller in the context object, which solves both problems. The downside is that controllers get the KubernetesClient passed in their constructor so it looks a bit weird to pass the custom resource client in the context.

adam-sandor:
The workaround for the first problem is to pass the custom resource client to the Controller after it has been registered in the Operator, so the Operator has already created it. Surprisingly the Operator has a getter for these clients.

```
        KubernetesClient client = new DefaultKubernetesClient(config);
        Operator operator = new Operator(client);

        TomcatController tomcatController = new TomcatController(client);
        operator.registerControllerForAllNamespaces(tomcatController);
        tomcatController.setTomcatOperations(operator.getCustomResourceClients(Tomcat.class));
```

metacosm:
Should this be closed now?

PR #228, opened 2020-11-17
WIP: feat: make it possible to externally configure the operator

metacosm:
The operator gets a client based on the configuration.
ResourceControllers now get their client from the operator when they get
registered. Spring Boot starter needs to get adapted.

### DIFF
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -1,7 +1,6 @@
 package io.javaoperatorsdk.operator.api;
 
 import io.fabric8.kubernetes.client.CustomResource;
-import io.fabric8.kubernetes.client.KubernetesClient;
 
 public interface ResourceController<R extends CustomResource> {
 
@@ -28,6 +27,4 @@ public interface ResourceController<R extends CustomResource> {
      * <b>However we will always call an update if there is no finalizer on object and its not marked for deletion.</b>
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
-    
-    void setClient(KubernetesClient client);
 }
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -1,6 +1,7 @@
 package io.javaoperatorsdk.operator.api;
 
 import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.client.KubernetesClient;
 
 public interface ResourceController<R extends CustomResource> {
 
@@ -27,4 +28,6 @@ public interface ResourceController<R extends CustomResource> {
      * <b>However we will always call an update if there is no finalizer on object and its not marked for deletion.</b>
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
+    
+    void setClient(KubernetesClient client);
 }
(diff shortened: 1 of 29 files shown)


operator-framework/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -27,5 +28,6 @@
      * <b>However we will always call an update if there is no finalizer on object and its not marked for deletion.</b>
      */
     UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
-
+    
+    void setClient(KubernetesClient client);
```
psycho-ir, line 32: I would challenge adding a setter method signature into the interfaces in general.
and specifically here IMO
- The usage is not really simpler than getting the client via constructor **if it's really needed**
- The KubernetesClient is an implementation dependency and not really specification, I can imagine cases that developers would like to use any other client or even a controller that doesn't need a KubernetesClient at all. 
- Setting the value via setter makes it unpredictable about when the object state is ready to use.  e.g. what if some reasons developer wants to access the `client` in the constructor or in `InitMethod`?

metacosm: > I would challenge adding a setter method signature into the interfaces in general.

I agree unless that is a part of the expected behavior of the implementations to have that information provided to them.

> and specifically here IMO
> 
>     * The usage is not really simpler than getting the client via constructor **if it's really needed**

Agreed in general. That said, the goal here to be able to configure the client externally to the app once and then retrieve that instance where it's needed. This is still a work in progress. A next step would be to provide automatic injection of the client.

> 
>     * The KubernetesClient is an implementation dependency and not really specification, I can imagine cases that developers would like to use any other client or even a controller that doesn't need a KubernetesClient at all.

This project is a framework built on top of the KubernetesClient with the aim of easing using said client to create operators. I'd argue that most implementations would need a client injected to do their work. If your controller doesn't need the client, why use this framework at all? Do you have examples of cases where it would make sense to use this framework and not require a client instance (test fixture excepted)?

>     * Setting the value via setter makes it unpredictable about when the object state is ready to use.  e.g. what if some reasons developer wants to access the `client` in the constructor or in `InitMethod`?

What's preventing implementations from retrieving the client from the operator instance and use it as they see fit (or even create a separate one if needed)? I do agree that it makes the ready state of the controller a little less clear in some instances but the aim of this work is to streamline the most common use case where the controller will only be used after it has been registered with the operator.

csviri: I agree with @psycho-ir , we don't really get too much adding this to the interface. In an operator we register an instance of a controller. So who implements the operator can prepare this instance as he wants. 
In most of the cases we need the client (except when we manage only non-kuberntetes resources).

metacosm: I agree that using a setter is sub-par. The proper solution would be to be able to use injection so that the client is automatically retrieved where it's needed but I didn't want to go there just yet… 😀

adam-sandor: I get where you're coming from @metacosm I keep having the same thought that this should be more controlled. But then I always have to admit that it's just so straightforward to get the client into the controller in the main method that it doesn't make sense to complicate the framework with this. We do have a serious issue with initializing the custom resource client that happens during controller registration but there is no way to get it into the controller in a nice way. You see that here: https://github.com/java-operator-sdk/tomcat-operator/blob/master/src/main/java/com/example/jaxdemo/TomcatOperator.java 
and this is the issue for it: https://github.com/java-operator-sdk/java-operator-sdk/issues/199

psycho-ir: > I agree that using a setter is sub-par. The proper solution would be to be able to use injection so that the client is automatically retrieved where it's needed but I didn't want to go there just yet… 😀

Given that using a preferred DI by the developer for an operator is a no brainer  task, I doubt whether it makes sense to implement functionality as part of the core or not.

PR #229, opened 2020-11-18
feat: upgrade to k8s client 5.0 to simplify CR handling

metacosm:


adam-sandor:
Wow.. this is awesome, I didn't even think of jumping on this yet and now we have a PR :)

csviri:
Looks good to me!
Just we should take a look why the integration tests are failing.

metacosm:
> Looks good to me!
> Just we should take a look why the integration tests are failing.

Yes, I'm looking into it and have found an issue with the kube client implementation which we are working to fix as well 😄

metacosm:
That said the current IT failures look scary: VM crash! 😱

kirek007:
> That said the current IT failures look scary: VM crash! 😱

Yep, looks like crash comes from kubrenetes-client internals, something related to resource watch feature

kirek007:
I've found that kubernetes client is calling Watch.onClose which is deletaed to our code here:
`java-operator-sdk/operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/EventScheduler.java` line 148.
Removing system.exit(1) line allows to get more understandable error message :)

metacosm:
Yep been looking at it all morning, trying to find a solution… 😥

metacosm:
Ready for review @adam-sandor @csviri

kirek007:
@metacosm thx for PR, I'll going to review it 

It would be great to release this as alpha of our sdk 
@adam-sandor  @csviri.

.github/workflows/ci.yml
```diff
@@ -33,5 +33,5 @@ jobs:
           kubernetes version: ${{ matrix.kubernetes }}
           driver: 'docker'
       - name: Run integration tests
-        run: mvn -B package -P no-unit-tests --file pom.xml
+        run: mvn -e -B package -P no-unit-tests --file pom.xml
```
kirek007, line 36: Do we need -e switch here?

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
```diff
@@ -1,89 +1,39 @@
...
-
-    private final static double JAVA_VERSION = Double.parseDouble(System.getProperty("java.specification.version"));
-    private static final String FINALIZER_NAME_SUFFIX = "/finalizer";
-
-    // this is just to support testing, this way we don't try to create class multiple times in memory with same name.
-    // note that other solution is to add a random string to doneable class name
-    private static Map<Class<? extends CustomResource>, Class<? extends CustomResourceDoneable<? extends CustomResource>>>
-            doneableClassCache = new HashMap<>();
-
+    private static final String FINALIZER_NAME_SUFFIX = ".javaoperatorsdk.io/finalizer";
```
psycho-ir, line 9: Now it should be fine to revert this part, right?

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
```diff
@@ -1,89 +1,39 @@
...
         }
-        final String crdName = getAnnotation(controller).crdName() + FINALIZER_NAME_SUFFIX;
-        return crdName;
+        return getDefaultFinalizerIdentifier(controller);
     }
-
+    
+    static String getDefaultFinalizerIdentifier(ResourceController controller) {
+        final Class<? extends ResourceController> controllerClass = controller.getClass();
+        return controllerClass.getSimpleName().toLowerCase() + FINALIZER_NAME_SUFFIX;
```
psycho-ir, line 21: The class name can be changed due to refactoring, it's risky to use it as part of the finalizer name. i.e Renaming the controller class name without changing the CR name shouldn't need data migration.

operator-framework/src/test/java/io/javaoperatorsdk/operator/SubResourceUpdateIT.java
```diff
@@ -1,36 +1,41 @@
...
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class SubResourceUpdateIT {
-
+    
     private IntegrationTestSupport integrationTestSupport = new IntegrationTestSupport();
-
+    
+    @BeforeAll
+    void checkFinalizer() {
+        assertThat(HasMetadata.FINALIZER_NAME_MATCHER.reset(ControllerUtils.getDefaultFinalizerIdentifier(new SubResourceTestCustomResourceController())).matches()).isTrue();
```
psycho-ir, line 29: What about `HasMetada.isFinalizerValid(String finalizerName): boolean` instead of exposing the matcher object to the user? The usage would be quite simpler and the implementation details (Regex here) wouldn't be exposed.

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
kirek007: This is default fallback if there is no name defined, so even after refactoring developers can introduce own name. I would keep it like that.

psycho-ir: The developer can easily overlook this fact that just because a simple refactoring in the class name they now have to explicitly set the finalized name. 
What do you think is wrong with the current implementation? (`crdName`/`finalizer`)

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
kirek007: :+1:

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
kirek007: Ah, I got your point. Then it would be nice to use crd name and revert change pf `FINALIZER_NAME_SUFFIX` const

psycho-ir: The current implementation of the `getFinalizer` (in the master) does so.

.github/workflows/ci.yml
metacosm: I think it's helpful when the tests fail because of an unexpected exception but it's definitely not needed.

operator-framework/src/test/java/io/javaoperatorsdk/operator/SubResourceUpdateIT.java
metacosm: Good point.

metacosm: Only problem is that this will require a new kube client release…

operator-framework/src/main/java/io/javaoperatorsdk/operator/ControllerUtils.java
metacosm: The issue I have with reverting this change is that, currently, the CR names we have don't really follow the spec for finalizer names… so we're not really setting a good example, which I think we should. Are we OK with renaming the CRDs so that their names provide good examples?

kirek007: Yes, definitely we should update examples to provided best practices.

psycho-ir: > Are we OK with renaming the CRDs so that their names provide good examples?
🙌

kirek007:
I've change merge target to the dedicated branch where we'll track update of kubernetes client. We could release test versions from this branch if needed.

metacosm:
I'm keeping this one opened until I can finish #288 which will supersede this PR.

metacosm:
This has been implemented in #288

PR #234, opened 2020-11-23
Support only java version 11 and above!

### LINKED ISSUE #232: Support only java version 11 and above!
As any projects that are going to use this SDK most probably would be from scratch I would suggest increasing the minimum supported Java version to **11**.

So that we can use new features defined till this version and encourage the teams to use the newer version of the Java!

psycho-ir:
closes #232


csviri (approved):

csviri:
thx @psycho-ir !!

PR #235, opened 2020-11-24
Event sources M1

csviri:
This is the core of the new extensible approach for the framework. Where an user can register and de-register sources of events other than custom resource events. This PR intends to be the first Milestone. But the intention is to merge it into the master, so all the upcoming work is already based on this approach. 

Current state is that it builds, tests are passing. Will work on further tests, and improvements. But please review, also we can have a call and discussion with bugger audience.


csviri:
@psycho-ir @metacosm  Pls take a look if you have time, happy to give you an intro / discuss it, just ping me on discord.

metacosm:
Going to take a look now. Thanks!

metacosm:
@csviri I have to go deeper on the design but how would you provide a different event source because it seems that a lot is currently hardcoded to use the default, CR-based implementation right now. Maybe that's on purpose at this stage of the implementation? It would be interesting to see an alternative implementation to see how things work more concretely, especially because some of the interfaces seem quite tied to the current implementation (`EventSource` and `EventSourceManager` in particular)…

csviri:
> @csviri I have to go deeper on the design but how would you provide a different event source because it seems that a lot is currently hardcoded to use the default, CR-based implementation right now. Maybe that's on purpose at this stage of the implementation? It would be interesting to see an alternative implementation to see how things work more concretely, especially because some of the interfaces seem quite tied to the current implementation (`EventSource` and `EventSourceManager` in particular)…

@metacosm  currently an event (coming from any event source other than custom resource ES ) is always bound to a custom resource, so yes we have 1 special event source for custom resources which is the backbone of the system, and all related maybe user provided event sources are always registered in a controller, where it is aware of the custom resource for which it is releated. 

Don't want to create docs until its finalized, but happy to have a call to explain it in details, and record it maybe.
@psycho-ir @metacosm maybe we should schedule a call for this week to take a look.

Also we will prepare an example.

psycho-ir:
> > @csviri I have to go deeper on the design but how would you provide a different event source because it seems that a lot is currently hardcoded to use the default, CR-based implementation right now. Maybe that's on purpose at this stage of the implementation? It would be interesting to see an alternative implementation to see how things work more concretely, especially because some of the interfaces seem quite tied to the current implementation (`EventSource` and `EventSourceManager` in particular)…
> 
> @metacosm currently an event (coming from any event source other than custom resource ES ) is always bound to a custom resource, so yes we have 1 special event source for custom resources which is the backbone of the system, and all related maybe user provided event sources are always registered in a controller, where it is aware of the custom resource for which it is releated.
> 
> Don't want to create docs until its finalized, but happy to have a call to explain it in details, and record it maybe.
> @psycho-ir @metacosm maybe we should schedule a call for this week to take a look.
> 
> Also we will prepare an example.
> 
> Also created a new text channel on discord: https://discord.gg/VQGMhx7Dda

I'll review the design today to get some context. 
Sure would be great if we could have a call on this topic.

operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -35,51 +36,84 @@ public EventDispatcher(ResourceController controller,
...
-        if (markedForDeletion(resource) && !ControllerUtils.hasGivenFinalizer(resource, resourceFinalizer)) {
-            log.debug("Skipping event dispatching since its marked for deletion but does not have finalizer: {}", event);
-            return;
+    }
+
+    private PostExecutionControl handDispatch(ExecutionScope executionScope) {
+        CustomResource resource = executionScope.getCustomResource();
+        log.debug("Handling events: {} for resource {}", executionScope.getEvents(), resource.getMetadata());
+
+        if (ProcessingUtils.containsCustomResourceDeletedEvent(executionScope.getEvents())) {
```
psycho-ir, line 52: Let's say we have a few update events and then a delete event. 
Isn't it better to split the events into the lists with similar action and dispatch them in the coming order? 

One imaginary case could be:
1. update resource to restore a backup
2. update resource to do some cleanup
3. update resource to do get back and put into an object storage
4. Delete the resource.

If they all fall into the same execution scope, if I understood it correctly the resource will be deleted and the first 3 events will be ignored.

psycho-ir (commented):
Generally speaking, the proposed design looks good to me. 
I added a few comments mostly related to the implementation details. 
I also have this feeling that some of the classes can be turned into `internal` classes and being hidden from the client. 
Something to work on out of the scope of this PR but before the first public release.

operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -0,0 +1,89 @@
...
+    private boolean isWatchAllNamespaces() {
+        return targetNamespaces == null;
+    }
+
+    public void addedToEventManager() {
+        registerWatch();
+    }
+
+    private void registerWatch() {
+        CustomResourceOperationsImpl crClient = (CustomResourceOperationsImpl) client;
```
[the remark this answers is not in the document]
metacosm: It should be possible assuming a more precisely typed `MixedOperation` instance. I will look into it.

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/DeleteControl.java
```diff
@@ -2,25 +2,17 @@
 
 public class DeleteControl {
```
metacosm, line 3: Make that class an enum?

operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -0,0 +1,92 @@
...
+            log.debug("Skipping {} event for custom resource uid: {}, version: {}", action,
+                    getUID(customResource), getVersion(customResource));
+            return;
+        }
+        eventHandler.handleEvent(new CustomResourceEvent(action, customResource, this));
+    }
+
+    @Override
+    public void onClose(KubernetesClientException e) {
+        if (e.getCode() == HTTP_GONE) {
```
psycho-ir, line 82: Based on the Watcher.onclose javadoc, on normal closes the exception is null.
```suggestion
        if (e == null) {
            return;
        }
        if (e.getCode() == HTTP_GONE) {
```

operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
psycho-ir: @csviri pinging because not sure if you saw this comment.

psycho-ir:
Do you think the current design is good enough to be merged into the master? 
Asking because I need the event sourcing with different sources for the operator I would like to work on. would be nice to try it out there to see how it works in practice.

csviri:
> Do you think the current design is good enough to be merged into the master?
> Asking because I need the event sourcing with different sources for the operator I would like to work on. would be nice to try it out there to see how it works in practice.

We started to do testing / updating tomcat example (adam) today. Its kind of a manual testing. 
But I will work on implementing integration tests next week. Hopefully it will be stable / tested next week.
I think we can release it in this form, but we would like to do improvements and add usability features also soon.

Mallington:
> @jfgosselin in version 1.4 the retry is not there yet, currently working on it. Should be ready this or next week.
> Currently what you can do is manage it for yourself, catching all the exceptions in the controller, and scheduling a timed event with TimerEventSource. Or wait few days.

What will the retry mechanism look like when it is released?

csviri:
> > @jfgosselin in version 1.4 the retry is not there yet, currently working on it. Should be ready this or next week.
> > Currently what you can do is manage it for yourself, catching all the exceptions in the controller, and scheduling a timed event with TimerEventSource. Or wait few days.
> 
> What will the retry mechanism look like when it is released?

Pls take a look on the changes on this branch: https://github.com/java-operator-sdk/java-operator-sdk/tree/retry-experiment

Mallington:
> @jfgosselin in version 1.4 the retry is not there yet, currently working on it. Should be ready this or next week.
> Currently what you can do is manage it for yourself, catching all the exceptions in the controller, and scheduling a timed event with TimerEventSource. Or wait few days.

Hi @csviri, could you elaborate a bit more on where these exceptions would be caught and what I would have to reschedule? Many thanks

Mallington:
In addition, I see the retry functionality has just been merged. Are you planning on doing a release?
Kind regards.

csviri:
@Mallington it's released, so you have just to specify a retry when registering a controller.

Issue #237, opened 2020-11-25
Define how controllers / operator should be externally configured


metacosm:
#228 was created as an experiment to see how things could be done but I think we need to formalize how configuration should be handled so that users can configure their controllers / operator without having to modify / rebuild the code.

metacosm:
## Proposal

- Be able to configure things in an `application.properties` or `application.yml` file.
- Configure the client at the operator level
- Name controllers so that they can be configured individually in particular for scope and retry

### Client 

- `io.javaoperatorsdk.kubernetes.client` prefix
- Properties similar to what's already defined in [OperatorProperties](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/spring-boot-starter/src/main/java/io/javaoperatorsdk/operator/springboot/starter/OperatorProperties.java#L9-L12)
- OpenShift status should probably be automatically inferred and not "hardcoded"

### Controllers

- `io.javaoperatorsdk.controller.<controller name>` prefix
- `io.javaoperatorsdk.controller.<controller name>.namespaces`: comma-separated list of target namespaces the controller watches, if no value is provided, current namespace (i.e. the one the client is connected to) is assumed. If the value is (or contains) `ALL_NAMESPACES`, the controller watches all namespaces.
- `io.javaoperatorsdk.controller.<controller name>.retry` prefix: properties similar to what's already defined in [RetryProperties](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/spring-boot-starter/src/main/java/io/javaoperatorsdk/operator/springboot/starter/RetryProperties.java)

psycho-ir:
Thanks, @metacosm, the proposal looks good to me.
Maybe one thing to add, we may need a way to make the operator **cluster-scoped**. 
Adding a new field like: 
`io.javaoperatorsdk.controller.<controller name>.scope = namespace|cluster`  may help in this regard.

metacosm:
> Thanks, @metacosm, the proposal looks good to me.
> Maybe one thing to add, we may need a way to make the operator **cluster-scoped**.
> Adding a new field like:
> `io.javaoperatorsdk.controller.<controller name>.scope = namespace|cluster` may help in this regard.

We could make it more explicit but it should be covered using `io.javaoperatorsdk.controller.<controller name>.namespaces=ALL_NAMESPACES`, no?

psycho-ir:
> > Thanks, @metacosm, the proposal looks good to me.
[...]

PR #238, opened 2020-11-26
Extract controller configuration handling and provide Quarkus extension

metacosm:
The idea is for Quarkus to be able to replace the default implementation by an implementation computed at build time.


psycho-ir:
If I understand it correctly the intention is making the Quarkus native builds possible, right?
If so, I think we can support Quarkus without adding anything additional that is going to be exposed to the developer. otherwise would be great if you could elaborate its advantage.

metacosm:
> If I understand it correctly the intention is making the Quarkus native builds possible, right?

The intent is to be able to replace a lot (all) of what is currently done at runtime (e.g. annotation processing, reflection) by build time processing. 

> If so, I think we can support Quarkus without adding anything additional that is going to be exposed to the developer. otherwise would be great if you could elaborate its advantage.

What do you mean by "exposed to the developer"? As things are, the intention is that things will keep working as they currently are when not using Quarkus but the configuration implementation will be swapped out by Quarkus.

psycho-ir:
> What do you mean by "exposed to the developer"? As things are, the intention is that things will keep working as they currently are when not using Quarkus but the configuration implementation will be swapped out by Quarkus.

The controller interface is having a new method called `getConfiguration()`  as part of the interface. 
I as a developer would rather to not expose to the stuff that is supposed to be handled by the framework and I should have nothing to deal with them. 

I think we can achieve the same things just by giving a hint to the graalVM to keep the Reflection metadata in the image. By doing so we don't need to put the effort into having 2 different implementations.

metacosm:
Note that this change is not Quarkus-specific: if we want to be able to externalize the configuration, we will need to provide different implementations of how the configuration is retrieved/computed. This is also helpful for Spring Boot or other frameworks for example, where each could provide their own way of retrieving the configuration. 

If adding a `getConfiguration` method to the `ResourceController` interface is considered too intrusive, we could consider introducing an SPI that would need to be conformed to by implementation providers.

Another point that I think is worth mentioning: there's a lot more to Quarkus than native compilation. Its JVM mode also provides reduced memory size, faster startup, smaller binary size. I think it's worth making sure that we can support all these aspects and not simply focus on native compilation. In order to be able to support these other benefits, more work is needed, especially when things can be properly computed at build time instead of runtime. Registering things for reflection is quite costly even for native compilation.

kirek007:
HI, I'm all in for replacing whole runtime logic with built time configuration.

metacosm:
So things are getting interesting and I'm making some rather significant changes which I would like input on. The main idea behind these changes is to extract the parts that are done to make it easier to work with native compilation and isolate them in such a way that they can be used when needed but don't impact a future Quarkus extension. The parts in question are all the runtime configuration aspects along with the annotation processor and mapper parts as they would be not only be redundant with what Quarkus would be doing but actually be detrimental.

What do you think @adam-sandor @csviri @psycho-ir @kirek007?

psycho-ir:
> So things are getting interesting and I'm making some rather significant changes which I would like input on. The main idea behind these changes is to extract the parts that are done to make it easier to work with native compilation and isolate them in such a way that they can be used when needed but don't impact a future Quarkus extension. The parts in question are all the runtime configuration aspects along with the annotation processor and mapper parts as they would be not only be redundant with what Quarkus would be doing but actually be detrimental.
> 
> What do you think @adam-sandor @csviri @psycho-ir @kirek007?

Do you think it makes sense to write a one-pager doc about the change? I'm not sure about others but at least it would be very helpful for me and also can be added to the decision log or part of the SDK doc later on.

metacosm:
@psycho-ir the changes are now made in this PR (though I still need to go through the tests to see if everything works as expected). The main idea as previously mentioned: there's a new `runtime-configuration` module that people are need to depend on to be able to use the current behavior with the annotation processing and the controller to CR class mapping. Basically, that module provides a default configuration service which extracts configuration from the `Controller` implementations at runtime or using annotation processing at built time.

Both approaches are discouraged when using Quarkus, which is why this PR makes that behavior optional so that other implementations such as one based on Quarkus, could implement a `ConfigurationService` that resolves the configuration some other way. This PR also adds a layer to be able to use external configuration of Controllers though it doesn't yet work completely as defined in #237.

The new approach is visible in the sample projects where projects now depend on the `runtime-configuration` module which they use by creating their `Operator` using the `DefaultConfigurationService.instance()` method. For Quarkus, the `ConfigurationService` is automatically injected. That's also the case for Spring Boot, though Spring Boot uses the current default configuration mechanism.

Of note, an `api` module was extracted from the `operator-framework` one so that both it and the `runtime-configuration` modules could depend on it without risking dependency cycles, though in the end, I'm not sure that was needed but I think it does make things a little cleaner.

metacosm:
Quarkus extension is not finished yet but I'd like to get some feedback on the rest of the changes.

metacosm:
Seems like the issue with the tests is the fact that the sample classes are currently in the `operator-framework-core` module which doesn't have the annotation processor, resulting in the classes not being registered to be later found.

operator-framework/src/test/java/io/javaoperatorsdk/operator/processing/event/EventListTest.java
```diff
@@ -18,6 +18,6 @@ public void returnsLatestOfEventType() {
         new EventList(
             Arrays.asList(mock(Event.class), new TimerEvent("2", null), event2, mock(Event.class)));
 
-    assertThat(eventList.getLatestOfType(TimerEvent.class).get()).isSameInstanceAs(event2);
+    Assertions.assertThat(eventList.getLatestOfType(TimerEvent.class).get()).isEqualTo(event2);
```
psycho-ir, line 19: IMO the static import makes it cleaner

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
```diff
@@ -0,0 +1,172 @@
...
+                "finalizerName",
+                AnnotationValue::asString,
+                () -> ControllerUtils.getDefaultFinalizerName(crdName)),
+            valueOrDefault(
+                controllerAnnotation,
+                "generationAwareEventProcessing",
+                AnnotationValue::asBoolean,
+                () -> true),
+            valueOrDefault(
+                controllerAnnotation, "isClusterScoped", AnnotationValue::asBoolean, () -> false),
```
psycho-ir, line 150: is `isClusterScoped` used anywhere?

metacosm: It's used when the controller is registered with the Operator to know under which scope to register (namespace vs. cluster).

operator-framework/src/test/java/io/javaoperatorsdk/operator/IntegrationTestSupport.java
```diff
@@ -67,7 +70,7 @@ public void initialize(
                   .withMetadata(new ObjectMetaBuilder().withName(TEST_NAMESPACE).build())
                   .build());
     }
-    operator = new Operator(k8sClient);
+    operator = new Operator(k8sClient, configurationService);
```
psycho-ir, line 73: Seems like the sample codes in the readme has to be updated.

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
metacosm: Or at least, it should be but, indeed, it doesn't appear to be used at the moment. I'd keep it, though, because that's an option that we'd want to enable for controllers. That said, I'd rather fix this issue (and possibly others) into subsequent PRs than keeping this one open any longer. It's a pain to maintain and a huge Damocles' sword over people's head :)

psycho-ir: Can you point me to the code? Couldn't find it in the changeset.

operator-framework-quarkus-extension/runtime/src/main/java/io/javaoperatorsdk/quarkus/extension/QuarkusControllerConfiguration.java
```diff
@@ -0,0 +1,126 @@
...
+    this.name = name;
+    this.crdName = crdName;
+    this.finalizer = finalizer;
+    this.generationAware = generationAware;
+    this.clusterScoped = clusterScoped;
+    this.namespaces = namespaces;
+    this.crClass = crClass;
+    this.doneableClassName = doneableClassName;
+    this.watchAllNamespaces = this.namespaces.contains(WATCH_ALL_NAMESPACES_MARKER);
+    this.retryConfiguration =
```
psycho-ir, line 44: how can the user configure the retry in quarkus projects?

metacosm: I need to add the configuration part, indeed. I will create an issue for it.

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
psycho-ir: > Or at least, it should be but, indeed, it doesn't appear to be used at the moment. I'd keep it, though, because that's an option that we'd want to enable for controllers. That said, I'd rather fix this issue (and possibly others) into subsequent PRs than keeping this one open any longer. It's a pain to maintain and a huge Damocles' sword over people's head :)

As an alternative, it can be removed from the code and be addressed as part of the relevant issue. 
IMHO it Would be good practice to develop a feature that has **definition of done**.

README.md
```diff
@@ -63,14 +63,15 @@ Add [dependency](https://search.maven.org/search?q=a:operator-framework) to your
...
+Main method initializing the Operator and registering a controller.
 
 ```java
 public class Runner {
 
    public static void main(String[] args) {
-       Operator operator = new Operator(new DefaultKubernetesClient());
-       operator.registerController(new WebServerController());
+       Operator operator = new Operator(new DefaultKubernetesClient(),
+           DefaultConfigurationService.instance());
```
psycho-ir, line 73: It's not really easy to understand what DefaultConfigurationService is :(. 
I think there is a way to make better before the final release.

metacosm: The problem is to be able to provide the configuration from the annotation processor in the absence of injection. What would you propose to make this simpler/clearer?

metacosm: `ConfigurationProvider` might be a better name?

psycho-ir: From the user's point of view, there is only one ConfigurationService in the classpath belonging to the imported dependency, so would be best to hide it completely from the user by default. 
I'm fine with the current approach and naming though, can become simpler in next iterations.

psycho-ir (approved):

README.md
metacosm: @psycho-ir the problem is that we cannot put it there by default because then it would be pulled in by the Quarkus implementation as well, which is exactly what I'm trying to avoid 😄

PR #239, opened 2020-11-26
Quarkus provided dependency

psycho-ir:
I made a Quarkus core dependency **provided**, and is working as expected. i.e The dependency won't exist in the classpath unless the client does that. 
I think this changeset suffice to support Quarkus native builds! 
Would be great to hear what you think about this solution and if there are any concerns.
@csviri @kirek007 @adam-sandor @metacosm 

**Update** The scope **provided** didn't wok :(, I borrowed the Annotation (with full package name) to this project. 
I think this should work, how is your feeling about it :D?

### DIFF
--- a/operator-framework/pom.xml
+++ b/operator-framework/pom.xml
@@ -78,33 +78,22 @@
             <version>4.0.3</version>
             <scope>test</scope>
         </dependency>
-
         <dependency>
             <groupId>com.google.testing.compile</groupId>
             <artifactId>compile-testing</artifactId>
             <version>0.19</version>
             <scope>test</scope>
         </dependency>
-
         <dependency>
             <groupId>com.google.auto.service</groupId>
             <artifactId>auto-service</artifactId>
             <version>1.0-rc2</version>
-            <scope>provided</scope>
         </dependency>
-
         <dependency>
             <groupId>com.squareup</groupId>
             <artifactId>javapoet</artifactId>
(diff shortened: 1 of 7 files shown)


kirek007:
Nice try with moving just the annotation class :) 
In my opinion this hack is not needed and it's safer to simply use quarkus core even if it's heavier. Shadowing 3rd party classes is never a good idea :) 
What do you think @csviri?

csviri:
Hehe, I think the best would be if quarkus would separate its API as a dependency :)  
Don't think this is a bad solutions, Its quite smelly to have the whole framework on the classpath. Especially when somebody would use the sdk with spring. 

Probably both ways would work. This is nicer not having the whole quarkus on classpath, but it's less safe since there is a minimal chance that this interface or it's package will change in quarkus.

This is a rare occasion when voting might be a good ide :D

@psycho-ir do I understand correctly that this is still a temp. solution? So at the end we plan to have annotation processor to generate json files to support quarkus without annotation?

psycho-ir:
> @psycho-ir do I understand correctly that this is still a temp. solution? So at the end we plan to have annotation processor to generate json files to support quarkus without annotation?

Yes absolutely, this would just enable the users to build quarks natives images until we come up with a better solution.

csviri:
@adam-sandor  what do you think?

metacosm:
I'm working on a quarkus extension as we speak in #238.

PR #241, opened 2020-11-27
Snapshot release

### LINKED ISSUE #240: SNAPSHOT releases from the master
Would be nice to release snapshots from the **master** branch on every merge, so using the latest unreleased changes on other projects would be more convenient.

psycho-ir:
closes #240

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 name: Release to Maven Central
 on:
   release:
-    types: [created]
+    types: [ released ]
 jobs:
   publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -1,7 +1,7 @@
 name: Release to Maven Central
 on:
   release:
-    types: [created]
+    types: [ released ]
```
psycho-ir, line 4: Releasing on `created` event may be a little bit soon, Doing so on release guarantees that the build will take place just after the release is published.

csviri:
@psycho-ir We already have a ci.yaml, where we run the tests etc in case of the branch push. I think:

- On push to master we should run the tests as in this CI and then release the snapshot. Currently I don't see the tests running   on this new pipeline.  
- On the other hand we should then remove the event on the CI.yaml pipeline, so it just runs on a PR. (Maybe also rename this to PR.yaml or such)

psycho-ir:
> @psycho-ir We already have a ci.yaml, where we run the tests etc in case of the branch push. I think:
> 
> * On push to master we should run the tests as in this CI and then release the snapshot. Currently I don't see the tests running   on this new pipeline.
> * On the other hand we should then remove the event on the CI.yaml pipeline, so it just runs on a PR. (Maybe also rename this to PR.yaml or such)

Makes sense, will update it accordingly.

csviri (approved):
LGTM

kirek007 (approved):
Makes sense.Thanks for work! Everything looks good to me.

PR #242, opened 2020-11-30
* Event becomes non generic

adam-sandor:
* EventList class introduced for convenient helper methods
* Event becomes non-generic


csviri:
Lgtm

PR #246, opened 2020-12-03
Fix resource/class loading in environments with hierarchical Classpath

psycho-ir:
In circumstances that the classpath is not flat the current class/resource loading fails. 
I encountered this problem by running an operator with `quarkus:dev`.


csviri:
Is there an easy way to unit test the problematic case? @psycho-ir

psycho-ir:
> Is there an easy way to unit test the problematic case? @psycho-ir

Good point, not sure if there is any easy way to test it though!
I just tested it manually to run my operator via `quarkus:dev`

PR #248, opened 2020-12-04
Setup code formatter

psycho-ir:
I setup the default google formatter.
Let me know how you like this format.
@metacosm @adam-sandor @csviri @kirek007


metacosm:
Not a big fan of the Google format but what matters most is consistency so I'm fine with it if people like it. 😄

csviri (approved):
LGTM

psycho-ir:
If we are all fine with it, I'll add a check to the pr.yml to ensure that the code is formatted before being merged into the master branch

metacosm:
> If we are all fine with it, I'll add a check to the pr.yml to ensure that the code is formatted before being merged into the master branch

Even better would be to be able to format the code on build or as a pre-commit hook.

psycho-ir:
> > If we are all fine with it, I'll add a check to the pr.yml to ensure that the code is formatted before being merged into the master branch
> 
> Even better would be to be able to format the code on build or as a pre-commit hook.

I've never had a good experience in using pre-commits, they usually make the commits way slow.
Although It's possible to format the code on every build, we would still need the validation to ensure the pushed code changes are actually formatted.

metacosm (approved):
LGTM

csviri (approved):
LGTM

psycho-ir:
Would be good to merge this PR if there is no objection.

metacosm:
@psycho-ir how do we set up our IDEs to use the format automatically?

csviri:
> @psycho-ir how do we set up our IDEs to use the format automatically?

If you use intelliJ there is a plugin:
https://plugins.jetbrains.com/plugin/8527-google-java-format

psycho-ir:
The google-java-format plugin does not optimize with respect to the google format style. 
There are XML schemes for both eclipse and Intellij that you can import as code style and no more plugin would be needed. 
https://github.com/google/styleguide
We can also add them to the codebase and document in **CONTRIBUTING** that we follow these styles, wdyt?

csviri:
yep that sounds good!

metacosm:
On retrospect, allowing this PR to go through at this point was a huge mistake… I'm in a world of pain to rebase my quite consequent PRs and I fear it's going to take days just to fix everything… 😭

adam-sandor:
revert the commits from this PR?

csviri:
I think we already merged things after that. We should be more careful next time with these changes :( 
I also head conflicts just did not have that much changes. But I also have some code with the new format.

csviri:
@metacosm it would not help if you would format your code with the new formatter too?

metacosm:
@csviri that's what I did and I managed to get one branch updated. I need to update the fabric8 client 5.0 branch now… Depending on how it goes, I might just re-apply the changes on top of a new branch 😄

PR #252, opened 2020-12-08
Retry Support

csviri:
This PR introduces the retry support back for new core with event sources.


operator-framework/src/main/java/io/javaoperatorsdk/operator/api/RetryInfo.java
```diff
@@ -2,19 +2,24 @@
 
 public class RetryInfo {
```
psycho-ir, line 3: What about turning it into an interface extended by **RetryExecution** and implemented by **GenericRetryExecution**?
so the execution can be returned as info and no need to instantiate a new RetryInfo on every call.

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/RetryInfo.java
```diff
@@ -2,19 +2,24 @@
...
 
   public RetryInfo(int retryNumber, boolean lastAttempt) {
-    this.retryNumber = retryNumber;
+    this.attemptIndex = retryNumber;
     this.lastAttempt = lastAttempt;
   }
 
-  public int getRetryNumber() {
-    return retryNumber;
+  public int getAttemptIndex() {
```
metacosm, line 13: Wouldn't `attemptCount` be a better name?

metacosm (commented):
I will try to look at things more closely over the week-end.

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/RetryInfo.java
csviri: Hehe, I like to work with Pojos, but will check how that would look like, I change it in case its nice.

operator-framework/src/main/java/io/javaoperatorsdk/operator/api/RetryInfo.java
csviri: Hmm, fine by me, I meant with this that the current execution attempt index is this number.

operator-framework/src/main/java/io/javaoperatorsdk/operator/processing/ExecutionConsumer.java
```diff
@@ -22,7 +22,7 @@
 
   @Override
   public void run() {
-    PostExecutionControl postExecutionControl = eventDispatcher.handleEvent(executionScope);
+    PostExecutionControl postExecutionControl = eventDispatcher.handleExecution(executionScope);
```
psycho-ir, line 25: just loud thinking, wouldn't the code be simpler if we had 2 ExecutionConsumers called 
SimpleConsumer and RetryableConsumer and let the consumer to take care of the retrying rather than the handler?

csviri: @psycho-ir thing is this is closely relates with the events are comming in while the controller is executed, that layer does not have enough info basically now. (Also its just really a wraper around executor).

csviri:
Will make the changes an fix this test next week.

adam-sandor:
How does the user configure the retries?

csviri:
> How does the user configure the retries?

Just as before, when registering the controller.

csviri:
@adam-sandor @metacosm @psycho-ir and ALL.  This is the **MVP** of the Retry. This works in a way that if there is an exception thrown from the controller, we will schedule a timer event to retry the execution. (Also we are putting back the events, which were in the scope of the current execution). The retry can be configured when registering the controller.

Note that there are all kind of situations which can happen, like an event is received while there is already a reatry planned or a new event is received when there is already a max attempt of retry is reached. In all these situations I execute the controller. So basically we are not blocking the execution because of an ongoing retry. This is something that we might discuss and change in the future. I think its not straightforward how this should go. 

Maybe it would be more intuitive behavior that if there is already a scheduled event for retry, we don't execute the controller for new events just for the retry event. But this is up to discussion. 

Pls let me know if you have any questions, we can also go through the impl. Otherwise would like to merge it.

metacosm:
> Note that there are all kind of situations which can happen, like an event is received while there is already a reatry planned or a new event is received when there is already the max attempt reached of a retry. In all these situations we execute the controller. So basically we are not blocking the execution because of an ongoing retry. This is something that we might discuss and change in the future. I think its not straightforward how this should go.

Can you give a more concrete example of what you mean here? What are the different options?

> Maybe it would be more intuitive behavior that if there is already a scheduled event for retry, we don't execute the controller for new events just for the retry event. But this is up to discussion.

The thing that I'm not clear about is what these "new events" are? If they are events from the cluster (i.e. there was a change in the resource state on the cluster), whatever the controller is doing should probably be stopped to try to reconcile the new state instead.

psycho-ir:
> Maybe it would be more intuitive behavior that if there is already a scheduled event for retry, we don't execute the controller for new events just for the retry event. But this is up to discussion.

I would say the current implementation is more reasonable, stop reconciling on new events may lead to stop the world for a while just because of a runtime exception happened. 

The changes LGTM, What IMHO is crucial is documenting this feature and also the event sources via javadoc ASAP.

Mallington:
Looks Great, this is exactly what I was looking for. Are you planning on creating a new release tag with these changes? I plan to uptake them once they are released.
Kind regards.

psycho-ir:
> Looks Great, this is exactly what I was looking for. Are you planning on creating a new release tag with these changes? I plan to uptake them once they are released.
> Kind regards.

You should already be able to play around with these changes by changing your dependency version to `1.4.1-SNAPSHOT`, 
We will do the actual release of **1.5.0** very soon (most probably tomorrow).

jfgosselin:
Hi, 

How/where is the maxElapsedTime  used ? : 

https://github.com/java-operator-sdk/java-operator-sdk/blob/v1.5.0/spring-boot-starter/src/main/java/io/javaoperatorsdk/operator/springboot/starter/RetryProperties.java#L12

csviri:
@jfgosselin this is a leftover, should not be there, will remove it, thx!

PR #255, opened 2020-12-10
Support Parameterized abstract controllers

### LINKED ISSUE #254: The annotation processor does not work for Parameterized abstract controllers.
If the controller has a definition like below the processor doesn't work as expected:
```
public class ConfigResourceController extends BaseResourceController<ConfigSourceResource>

public abstract class BaseResourceController<T extends CustomResource> implements ResourceController<T>
```

psycho-ir:
closes #254


metacosm (approved):
LGTM but maybe add some documentation on the annotation processor logic?

psycho-ir:
> LGTM but maybe add some documentation on the annotation processor logic?

Sure, I'll document what the involved classes are doing.

metacosm:
Seems like there are some changes (styles, contributing, in particular) that are already part of master…

PR #256, opened 2020-12-16
update Tomcat sample to use EventSources

adam-sandor:
Sample demonstrating multiple Controllers and EventSource on Deployments.

### DIFF
--- a/samples/tomcat/README.md
+++ b/samples/tomcat/README.md
@@ -2,6 +2,12 @@
 
 Creates a Tomcat deployment from a Custom Resource, while keeping the WAR separated with another Custom Resource.
 
+This sample demonstrates the following capabilities of the Java Operator SDK:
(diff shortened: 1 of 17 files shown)


csviri:
@adam-sandor I'm not convinced we need to put this back, I would keep it in the separate repo. 
@psycho-ir @metacosm  what do you think?

adam-sandor:
I would do the update here, because that's where all the samples are, then move it to where we agree to put them.

csviri:
@adam-sandor but we already moved this to a separate repo. Don't understand why we want to put this back.

adam-sandor:
It was never removed so I'm not really moving it back. I would like to deprecate the tomcat-operator repo.

csviri:
kk, then its fine. 
However let's discuss what samples to keep here in main repo, since are more samples with no clear purpose. We should have clearly defined goal with each of them, since it takes time to maintain them.

metacosm (approved):
LGTM

PR #263, opened 2020-12-18
simplify operator instantiation

s-soroosh:


metacosm:
I'm not sure about this one, to be honest… It seems to complicate things for not much benefit. Is needing to call to the default configuration service really that big of an issue? Why wouldn't we also hide the `DefaultKubernetesClient` as well?

psycho-ir:
> I'm not sure about this one, to be honest… It seems to complicate things for not much benefit. Is needing to call to the default configuration service really that big of an issue? Why wouldn't we also hide the `DefaultKubernetesClient` as well?

what did you find complicated here? what I did was just a Dependency inversion to ensure the clients are not bound to actual implementation. 

> Why wouldn't we also hide the `DefaultKubernetesClient` as well? 

The StandaloneOperator is dependent on `KubernetesClient`, the user can use different implementations like OpenshiftClient. But ConfigurationService is a thing introduced by us to support different environments, I don't think users needs to be aware of it unless they have advanced feature. i.e. they can use `AbstractOperator`.

adam-sandor:
Can you provide a description of what this change intends to accomplish? By reading the code I don't see the point.

psycho-ir:
> Can you provide a description of what this change intends to accomplish? By reading the code I don't see the point.

Sure, by this changeset the `Operator` becomes an interface which has different implementations for different environments. 

From the user pov in pure-java they can instantiate the operator like this: 
```
       Operator operator = new StandaloneOperator(new DefaultKubernetesClient());
       operator.register(new WebServerController());
```

i.e they don't need to know about the configuration services the SDK has. 
In future, we can have implementations like `MockOperator` or more specific implementation for frameworks where they support native **monitoring**, **health check** and so forth.

adam-sandor:
I'm still missing some good examples of where this change will bring benefits. The current changes in examples are just a change in what class gets instantiated.

psycho-ir:
> I'm still missing some good examples of where this change will bring benefits. The current changes in examples are just a change in what class gets instantiated.

@adam-sandor  It's not just that, the required ctor parameters is also less. 
```
Operator operator = new Operator(new DefaultKubernetesClient(), DefaultConfigurationService.instance());
```

vs 
```
Operator operator = new StandaloneOperator(new DefaultKubernetesClient());
```

PR #267, opened 2020-12-21
Spring boot test module

### LINKED ISSUE #261: unit-test support for spring-boot
In order for unit-tests to work some prerequisites must be met, these should be provided out-of-the-box. Most important is that there is no need for a connection to a real k8s-connection.

Testing support from the fabric8-lib might help getting started https://github.com/fabric8io/kubernetes-client#mocking-kubernetes

psycho-ir:
closes #261

### DIFF
--- a/operator-framework-spring-boot-starter-test/src/main/java/io/javaoperatorsdk/operator/springboot/starter/test/EnableMockOperator.java
+++ b/operator-framework-spring-boot-starter-test/src/main/java/io/javaoperatorsdk/operator/springboot/starter/test/EnableMockOperator.java
@@ -1,12 +1,9 @@
 package io.javaoperatorsdk.operator.springboot.starter.test;
 
-
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import org.springframework.context.annotation.Import;
 
 @Retention(RetentionPolicy.RUNTIME)
 @Import(TestConfiguration.class)
-public @interface EnableMockOperator {
-
-}
+public @interface EnableMockOperator {}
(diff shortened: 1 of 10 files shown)


samples/spring-boot-auto-config/src/test/resources/application.yaml
```diff
@@ -0,0 +1,8 @@
+io.javaoperatorsdk:
```
kirek007, line 1: I don't think this is a good way of defining config tree. 
I would go for `javaoperatorsdk.*` and `javaoperatorsdk.test.*`

operator-framework-spring-boot-starter-test/src/main/java/io/javaoperatorsdk/operator/springboot/starter/test/EnableMockOperator.java
```diff
@@ -0,0 +1,9 @@
+package io.javaoperatorsdk.operator.springboot.starter.test;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import org.springframework.context.annotation.Import;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Import(TestConfiguration.class)
+public @interface EnableMockOperator {}
```
kirek007, line 9: You're not injecting operator, it's more like mocking k8s and client, so maybe it's better to call it `InjectMockedKuberentes Client` ?

kirek007 (approved):
Just a few notes about naming, besides that it's very useful peace of code!

samples/spring-boot-auto-config/src/test/resources/application.yaml
psycho-ir: I completely agree, let's change the naming in another PR as it will affect other parts too.

operator-framework-spring-boot-starter-test/src/main/java/io/javaoperatorsdk/operator/springboot/starter/test/EnableMockOperator.java
psycho-ir: I had named it like this first but then I thought we may want to provide a `MockOperator` as well later on.

PR #277, opened 2021-01-01
use full class name of the controller as name instead of simple name

psycho-ir:
The current default name of the controller does end up with conflicts if the same controller class name is used in more than one package. This PR fixes this issue by using the canonical name of the class instead of the simple class name!

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ public class Runner {
 
    public static void main(String[] args) {
        Operator operator = new Operator(new DefaultKubernetesClient(),
-           DefaultConfigurationService.instance());
+           DefaultConfigurationService.INSTANCE);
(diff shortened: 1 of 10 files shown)


adam-sandor:
It's probably a good idea to switch to using the full name of the Controller. However I don't like that this tiny change is bundled with the much bigger change in how the DefaultConfigurationService singleton is retrieved. Can we make that a separate PR?

psycho-ir:
> It's probably a good idea to switch to using the full name of the Controller. However I don't like that this tiny change is bundled with the much bigger change in how the DefaultConfigurationService singleton is retrieved. Can we make that a separate PR?

Done

adam-sandor:
LGTM

PR #282, opened 2021-01-04
fix: clean up controller default name generation

### LINKED ISSUE #279: Change in default controller naming scheme breaks Quarkus extension
The change in naming scheme introduced in #277 breaks the Quarkus extension because we're not consistently using the same name generation algorithm everywhere it's needed.

metacosm:
Fixes #279

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -6,23 +6,6 @@ import java.util.Locale;
 
 public interface ResourceController<R extends CustomResource> {
 
-  static String getDefaultNameFor(ResourceController controller) {
-    return getDefaultNameFor(controller.getClass());
-  }
-
-  static String getDefaultNameFor(Class<? extends ResourceController> controllerClass) {
-    return getDefaultResourceControllerName(controllerClass.getCanonicalName());
-  }
-
-  static String getDefaultResourceControllerName(String rcControllerClassName) {
-    if (rcControllerClassName.indexOf('.') < 0) {
-      throw new IllegalArgumentException(
-          "Must provide a fully-qualified resource controller class name, was: "
-              + rcControllerClassName);
-    }
-    return rcControllerClassName.toLowerCase(Locale.ROOT);
-  }
-
   /**
    * The implementation should delete the associated component(s). Note that this is method is
    * called when an object is marked for deletion. After its executed the custom resource finalizer
@@ -57,10 +40,12 @@ public interface ResourceController<R extends CustomResource> {
   default void init(EventSourceManager eventSourceManager) {}
 
   default String getName() {
-    final var clazz = getClass();
+    return getNameFor(this);
+  }
 
+  static String getNameFor(Class<? extends ResourceController> controllerClass) {
     // if the controller annotation has a name attribute, use it
-    final var annotation = clazz.getAnnotation(Controller.class);
+    final var annotation = controllerClass.getAnnotation(Controller.class);
     if (annotation != null) {
       final var name = annotation.name();
       if (!Controller.NULL.equals(name)) {
@@ -69,6 +54,27 @@ public interface ResourceController<R extends CustomResource> {
     }
 
     // otherwise, use the lower-cased full class name
-    return getDefaultNameFor(this);
+    return getDefaultNameFor(controllerClass);
+  }
+
+  static String getNameFor(ResourceController controller) {
+    return getNameFor(controller.getClass());
+  }
+
+  static String getDefaultNameFor(ResourceController controller) {
+    return getDefaultNameFor(controller.getClass());
+  }
+
+  static String getDefaultNameFor(Class<? extends ResourceController> controllerClass) {
+    return getDefaultResourceControllerName(controllerClass.getCanonicalName());
+  }
+
+  static String getDefaultResourceControllerName(String rcControllerClassName) {
+    if (rcControllerClassName.indexOf('.') < 0) {
+      throw new IllegalArgumentException(
+          "Must provide a fully-qualified resource controller class name, was: "
+              + rcControllerClassName);
+    }
+    return rcControllerClassName.toLowerCase(Locale.ROOT);
   }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -1,8 +1,8 @@
 package io.javaoperatorsdk.operator.api;
 
 import io.fabric8.kubernetes.client.CustomResource;
+import io.javaoperatorsdk.operator.ControllerUtils;
 import io.javaoperatorsdk.operator.processing.event.EventSourceManager;
-import java.util.Locale;
 
 public interface ResourceController<R extends CustomResource> {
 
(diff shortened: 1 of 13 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -6,6 +6,23 @@
 
 public interface ResourceController<R extends CustomResource> {
 
+  static String getDefaultNameFor(ResourceController controller) {
```
psycho-ir, line 9: Wonder why these methods are needed to be here.  Let's keep the interfaces as clean and relevant as possible! 
I still believe the getName also should be pulled out from this interface, I think it's been missed in the main PR wrt quarkus extension.

metacosm: We don't need to put these methods there, it just that it made sense to me to move them here since they pertain to `ResourceControllers` instead of adding them to a util class that's less discoverable.

metacosm:
The change to use the fully-qualified name of the controller implementations is actually problematic and only works currently because of the fact that names are not being used consistently in all places.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
metacosm: @psycho-ir why do you think the `getName` method should be pulled? It's needed to be able to configure controllers individually.

psycho-ir:
> The change to use the fully-qualified name of the controller implementations is actually problematic and only works currently because of the fact that names are not being used consistently in all places. In particular, using FQNs doesn't work in `application.yml` files as dotted names are interpreted as composite paths instead of simple names. This is why the tests are currently failing with this new version of the code that tries to unify names.
> 
> This means that if we want to be able to configure a controller via a properties file, you need to add a name to the controller which doesn't seem like a good default behavior to me.
> 
> It seems like it would make more sense to require changing the default name as it was previously generated (i.e. using `getSimpleName`) only when there's a conflict so I'd actually be in favor of reverting that change. What do you think @adam-sandor?

if the key has dots, then you just need to wrap the key with brackets, e.g:
```
controllers:
  '[io.javaoperatorsdk.operator.springboot.starter.testcontroller]':
    retry:
      maxAttempts: 3
      initialInterval: 1000
      intervalMultiplier: 1.5
      maxInterval: 50000
      maxElapsedTime: 100000

```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
psycho-ir: It let the user set the name in 2 different ways, @Controller.name & overriding the `getName` method and that's what we should strive to avoid.

metacosm:
Sure but can we expect most people to know that? Personally, I'd rather add a name to my `ResourceController` annotation than have to remember the syntax and have the risk of a typo in the potentially long package name.

metacosm:
> > Sure but can we expect most people to know that? Personally, I'd rather add a name to my `ResourceController` annotation than have to remember the syntax and have the risk of a typo in the potentially long package name.
> 
> we need examples to illustrate how the controller can be configured via config files anyway, if one prefer to have shorter name for controller for sure it can be overridden with `@Controller.name`

Right but my point is that the default behavior should be as simple as possible. Right now, with this change, I feel we're making the default case more complex than it needs to be: how often will people have several controllers with the exact same simple name in their operators?

psycho-ir:
> > > Sure but can we expect most people to know that? Personally, I'd rather add a name to my `ResourceController` annotation than have to remember the syntax and have the risk of a typo in the potentially long package name.
> > 
> > 
> > we need examples to illustrate how the controller can be configured via config files anyway, if one prefer to have shorter name for controller for sure it can be overridden with `@Controller.name`
> 
> Right but my point is that the default behavior should be as simple as possible. Right now, with this change, I feel we're making the default case more complex than it needs to be: how often will people have several controllers with the exact same simple name in their operators?

I can't really say how often that can happen. however, it's already happening in the tests the SDK itself has.
it sounds more like concurrency issues, they don't happen so often but we don't really want even those few times it happens, so we need a way to prevent them and at the same time having a predictable behaviour for the user.

metacosm:
> > Right but my point is that the default behavior should be as simple as possible. Right now, with this change, I feel we're making the default case more complex than it needs to be: how often will people have several controllers with the exact same simple name in their operators?
> 
> I can't really say how often that can happen. however, it's already happening in the tests the SDK itself has.
> it sounds more like concurrency issues, they don't happen so often but we don't really want even those few times it happens, so we need a way to prevent them and at the same time having a predictable behaviour for the user.

Yes but I think that the fix should be to tell people to annotate their controller with `name` in that case, instead of making the default behavior more complex, and, of course, document this… 😄

psycho-ir:
> > > Right but my point is that the default behavior should be as simple as possible. Right now, with this change, I feel we're making the default case more complex than it needs to be: how often will people have several controllers with the exact same simple name in their operators?
> > 
> > 
> > I can't really say how often that can happen. however, it's already happening in the tests the SDK itself has.
> > it sounds more like concurrency issues, they don't happen so often but we don't really want even those few times it happens, so we need a way to prevent them and at the same time having a predictable behaviour for the user.
> 
> Yes but I think that the fix should be to tell people to annotate their controller with `name` in that case, instead of making the default behavior more complex, and, of course, document this… 😄

If we raise compiletime/runtime error in such cases, then it makes sense

metacosm:
> If we raise compiletime/runtime error in such cases, then it makes sense

I agree completely and I was planning on looking at implementing something to that effect. Just wanted to make sure we were on the same page regarding the default behavior first. What do you think @adam-sandor @csviri?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
metacosm: Ah yes, didn't think of that but I assume that someone would need to know what they're doing if they override the default `getName` method… since we provide a default implementation. And if they provide their own implementation, then I guess that's the one that would be used and the annotation would be ignored anyway. I don't think that's something that would be an issue in practice.

psycho-ir: Well, API design is mainly about the things we let users do and the reason behind them. 
The getName method has been implemented for internal use and is effectively final (no reason to be overridden by the client).

Having that in the interface as the method will lead to other questions like why `finalizerName`, `crdName`, `generationAwareEventProcessing` and so on don't have getter method in interface accordingly if there is a getter method with default implementation why @Controller annotation is getting the `name` field too? and several similar questions.

metacosm: The name is not internal: to be able to configure a controller, you need to be able to identify it. That's the name's purpose. While there are no obvious reasons that I can think of to override the default naming scheme, some people might still want to provide their own way to name a controller independently of what we provide (because for example, they're not using annotations and don't want to incur that impact at runtime). As such, it makes sense (at least to me) that it lives on the `ResourceController` interface. 

The other methods, on the other hand, are part of the controller's configuration, which, again, you can't retrieve without first knowing the name of the controller. These values are not part of what is essential to the controller.

psycho-ir: In comment, I was talking about the **getName** and not **name**, ofc **name** is not an internal thing, thus the user can set it via @Controller.name. 

**getName** for sure can be added **later on** (it's backward compatible change) to the interface if it's been decided to get controller essential configuration from anywhere other than the controller annotation, AFAIK it's not currently the case and annotation + its crdName property is required.

metacosm: Fixed.

metacosm:
Ready to merge.

adam-sandor:
LGTM

PR #288, opened 2021-01-07
Upgrade to Fabric8's 5.0 client and Quarkus 1.11.0.Final

metacosm:
Quarkus 1.11.0.Final is the first version to use Fabric8's client 5.0 so
we need to update both at the same time. 

Overview of the changes:
- `Doneable` classes have been removed along with all the involved complexity
- `Controller` annotation has been simplified: the `crdName` field has been removed as that value is computed from the associated custom resource implementation
- Custom Resource implementation classes need to be annotated with `Group` and `Version` annotations so that they can be identified properly. Optionally, they can also be annotated with `Kind` (if the name of the implementation class doesn't match the desired kind) and `Plural` if the plural version cannot be automatically computed (or the default computed version doesn't match your expectations).
- The `CustomResource` class that needs to be extended is now parameterized with spec and status types so you can have an empty default implementation that does what you'd expect. If you don't need a status, using `Void` for the associated type should work.


adam-sandor (commented):
This is really cool

goldmann:
This is pure awesomeness, it simplifies usage a lot! Both in SDK and in the code we need to write to use CustomResources. So big :+1:

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/internal/TimerEventSourceTest.java
```diff
@@ -22,7 +22,7 @@
 
   public static final int INITIAL_DELAY = 50;
   public static final int PERIOD = 50;
-  public static final int TESTING_TIME_SLACK = 20;
+  public static final int TESTING_TIME_SLACK = 40;
```
metacosm, line 25: Doesn't look like it's working… 😞

adam-sandor: Right... maybe this type of test just isn't a good idea. Probably the Github runner is just really slow. Let me take another look.

metacosm: They also fail on my system, which, granted is not the fastest around… 😅

adam-sandor: I'll rewrite those tests so the timer is mocked, that's the only good solution.

metacosm:
Reverting back to draft pending release of Quarkus 1.11.0.Final so that the upgrade will be made to that version before this PR is merged.

metacosm:
Should we merge this, @adam-sandor @psycho-ir ?

psycho-ir:
> Should we merge this, @adam-sandor @psycho-ir ?

Haven’t reviewed this PR yet. Will review it tomorrow

secondsun:
I gave it a small smoke test (crud on my custom resource) and this PR seems to work.

metacosm:
Found an issue thanks to @goldmann so reverting to draft untill I can solve it.

goldmann:
I've tested latest changes with my not-so-trivial codebase and it works great!

adam-sandor:
LGTM

PR #294, opened 2021-01-12
Disable unit tests for TimerEventSource

adam-sandor:
Fixing this will taken a bit more time, so let's unblock verification of new quarkus and fabric8 until that's done


metacosm (approved):
LGTM

PR #303, opened 2021-01-14
Simplify controller registration and unify controller configuration

### LINKED ISSUE #302: Clean up handling of target namespaces
We need to settle how we configure which namespace(s) a controller targets and how we configure this. Right now, there are conflicting versions, in particular when it comes to interpreting what no passed namespaces means. `ControllerConfiguration` assumes that the current namespace is the target if the controller doesn't specify any, using the [`WATCH_ALL_NAMESPACES_MARKER`](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java#L10) to decide whether the controller watches all namespaces, while other parts of the code assume that if target namespaces is empty then that means that the controller should watch all namespaces. 

That historic behavior is problematic, in my opinion, because the default behavior for a controller should be to watch the namespace it is deployed on as it might not have privileges outside of that particular namespace.

We should also remove all the `register` methods on `Operator` apart from [this one](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java#L41) which uses the controller's configuration to appropriately register the controller accordingly.

metacosm:
Fixes #302

### DIFF
--- a/samples/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperator.java
+++ b/samples/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperator.java
@@ -24,7 +24,7 @@ public class MySQLSchemaOperator {
     Config config = new ConfigBuilder().withNamespace(null).build();
     KubernetesClient client = new DefaultKubernetesClient(config);
     Operator operator = new Operator(client, DefaultConfigurationService.instance());
-    operator.registerControllerForAllNamespaces(new SchemaController(client));
+    operator.register(new SchemaController(client));
 
     new FtBasic(new TkFork(new FkRegex("/health", "ALL GOOD!")), 8080).start(Exit.NEVER);
   }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -44,17 +44,7 @@ public class Operator {
       registerController(controller, configuration.watchAllNamespaces(), retry, targetNamespaces);
     }
   }
-
-  public <R extends CustomResource> void registerControllerForAllNamespaces(
-      ResourceController<R> controller, Retry retry) throws OperatorException {
-    registerController(controller, true, retry);
-  }
-
-  public <R extends CustomResource> void registerControllerForAllNamespaces(
-      ResourceController<R> controller) throws OperatorException {
-    registerController(controller, true, null);
-  }
-
+  
   public <R extends CustomResource> void registerController(
       ResourceController<R> controller, Retry retry, String... targetNamespaces)
       throws OperatorException {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -44,16 +44,15 @@ public class Operator {
       registerController(controller, configuration.watchAllNamespaces(), retry, targetNamespaces);
     }
   }
-  
-  public <R extends CustomResource> void registerController(
+
+  // currently only used by IntegrationTestSupport but this should also disappear as by default the
+  // controllers should register into the current namespace the client is connected to which should
+  // work as expected for tests, in particular if we want to start testing using random namespace
+  // names
+  <R extends CustomResource> void registerController(
       ResourceController<R> controller, Retry retry, String... targetNamespaces)
       throws OperatorException {
     registerController(controller, false, retry, targetNamespaces);
-  }
-
-  public <R extends CustomResource> void registerController(
-      ResourceController<R> controller, String... targetNamespaces) throws OperatorException {
-    registerController(controller, false, null, targetNamespaces);
   }
 
   @SuppressWarnings("rawtypes")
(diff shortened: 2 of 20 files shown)


samples/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperator.java
```diff
@@ -24,7 +24,7 @@ public static void main(String[] args) throws IOException {
     Config config = new ConfigBuilder().withNamespace(null).build();
     KubernetesClient client = new DefaultKubernetesClient(config);
     Operator operator = new Operator(client, DefaultConfigurationService.instance());
-    operator.registerControllerForAllNamespaces(new SchemaController(client));
+    operator.register(new SchemaController(client));
```
psycho-ir, line 27: How does user can configure the namespaces in runtime then?
Note that whatever is going to be implemented should work for pure-java operators as well.

metacosm: This is done by using `namespaces = ControllerConfiguration.WATCH_ALL_NAMESPACES_MARKER` on the `Controller` annotation.

psycho-ir: That’s not changale in runtime,
In pure java operators it still should be possible for user to get the namespaces some how and respct them in registration.

metacosm: This should be overridden via externalized configuration.

psycho-ir: Not sure what external configuration for pure-java applications mean!
Isn’t that too opinionated? Also do we have this possibility already?

I’m asking because I’m using the removed methods to configure my controllers already and there may be other users who use them, so we need to offer a reasonable alternative before removing them

metacosm: Isn't that contradictory though? If you use this method, then, by definition, you cannot change it at runtime. In this case, using the annotation field should work just fine, no? See for example: https://github.com/java-operator-sdk/java-operator-sdk/pull/303/files#diff-e0b2d5ca6dfbfa8a591df7661a794ce4bedb24ca5d360ee723c1d76dace47657R27

psycho-ir: This is not the only method removed from the API, I'm talking about the entire change we have in operator methods, 
with a combination of the, it's already possible to enable controller in cherry-picked namespaces or in **all** namespaces together.

metacosm: Could you please detail a scenario where removing the methods and using the annotation instead wouldn't work? 

The other aspect of this is that putting the configuration on the controller implementation makes it easier to reason because the controller explicitly states how it's supposed to work (watching one namespace, all/some namespaces, cluster-scoped) as opposed to possibly registering a controller in a context where it's not supposed to be used (though I guess, the later is also a risk with externalized configuration).

s-soroosh: User 1 wants to run operatopr **OP1** just for namespace **ns1** 
User 2 wants to run operatopr **OP1** for namespaces **ns1** and **ns2** 
User 3 wants to run operatopr **OP1** for **all**  the namespaces 

The operator is in pure-java and is deciding which namespaces are in the interest of the operator via command-line arguments. 

So it either calls `registerControllerForAllNamespaces` method if **ALL** is sent as argument method or call `registerController` with the list of given namespaces. 

Now the question is how the same thing can be achieved after removing these methods?

metacosm: That makes sense. How about a `register(controller, overrideConfiguration)` method? It's probably more complex to implement than keeping the separate methods but the issue is that we should always go through the configuration to register controllers because that also has impacts on the event sources and it's not done consistently everywhere.

s-soroosh: Yep it makes sense to have a configuration class with convenient to use method to let the user change the stuff that are useful to be configured in runtime like: retry, namespaces, cluster scoped,...

metacosm: Ok, I'll get on that then! 😄

metacosm: This is now done.

wtrocki:
+100 to get this merged.

adam-sandor:
Thanks @wtrocki ! let's merge it then

PR #304, opened 2021-01-14
CI improvements

### LINKED ISSUE #298: Consider using GitHub action to format code on push
A GitHub action is available to format the code on push: https://github.com/axel-op/googlejavaformat-action
This would allow us to automatically address format issues instead of failing the build for something sometimes as stupid as a wrong whitespace thus freeing developers to focus more on functionality instead of how their code looks like while maintaining a consistent style across the code base.

### LINKED ISSUE #299: Automate changelogs using conventional commits
For an overview of conventional commits, see: https://www.conventionalcommits.org/en/v1.0.0/
There are lots of tools to automate changelog creation based on the semantic information contained in conventional commits: https://www.conventionalcommits.org/en/v1.0.0/#tooling-for-conventional-commits
One of them is https://github.com/TriPSs/conventional-changelog-action which provide integration with GitHub actions. We should probably also set up a commit message linter to enforce that conventional commit messages are properly used, maybe using something like https://github.com/talos-systems/conform?

metacosm:
- [x] Automatically format code on push instead of failing. Fixes #298 
- [x] Enforce conventional commits so that they can be used for automatic changelog.
- [ ] Set up automatic generation of changelog based on conventional commits. Fixes #299


adam-sandor:
LGTM but I'm wondering if we need to update the contributor doc exlpaining how to use these commits.

metacosm:
Note that this doesn't work at the moment.

iocanel (commented):
Overall looks really nice!

My only question is how do you integrate conventional-chanelog-action with the yaml rules provided in a separate commit.

.github/workflows/release.yml
```diff
@@ -16,6 +16,21 @@ jobs:
         run: mvn versions:set -DnewVersion="${RELEASE_VERSION:1}" versions:commit
         env:
           RELEASE_VERSION: ${{ github.event.release.tag_name }}
+      - name: Conventional Changelog Action
+        id: changelog
+        uses: TriPSs/conventional-changelog-action@v3
+        with:
+          github-token: ${{ secrets.github_token }}
+          output-file: "false"
+      - name: Create Release
```
csviri, line 25: Won't this create a endless loop of releases? :)

metacosm:
Not sure this is still needed, at least, in this form with the new automated release notes support from GitHub… That said, it would still be nice to enforce conventional commit messages and automated formatting as well (though not sure that last one is practical).

PR #305, opened 2021-01-18
fix(watchers): gracefully handle watch disconnection

goldmann:
This is based on top of the https://github.com/java-operator-sdk/java-operator-sdk/tree/client-v5 branch. It's created to show a possible solution for the watcher cleanup.

Basically, the idea is to add a `close()` method for manual cleanup of watchers on the `Operator` class. I did not find an automatic way to solve it in general, ideas welcome.

For Quarkus though, this method is automatically called at the application shutdown, so for Quarkus - **it is** automatic.

Draft!


metacosm:
That's more or less what I had in mind, yes. #303 is also supposed to clean-up the event source registration so that we can more easily track them and, in this particular context, indeed close their associated watch more easily.

metacosm:
Hmm, looks like this got closed when I merged the client v5 branch and subsequently deleted it. Mind re-opening it against `master`, @goldmann?

goldmann:
Done: https://github.com/java-operator-sdk/java-operator-sdk/pull/309

metacosm:
Thank you!

PR #306, opened 2021-01-19
feat(quarkus): add support for externalized configuration

metacosm:


operator-framework-quarkus-extension/runtime/src/main/java/io/javaoperatorsdk/quarkus/extension/ExternalControllerConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+
+@ConfigGroup
+public class ExternalControllerConfiguration {
+
+  /**
+   * An optional list of comma-separated namespace names the controller should watch. If the list
+   * contains {@link ControllerConfiguration#WATCH_ALL_NAMESPACES_MARKER} then the controller will
+   * watch all namespaces.
+   */
+  @ConfigItem public Optional<List<String>> namespaces;
```
adam-sandor, line 17: How about if this list is empty then the controller watches all namespaces? I think watching all namespaces is a sensible default as this will mostly be controlled by RBAC anyway by cluster admins not by configuring the operator.

operator-framework-quarkus-extension/runtime/src/main/java/io/javaoperatorsdk/quarkus/extension/ExternalControllerConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+   * The optional name of the finalizer for the controller. If none is provided, one will be
+   * automatically generated.
+   */
+  @ConfigItem public Optional<String> finalizer;
+
+  /**
+   * Whether the controller should only process events if the associated resource generation has
+   * increased since last reconciliation, otherwise will process all events.
+   */
+  @ConfigItem public Optional<Boolean> generationAware;
```
adam-sandor, line 29: Document the default here

adam-sandor:
I see a that there are tests but I don't see any examples of how external configuration would look like. Am I missing something?

operator-framework-quarkus-extension/runtime/src/main/java/io/javaoperatorsdk/quarkus/extension/ExternalControllerConfiguration.java
metacosm: At the moment the default behavior is to watch the namespace the client is connected to (which should be the namespace the operator is deployed in).

adam-sandor: Ok let's leave it like that for now. We could check what's the default in the operator-sdk and just run with that later.

adam-sandor:
LGTM

operator-framework-quarkus-extension/runtime/src/main/java/io/javaoperatorsdk/quarkus/extension/ExternalControllerConfiguration.java
metacosm: I will address this in #303

PR #309, opened 2021-01-20
fix(watchers): gracefully handle watch disconnection

goldmann:
This shows a possible (but not necessarily elegant) way of cleaning up watcher when the application is shut down. In order to gracefully shut down, the `close()` method should be called manually. For Quarkus based apps - this is done automatically.


s-soroosh:
> This shows a possible (but not necessarily elegant) way of cleaning up watcher when the application is shut down. In order to gracefully shut down, the `close()` method should be called manually. For Quarkus based apps - this is done automatically.

For pure-java baed operators we can use java runtime hooks to stop the hooks shutdown, for Spring-based operators we can do a similar thing as what's done already for Quarkus. 

One issue might be to distinguish the environment and close the watchers in the corresponding way, 
This changeset would help to have platform-dependent changes inside the operator. 
https://github.com/java-operator-sdk/java-operator-sdk/pull/263

goldmann:
@s-soroosh Maybe a generic JVM shutdown hook would do a great job for all of the platforms? I guess it's worth a try. I can take a look at this.

goldmann:
I'll close this for now.

metacosm:
I'm going to look at this in the context of the new extension as well.

PR #320, opened 2021-01-23
Remove finalizer in case of exception

### LINKED ISSUE #310: Controller create/update handling called even if exception when applying finalizer

ppatierno:
This PR should fix #310 removing the finalizer in case of exception when trying to apply it on the resource but operation on the k8s server API fails.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
@@ -151,7 +151,12 @@ public class EventDispatcher {
     log.debug(
         "Adding finalizer for resource: {} version: {}", getUID(resource), getVersion(resource));
     addFinalizerIfNotPresent(resource);
-    replace(resource);
+    try {
+      replace(resource);
+    } catch (RuntimeException e) {
+      removeFinalizer(resource);
+      throw e;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


metacosm:
That's actually the first approach I thought of but I thought that the deeper issue was the fact that things were being caches as if the update had occurred even though they didn't. I'm planning on trying to capture this faulty behavior in a test.

ppatierno:
Yeah the problem is that the code about caching is not reached at all when the failure happens as I wrote in your PR.

PR #324, opened 2021-01-26
feat: initial PoC for a Java to CRD generator

metacosm:
The code is still incomplete, sharing for early feedback.


java-to-crd/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+      <version>5.0.0</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-yaml</artifactId>
+      <version>2.11.2</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
```
iocanel, line 41: You can just add a the `META-INF/services/javax.annotation.processing.Processor` and get rid of the dependency.

iocanel: The less dependencies the better.

java-to-crd/src/main/java/io/javaoperatorsdk/crd/JavaToCrdAnnotationProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+  public synchronized void init(ProcessingEnvironment processingEnv) {
+    env = processingEnv;
+  }
+
+  @Override
+  public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+    try {
+      for (TypeElement annotation : annotations) {
+        Set<? extends Element> annotatedElements = roundEnv.getElementsAnnotatedWith(annotation);
+        annotatedElements.stream().map(e -> (TypeElement) e).forEach(this::generateCRD);
```
iocanel, line 46: The  `generateCRD` method only makes sense for `TypeElement` annotated with `CRD`.
Currently this is only enforced by the value of `@SupportedAnnotationTypes`. As, this is something that will definitely change moving forward it might be a good idea to actually guard it with code.

java-to-crd/pom.xml
metacosm: Yes, I debated this but then it's a compile-time dependency and it has the added advantage of being resilient to refactoring (though IDEs can probably deal with this adequately these days).

java-to-crd/src/main/java/io/javaoperatorsdk/crd/JavaToCrdAnnotationProcessor.java
metacosm: I'll cross that bridge when I get there but yes, this is something to keep in mind. That said, right now, I envision requiring the `CRD` annotation because I'm not sure what the alternative would be (unless using extension of the `CustomResource` class as a criterium?).

metacosm:
This will be addressed in the 5.1 release of the fabric8 kubernetes client.

PR #344, opened 2021-02-18
Delayed registration

FroMage:
Hey @ppatierno is this something that would work for you?

The idea is that you mark which resource controller you want delayed:

```java
@Controller(
    name = TestController.NAME,
    delayRegistrationUntilEvent = TestController.RegisterEvent.class)
public class TestController implements RegistrableController<TestResource> {
  // CDI Event to trigger registration
  public static class RegisterEvent {}
…
}
```

And then it will not be registered on the `Operator` until someone fires this CDI event:

```
  @Inject Event<TestController.RegisterEvent> event;

  @POST
  @Path("register")
  public void registerController() {
    event.fire(new TestController.RegisterEvent());
  }
```

Let me know if that works for you?

### DIFF
--- a/operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
+++ b/operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
@@ -1,7 +1,6 @@
 package io.javaoperatorsdk.quarkus.extension.deployment;
 
 import io.fabric8.kubernetes.client.CustomResource;
-import io.javaoperatorsdk.operator.ControllerUtils;
 import io.javaoperatorsdk.operator.Operator;
 import io.javaoperatorsdk.operator.api.Controller;
 import io.javaoperatorsdk.operator.api.ResourceController;
@@ -36,20 +35,14 @@ import io.quarkus.gizmo.ResultHandle;
 import java.lang.annotation.Annotation;
 import java.util.List;
 import java.util.Optional;
-import java.util.function.Function;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.CDI;
 import javax.inject.Singleton;
-import org.jboss.jandex.AnnotationInstance;
-import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
-import org.jboss.jandex.Type;
-import org.jboss.jandex.Type.Kind;
 import org.jboss.logging.Logger;
 
 class QuarkusExtensionProcessor {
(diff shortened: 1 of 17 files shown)


operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
```diff
@@ -62,6 +73,78 @@ void indexSDKDependencies(
...
+                              mc.invokeVirtualMethod(
+                                  selectMethod,
+                                  cdi1,
+                                  mc.loadClass(Operator.class),
+                                  mc.newArray(Annotation.class, 0));
+                          ResultHandle operator =
+                              mc.checkCast(
+                                  mc.invokeInterfaceMethod(getMethod, operatorInstance),
+                                  Operator.class);
+                          ResultHandle cdi2 = mc.invokeStaticMethod(cdiMethod);
```
metacosm, line 123: Any reason why `CDI.current()` is re-invoked?

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
```diff
@@ -62,6 +73,78 @@ void indexSDKDependencies(
...
+                          ResultHandle resourceInstance =
+                              mc.invokeVirtualMethod(
+                                  selectMethod,
+                                  cdi2,
+                                  mc.loadClass(info.name().toString()),
+                                  mc.newArray(Annotation.class, 0));
```
metacosm, line 129: Can't we reuse the one from line 118?

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
```diff
@@ -62,6 +73,78 @@ void indexSDKDependencies(
         new IndexDependencyBuildItem("io.javaoperatorsdk", "operator-framework-core"));
   }
 
+  @BuildStep
+  void createConfigurationServiceAndOperator(
```
metacosm, line 77:
```suggestion
  void createDelayedRegistrationObserver(
```

metacosm, line 77: Could this be done during the build step as the configuration creation step? If not, why not?

metacosm (changes requested):
Perhaps add some comments, explaining that a bean is created to observe events of the type specified by  `Controller.delayRegistrationUntilEvent` and registers the associated controller at that time? 🤔 Makes me think that we probably need to make sure that `Operator.register` is idempotent in case there are multiple such events and leading to the observer trying to register the controller multiple times…

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
FroMage: Good question.

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
FroMage: I don't think that's advised in general, to reuse varargs arrays. It will surely work in this case, but I doubt javac produces bytecode like this.

operator-framework-quarkus-extension/deployment/src/main/java/io/javaoperatorsdk/quarkus/extension/deployment/QuarkusExtensionProcessor.java
FroMage: It can't, otherwise there are cycles.

FroMage:
Where do you think I should add those docs?

metacosm:
> Where do you think I should add those docs?

Inline comments in the code are fine with me for now. We will probably need to add something to the documentation at some point, though… but that could be taken care of later.

FroMage:
OK, is that enough?

ppatierno:
@FroMage sorry, but where I am supposed to write the logic to execute before the controller registration happens and then saying "ok now you can continue and register the controller". It's not really clear to me. /cc https://github.com/tomncooper

metacosm:
@ppatierno: 
> @FroMage sorry, but where I am supposed to write the logic to execute before the controller registration happens and then saying "ok now you can continue and register the controller". It's not really clear to me. /cc @tomncooper

You could simply annotate your controller with the FQN of a CDI event and check for whatever you need to check for in your operator's `run` method and then fire the event when the condition is reached. The controller wouldn't get registered until the event got fired and you wouldn't need to tell the controller to get registered itself, it would happen automatically. See https://github.com/FroMage/java-operator-sdk/blob/relayed-registration/operator-framework-quarkus-extension/tests/src/main/java/io/javaoperatorsdk/quarkus/it/TestController.java#L11 for how the event is declared and https://github.com/FroMage/java-operator-sdk/blob/relayed-registration/operator-framework-quarkus-extension/tests/src/main/java/io/javaoperatorsdk/quarkus/it/TestOperatorApp.java#L36 for how the event is fired.

csviri:
@ppatierno is there an issue for this PR, could you describe what is  the problem we are solving?

FroMage:
No issue that I know of. The problem is that some resources should not be started until other services become available, because they depend on them. This allows that sort of dependency to be expressed.

csviri:
> No issue that I know of. The problem is that some resources should not be started until other services become available, because they depend on them. This allows that sort of dependency to be expressed.

Can we describe the use case then in more detail? Like an example real life scenario where this would be needed.

metacosm:
@csviri Care to elaborate on the issues you see? I'd rather not wait until the weekly to hash any issues out…

csviri:
> @csviri Care to elaborate on the issues you see? I'd rather not wait until the weekly to hash any issues out…
We can discuss it also on a call, or here.  

So first, would like to see a description of the problem / use case we are solving, with some real life example. 

Do I understand it correctly that we want to register controller on some quarkus internal event? Why we are putting this intro controller annotation why not a separate annoation?

metacosm:
> > @csviri Care to elaborate on the issues you see? I'd rather not wait until the weekly to hash any issues out…
> > We can discuss it also on a call, or here.
> 
> So first, would like to see a description of the problem / use case we are solving, with some real life example.

The problem this attempts to solve is that controllers might actually need to wait on some external conditions before they're able to do their work (e.g. DB, some other Kubernetes resource/app…). Right now, controllers are registered as soon as they're processed which can lead to controllers being in a failed state when starting.

> 
> Do I understand it correctly that we want to register controller on some quarkus internal event? Why we are putting this intro controller annotation why not a separate annoation?

It's not a Quarkus internal event. The Quarkus implementation uses CDI events but there is nothing that prevents alternate implementations to use a different mechanism. Basically, the class name passed in the `Controller` annotation is used as a latch of sort and doesn't imply any specific implementation. A plain java implementation could use a simple listener mechanism (at least, I think) :)

> 
> (Sorry maybe I'm just not aware about the context)

I don't have all the context either, to be honest 😅

csviri:
> The problem this attempts to solve is that controllers might actually need to wait on some external conditions before they're able to do their work (e.g. DB, some other Kubernetes resource/app…). Right now, controllers are registered as soon as they're processed which can lead to controllers being in a failed state when starting.

@metacosm I think this should not be coupled tighly with the controller. Why the controller should solve it's own registration? 

1. This could be done compeltely idependently from the operator SDK, with and event system, thus just call the registerController when its acutally needs to be, but the controller should not be aware this mechanism - neither the SDK.

2. The other solution for this problem is just to register an event source in the controller for example for the "DB", and expose an api of this EventSource where in the controller we can check if the DB is ready. If not just return from the controller. (Of course not emit events from the DB event source until its ready)

Neither of these solution require change in the core or an extension for some corner case mechanism. 

 @adam-sandor  pls take a look also.

metacosm:
> > The problem this attempts to solve is that controllers might actually need to wait on some external conditions before they're able to do their work (e.g. DB, some other Kubernetes resource/app…). Right now, controllers are registered as soon as they're processed which can lead to controllers being in a failed state when starting.
> 
> @metacosm I think this should not be coupled tighly with the controller. Why the controller should solve it's own registration?

The controller has nothing to do in this implementation. It just specifies that the SDK needs to wait for an event of the given class to register the controller. Where else would that information would live?

> 
>     1. This could be done compeltely idependently from the operator SDK, with and event system, thus just call the registerController when its acutally needs to be, but the controller should not be aware this mechanism - neither the SDK.

Again, the controller is not aware of this mechanism. The annotation is just informing the SDK, or more precisely the Quarkus implementation, that it needs to wait before registering the controller. I guess we could extract the functionality from the `Controller` annotation and make it Quarkus-only, if that would be more acceptable.

> 
>     2. The other solution for this problem is just to register an event source in the controller for example for the "DB", and expose an api of this EventSource where in the controller we can check if the DB is ready. If not just return from the controller. (Of course not emit events from the DB event source until its ready)
>

That seems overkill and would require a lot of coding for something that should be handled by the SDK, imo.
 
> Neither of these solution require change in the core or an extension for some corner case mechanism.
> 
> @adam-sandor pls take a look also.

csviri:
> The controller has nothing to do in this implementation. It just specifies that the SDK needs to wait for an event of the given class to register the controller. Where else would that information live?

> Again, the controller is not aware of this mechanism. The annotation is just informing the SDK, or more precisely the Quarkus implementation, that it needs to wait before registering the controller. I guess we could extract the functionality from the Controller annotation and make it Quarkus-only, if that would be more acceptable.

I meant exactly this, that the controller should not be aware even of the event. This can be decouple I guess. So please do. 

> That seems overkill and would require a lot of coding for something that should be handled by the SDK, imo.
You have to code that logic anyway somewhere, to when to register the controller. Would be exacly the same just inside an othe context. 

So again while we don't see a very concrete example of this, should not be part of the core. I see this maybe as a different module or extension. Also feels like very much a corner case for me.

@FroMage coudl you describe your use case?

ppatierno:
When you use the operator SDK in a plain Java application, you can run some other code before calling the `register` controller, right?
It seems not to be possible with the Quarkus extension due to the auto-wiring so, for example, if we want to run some specific code before the operator sdk register the controller and call the init on it.

csviri:
@ppatierno yes in that case its just a matter of a method call, and it's up to you when do you do this.

@metacosm would it be possible to change the quarkus to also have this possibility? 
So we can have this late registration functionality decoupled from the SDK?

metacosm:
> @ppatierno yes in that case its just a matter of a method call, and it's up to you when do you do this.
> 
> @metacosm would it be possible to change the quarkus to also have this possibility?
> So we can have this late registration functionality decoupled from the SDK core?

The whole point of this PR is to make it possible to have delayed registration with Quarkus. The impact on the core is minimal, in my opinion, though I will see if I can minimize it even more.

metacosm:
This feature has been integrated in the newly created Quarkiverse extension.

PR #364, opened 2021-03-09
feat: extract quarkus extension to the Quarkiverse

metacosm:
The extension is now found at `io.quarkiverse.operatorsdk:quarkus-operator-sdk`.


adam-sandor:
Let's do this!

PR #373, opened 2021-03-15
Improve ResourceController

lburgazzoli:
- make deleteResource a default method
- update the doc to reflect the actual code

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
@@ -7,27 +7,32 @@ public interface ResourceController<R extends CustomResource> {
 
   /**
    * The implementation should delete the associated component(s). Note that this is method is
-   * called when an object is marked for deletion. After its executed the custom resource finalizer
-   * is automatically removed by the framework; unless the return value is false - note that this is
-   * almost never the case. Its important to have the implementation also idempotent, in the current
-   * implementation to cover all edge cases actually will be executed mostly twice.
+   * called when an object is marked for deletion. After it's executed the custom resource finalizer
+   * is automatically removed by the framework; unless the return value is {@link
+   * DeleteControl#NO_FINALIZER_REMOVAL} - note that this is almost never the case. It's important
+   * to have the implementation also idempotent, in the current implementation to cover all edge
+   * cases actually will be executed mostly twice.
    *
    * @param resource
-   * @return true - so the finalizer is automatically removed after the call. false if you don't
-   *     want to remove the finalizer. Note that this is ALMOST NEVER the case.
+   * @return {@link DeleteControl#DEFAULT_DELETE} - so the finalizer is automatically removed after
+   *     the call. {@link DeleteControl#NO_FINALIZER_REMOVAL} if you don't want to remove the
+   *     finalizer. Note that this is ALMOST NEVER the case.
    */
-  DeleteControl deleteResource(R resource, Context<R> context);
+  default DeleteControl deleteResource(R resource, Context<R> context) {
+    return DeleteControl.DEFAULT_DELETE;
+  }
 
   /**
    * The implementation of this operation is required to be idempotent. Always use the UpdateControl
    * object to make updates on custom resource if possible. Also always use the custom resource
    * parameter (not the custom resource that might be in the events)
    *
-   * @return The resource is updated in api server if the return value is present within Optional.
-   *     This the common use cases. However in cases, for example the operator is restarted, and we
-   *     don't want to have an update call to k8s api to be made unnecessarily, by returning an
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -8,26 +8,31 @@
   /**
    * The implementation should delete the associated component(s). Note that this is method is
    * called when an object is marked for deletion. After its executed the custom resource finalizer
```
metacosm, line 10:
```suggestion
   * called when an object is marked for deletion. After it's executed the custom resource finalizer
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -8,26 +8,31 @@
...
-   * is automatically removed by the framework; unless the return value is false - note that this is
-   * almost never the case. Its important to have the implementation also idempotent, in the current
-   * implementation to cover all edge cases actually will be executed mostly twice.
+   * is automatically removed by the framework; unless the return value is {@link
+   * DeleteControl#NO_FINALIZER_REMOVAL} - note that this is almost never the case. Its important to
```
metacosm, line 12:
```suggestion
   * DeleteControl#NO_FINALIZER_REMOVAL} - note that this is almost never the case. It's important to
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -8,26 +8,31 @@
...
-   * @return The resource is updated in api server if the return value is present within Optional.
-   *     This the common use cases. However in cases, for example the operator is restarted, and we
-   *     don't want to have an update call to k8s api to be made unnecessarily, by returning an
-   *     empty Optional this update can be skipped. <b>However we will always call an update if
-   *     there is no finalizer on object and its not marked for deletion.</b>
+   * @return The resource is updated in api server if the return value is not {@link
+   *     UpdateControl#noUpdate()}. This the common use cases. However in cases, for example the
+   *     operator is restarted, and we don't want to have an update call to k8s api to be made
+   *     unnecessarily, by returning {@link UpdateControl#noUpdate()} this update can be skipped.
+   *     <b>However we will always call an update if there is no finalizer on object and its not
```
metacosm, line 34:
```suggestion
   *     <b>However we will always call an update if there is no finalizer on object and it's not
```

metacosm:
Can you please use the conventional commit format for the commit message? Thanks in advance!

metacosm (approved):

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -7,27 +7,32 @@
...
    *
    * @param resource
-   * @return true - so the finalizer is automatically removed after the call. false if you don't
-   *     want to remove the finalizer. Note that this is ALMOST NEVER the case.
+   * @return {@link DeleteControl#DEFAULT_DELETE} - so the finalizer is automatically removed after
+   *     the call. {@link DeleteControl#NO_FINALIZER_REMOVAL} if you don't want to remove the
+   *     finalizer. Note that this is ALMOST NEVER the case.
    */
-  DeleteControl deleteResource(R resource, Context<R> context);
+  default DeleteControl deleteResource(R resource, Context<R> context) {
```
csviri, line 21: I approved the PR. LGTM, I think this part might be controversial, thus to not force the users to override this method.
I guess the reasoning here is that the deletion is usually handled using cascading delete.

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
csviri: @adam-sandor what do you think?

adam-sandor: I'm on the fence here, so if @metacosm thinks this is good I'm ok with it. You're right that cascading delete will often be used.

lburgazzoli: the main reason for this change was bases on the observation that most of the operator I saw so far rely on cascading delete so making the method `default` was about to match such observation.

csviri: yeah, it makes sense from that perspectives, we managed lot's of external (not k8s) resources, there it does not. But this interface is small enough and well documented so I think it's fine this way. More like wanted to hear / document the arguments around this. thx!

PR #378, opened 2021-03-16
feat: Controllers can now watch only the namespace they're deployed in

### LINKED ISSUE #372: Controller to watch resources only where it is deployed
Previously it was specified that by default the controller would watch only resources in the namespace it is deployed in. https://github.com/java-operator-sdk/java-operator-sdk/commit/456ea3c5877a79f01a47c00928af08d88ce169bd

I would like to ask for this feature to be implemented.

```
@Controller(namespaces = CURRENT)
or
@Controller(namespaces = ALL_NAMESPACES)
```

metacosm:
This is achieved by using `JOSDK_WATCH_CURRENT` as the only value of the
namespaces configuration for the controller.

Fixes #372

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -58,7 +58,7 @@ public class Operator {
    * Registers the specified controller with this operator.
    *
    * @param controller the controller to register
-   * @param <R> the {@code CustomResource} type associated with the controller
+   * @param <R>        the {@code CustomResource} type associated with the controller
    * @throws OperatorException if a problem occurred during the registration process
    */
   public <R extends CustomResource> void register(ResourceController<R> controller)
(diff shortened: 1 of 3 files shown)


adam-sandor:
Can you describe what the indended business logic will be after this PR? Even reading the issue doesn't give me a good understanding of what the current state is and what are we trying to achieve here.

metacosm:
@adam-sandor the idea is that currently, it's not possible to configure a controller to watch only the namespace it is deployed on. It used to be possible (it was even the default configuration at some point) but since we reverted back to the behavior of watching all namespaces by default, it's not possible anymore. The use case is to be able to automatically watch only the "current" namespace where the operator is deployed in, without having to know its name beforehand. This makes most sense in multi-tenant, dynamically provisioned clusters.

adam-sandor:
Got it, thank for explaining. It makes sense to have this I agree. PR approved.

adam-sandor (approved):

PR #388, opened 2021-03-22
refactor: make AbstractConfigurationService easier to use by sub-classes

metacosm:


adam-sandor:
I'm again missing the explanation for the motivation behind the change.

metacosm:
@adam-sandor it's all in the title 😄 `AsbtractConfigurationService` is an abstract class meant to be extended but as it is, it's not very usable by subclasses because it really doesn't provide ways to interact with its internal structure.

metacosm:
@adam-sandor I'm going to go ahead and merge these changes as I need them for the quarkus extension.

adam-sandor:
sure, go for it :) but do know that I hate abstract classes 😂

metacosm:
Well, the class is not new, I just made it easier to use downstream… and yes, abstract classes certainly have their flaws. 😄

PR #391, opened 2021-03-29
Hide CustomResourceFacade, make EventDispatcher & ExecutionScope parameterized

### LINKED ISSUE #389: Remove CustomResourceFacade
Its only purpose seems to be for testing purpose and we should probably use the mock server provided by the fabric8 client for that purpose: https://github.com/fabric8io/kubernetes-client#mocking-kubernetes

metacosm:
It turns out that completely removing `CustomResourceFacade` is more work than is worth. This PR thus limits its exposure by making it only accessible from the defining package so that it can still be used to control tests but doesn't unnecessarily "pollute" the rest of the code.

Also took the opportunity to make `EventDispatcher` and `ExecutionScope` parameterized so that it's clearer that they work with a single type of resource.

Fixes #389.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -137,9 +137,7 @@ public class Operator {
       }
 
       final var client = k8sClient.customResources(resClass);
-      EventDispatcher dispatcher =
-          new EventDispatcher(
-              controller, finalizer, new EventDispatcher.CustomResourceFacade(client));
+      EventDispatcher dispatcher = new EventDispatcher(controller, finalizer, client);
 
       CustomResourceCache customResourceCache = new CustomResourceCache(objectMapper);
       DefaultEventHandler defaultEventHandler =
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -137,7 +137,7 @@ public class Operator {
       }
 
       final var client = k8sClient.customResources(resClass);
-      EventDispatcher dispatcher = new EventDispatcher(controller, finalizer, client);
+      EventDispatcher<R> dispatcher = new EventDispatcher<>(controller, finalizer, client);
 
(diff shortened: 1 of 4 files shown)


csviri:
@metacosm I think we should also change the samples at least in this repository, to see how this looks on them.

metacosm:
> @metacosm I think we should also change the samples at least in this repository, to see how this looks on them.

Not sure I follow as the classes that were changed are not meant to be accessed by client code?

csviri:
> > @metacosm I think we should also change the samples at least in this repository, to see how this looks on them.
> 
> Not sure I follow as the classes that were changed are not meant to be accessed by client code?
Ahh sorry, Context, but it is already there.

To be honest not think this is needed, I did not put there generics for these classes, since never will be used this way, so putting there generics is more confusing if somebody reading the code. Note that you never instantiate with the generics set for a concrete class. 

Could you phrase what is the benefit of this?

metacosm:
I have to admit that I don't understand your last comment. 

The benefit of using generics is that they make it clear and enforce the fact that a class is meant to only work with a single type, not a mix of different ones. This makes the intent clearer and makes the code easier to check.

csviri:
So originally this was intentional that we did not put there generics. Since it's never instantiated with a class in it. Like you never do in the code EventDispatcher<SampleCustomResource>, that's why we did not put there generics. So it's not confusing from this perspective.

csviri:
But If this is more readable for you I fine with it.

csviri (approved):

metacosm:
> So originally this was intentional that we did not put there generics. Since it's never instantiated with a class in it. Like you never do in the code `EventDispatcher<SampleCustomResource>,` that's why we did not put there generics. So it's not confusing from this perspective.

That indeed wasn't the case before but now it is, see https://github.com/java-operator-sdk/java-operator-sdk/pull/391/files#diff-796c0b8116ea0feea6e7d5a6c922d0b308f2629647c16a72bc7508f2da40984aR140. It also makes sense because the `EventDispatcher` is tied to the controller so we should enforce that it only deals with the specific custom resource that the controller deals with. It's not simply a matter of readability, it's also a matter of correctness and intent. It's not so obvious when generics are not used that an `EventDispatcher` only deals with one kind of custom resource.

PR #407, opened 2021-04-21
Enhance event handling

### LINKED ISSUE #368: Add lifecycle hooks to EventSource
The class `io.javaoperatorsdk.operator.processing.event.EventSource` does not provide any lifecycle method, so there is a chance a watch is registered and an event is received before an EventHandler is registered. 

As example, the [memcached example](https://github.com/java-operator-sdk/memcached-operator-sample/blob/main/src/main/java/io/javaoperatorsdk/sample/memcached/DeploymentEventSource.java#L22-L27) register a watcher before registering the EventSource.

I think it would be nice to have some sort of start/stop hook the implementors can use to safely start watch.

lburgazzoli:
This PR includes:
- add lifecycle hooks to EventSource (#368)
- review DefaultEventSourceManager instance variable modifiers
- make DefaultEventSourceManager::registerEventSource final as it is invoked by the constructor
- DefaultEventSourceManager use containsKey instead of null check to determine if an event source is already registerd
- add lifecycle hooks to the operator so i.e. watchers can be properly closed

Fixes #368

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -18,9 +18,8 @@ public class DefaultEventSourceManager implements EventSourceManager {
   private static final Logger log = LoggerFactory.getLogger(DefaultEventSourceManager.class);
 
   private final ReentrantLock lock = new ReentrantLock();
-  private Map<String, EventSource> eventSources = new ConcurrentHashMap<>();
-  private CustomResourceEventSource customResourceEventSource;
-  private DefaultEventHandler defaultEventHandler;
+  private final Map<String, EventSource> eventSources = new ConcurrentHashMap<>();
+  private final DefaultEventHandler defaultEventHandler;
   private TimerEventSource retryTimerEventSource;
 
   public DefaultEventSourceManager(DefaultEventHandler defaultEventHandler, boolean supportRetry) {
@@ -33,8 +32,7 @@ public class DefaultEventSourceManager implements EventSourceManager {
 
   public void registerCustomResourceEventSource(
       CustomResourceEventSource customResourceEventSource) {
-    this.customResourceEventSource = customResourceEventSource;
-    this.customResourceEventSource.addedToEventManager();
+    customResourceEventSource.addedToEventManager();
   }
 
   @Override
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -5,6 +5,7 @@ import io.javaoperatorsdk.operator.processing.event.internal.CustomResourceEvent
 import io.javaoperatorsdk.operator.processing.event.internal.TimerEventSource;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
@@ -37,7 +38,9 @@ public class DefaultEventSourceManager implements EventSourceManager {
   }
 
   @Override
-  public <T extends EventSource> void registerEventSource(String name, T eventSource) {
+  public void registerEventSource(String name, EventSource eventSource) {
+    Objects.requireNonNull(eventSource, "EventSource must not be null");
+
     try {
       lock.lock();
       EventSource currentEventSource = eventSources.get(name);
@@ -47,6 +50,22 @@ public class DefaultEventSourceManager implements EventSourceManager {
       }
       eventSources.put(name, eventSource);
       eventSource.setEventHandler(defaultEventHandler);
+      eventSource.start();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public Optional<EventSource> deRegisterEventSource(String name) {
+    try {
+      lock.lock();
+      EventSource currentEventSource = eventSources.remove(name);
+      if (currentEventSource != null) {
+        currentEventSource.close();
+      }
+
+      return Optional.ofNullable(currentEventSource);
     } finally {
       lock.unlock();
     }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -36,7 +36,7 @@ public class DefaultEventSourceManager implements EventSourceManager {
   }
 
   @Override
-  public void registerEventSource(String name, EventSource eventSource) {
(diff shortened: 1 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -155,7 +180,8 @@ public void onClose(WatcherException e) {
     if (e.isHttpGone()) {
       log.warn("Received error for watch, will try to reconnect.", e);
       try {
-        registerWatch();
+        close();
```
metacosm, line 183: Should we really close all the watches? I guess there's no other option since we can't really know which watcher caused the exception…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -30,23 +29,53 @@ public DefaultEventSourceManager(DefaultEventHandler defaultEventHandler, boolea
...
+
+      eventSources.clear();
+    } finally {
+      lock.unlock();
+    }
   }
 
   @Override
-  public <T extends EventSource> void registerEventSource(String name, T eventSource) {
+  public final void registerEventSource(String name, EventSource eventSource) {
```
metacosm, line 52: Why did you make this method final?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
lburgazzoli: yes, the issue was that in case of an error on of a single watcher, calling `registerWatch` would re-register all the watcher, potentially you could end up in duplicated watchers.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
lburgazzoli: because it is invoked by the constructor so either the class is marked as final or this method need to be final to avoid issues in classes extending `DefaultEventSourceManager` and overriding `registerEventSource`

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: That was another thing I wondered about (using a List for the watchers as opposed to a Set but I'm not sure we have  consistent `equals/hashCode` implementations for the `Watch` implementations)…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
metacosm: Ah yes. I remember hearing about this but never ran into such issues in practice… Have you?

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
lburgazzoli: yes we need a way to improve how watchers are handled, I may try to have a look once this is done

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
lburgazzoli: yes, got bitten hard by it 😂 that's why I notice them

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -17,9 +17,8 @@
   private static final Logger log = LoggerFactory.getLogger(DefaultEventSourceManager.class);
 
   private final ReentrantLock lock = new ReentrantLock();
-  private Map<String, EventSource> eventSources = new ConcurrentHashMap<>();
-  private CustomResourceEventSource customResourceEventSource;
```
csviri, line 21: Do you see it correctly that this has been moved to the Operator?  Why is that?
The intention was that this class will own all the custom resources related to a controller.
But are there particular reasons to move it? 
The downside is that if we want to have other events,between event sources and controllers vs event source, but also access access CustomResourceEventSource (where we thinking to move the caching for example) will not be accessible in this manager.

lburgazzoli: The reason was that the method:

```
public void registerCustomResourceEventSource(
      CustomResourceEventSource customResourceEventSource) {
    this.customResourceEventSource = customResourceEventSource;
    this.customResourceEventSource.addedToEventManager();
}
```

Was not really meaningful as yes the `customResourceEventSource` was assigned to an instance of the DefaultEventManager but was never accesses by the manager. A better option would have been either to use `registerEventSource` as any other event source or pass the `customResourceEventSource` as a constructor parameter to make it clear that the `DefaultEventManager` requires it whereas as today, it looks like an unrelated method.

csviri-form3: I see, ok, yes this part definetelly needs improvements and refactorings. 
As I mentioned I would like to keep the event sources in one place, I think the constructor would make sense. 
CustomResourceEventSource currently has quite special position, it's required always as an event source, so constructor is good ide also do express this.

csviri: I see, ok, yes this part definetelly needs improvements and refactorings. (Just like the others :) )
As I mentioned I would like to keep the event sources in one place, I think the constructor would make sense.
CustomResourceEventSource currently has quite special position, it's required always as an event source, so constructor is good ide also do express this.

metacosm: Note that I will probably make some other improvements after this, getting back to work on #392

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
lburgazzoli: @csviri yep I think I got the reasoning and I do agree but for the sake of the lifecycle it was quite confusing so I opted to make at little bit more clear (i.e to have set-up / tear-down on the same side, in this case on the Operator). If we move it on the contructor then we can move its set-up/tear-down as part of the event manager.

lburgazzoli:
@csviri @metacosm thx for the review

metacosm:
Thank you!

PR #416, opened 2021-05-04
chore: remove Spring Boot starters

### LINKED ISSUE #414: Extract Spring Boot starter and test starter into their own project
This move will follow the similar one that was made a while ago to separate the quarkus extension from the core project. The issue here is that this will allow us to focus on the core and let the starter evolve at a different pace. 
On a related note, we really need people to maintain the Spring Boot starter as it has a little fallen behind.

metacosm:
The starters have been moved to their own repositories in the
java-operator-sdk organization.
Fixes #414


adam-sandor (approved):
LGTM

attilameszaros-form3 (approved):

csviri (approved):

PR #419, opened 2021-05-06
feat: simplify wiring of event sources

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -107,14 +91,12 @@ public void close() {
...
   @Override
-  public void eventReceived(Watcher.Action action, CustomResource customResource) {
+  public void eventReceived(Watcher.Action action, T customResource) {
     log.debug(
         "Event received for action: {}, resource: {}",
         action.name(),
         customResource.getMetadata().getName());
 
-    resourceCache.cacheResource(
-        customResource); // always store the latest event. Outside the sync block is intentional.
```
csviri, line 117: I see you moved this to the DefaultEventHandler. That is not good. Since the custom resource does not gets to the cache if the generations are not changed, this can be a problem in case of a retry for example.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -59,27 +74,14 @@ public DefaultEventHandler(
...
-  }
-
   @Override
   public void close() {
-    if (eventSourceManager != null) {
-      log.debug("Closing EventSourceManager {} -> {}", controllerName, eventSourceManager);
-      eventSourceManager.close();
+    try {
+      log.debug("Closing handler for {}", controllerName);
+      executor.awaitTermination(10, TimeUnit.SECONDS);
```
csviri, line 81: This should be configurable by the user (with a default). Since we want to gracefully shutdown, so wait ideally all threads are finished.  There can be a meaningful timeout, but that depends on the implementation of the controller.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -135,28 +117,25 @@ public void eventReceived(Watcher.Action action, CustomResource customResource)
...
     }
-    if (!largerGenerationThenProcessedBefore(customResource)) {
-      return true;
-    }
-    return false;
+    return !hasGenerationAlreadyBeenProcessed(customResource);
   }
 
-  public boolean largerGenerationThenProcessedBefore(CustomResource resource) {
+  private boolean hasGenerationAlreadyBeenProcessed(T resource) {
```
csviri, line 138: Actually the I would keep the old one because:
1. that is what we actually doing (comparing if its larger)
2. In general this should be linear increasing number because of the strong consistency guarantees of etcd. Thus we can expect that that we receive these events in order. But checking for a large number makes it a nice defensive programming practice.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/retry/RetryTestCustomResourceController.java
```diff
@@ -44,7 +37,7 @@ public DeleteControl deleteResource(
     if (numberOfExecutions.get() < NUMBER_FAILED_EXECUTIONS + 1) {
       throw new RuntimeException("Testing Retry");
     }
-    if (context.getRetryInfo().isEmpty() || context.getRetryInfo().get().isLastAttempt() == true) {
```
csviri, line 47: Hehe, nice catch!

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/subresource/SubResourceTestCustomResourceController.java
```diff
@@ -23,12 +22,6 @@
       LoggerFactory.getLogger(SubResourceTestCustomResourceController.class);
   private final AtomicInteger numberOfExecutions = new AtomicInteger(0);
 
-  @Override
-  public DeleteControl deleteResource(
```
csviri, line 27: The default method for delete was a good idea!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -87,9 +81,7 @@ private PostExecutionControl handleDispatch(ExecutionScope<R> executionScope) {
     }
     Context<R> context =
         new DefaultContext<>(
-            eventSourceManager,
-            new EventList(executionScope.getEvents()),
-            executionScope.getRetryInfo());
+            new EventList(executionScope.getEvents()), executionScope.getRetryInfo());
```
csviri, line 84: What are these, formatting changes?

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/retry/RetryTestCustomResourceController.java
csviri, line 47: hehe, nice catch! :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: The tests are passing, though…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
metacosm: Agreed. I still have other changes I want to make but I want first to get that PR merged so that we can proceed step by step.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
attilameszaros-form3: Probably we don't have a test for this case.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
attilameszaros-form3: this is quite low level code so it's not bad it exposes the intention. And hasGenerationAlreadyBeenProcessed is clearly misleading, it can happen that the generation is not processed before, like when operator was down.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
attilameszaros-form3: Could we move this change to that PR?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
csviri: Probably we don't have a test for this case.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
attilameszaros-form3: this is quite low level code so it's not bad it exposes the intention. And hasGenerationAlreadyBeenProcessed is clearly misleading, it can happen that the generation is not processed before, like when operator was down.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
csviri: kk, that is fine by me, until we make sure we don't forget about it :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: Yes but the intent of the method is still to figure out whether the generation of the provided `CustomResource` has been seen/processed already.  
It also doesn't matter if the method if low level or not. People still need to read the code and figure out what its intent is. An alternative would be to remove that method altogether and merge it in the calling code (since it's only called in one spot).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
metacosm: https://github.com/java-operator-sdk/java-operator-sdk/issues/421

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -88,6 +90,12 @@ public void setEventSourceManager(DefaultEventSourceManager eventSourceManager)
 
   @Override
   public void handleEvent(Event event) {
+    // cache the latest version of the CR
```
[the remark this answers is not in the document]
metacosm: I honestly don't like this way of doing things. The behavior shouldn't depend on whether or not we have a cache or not. That's a bad design.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
metacosm: And actually, it's not a formatting change. The `eventSourceManager` parameter got removed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
csviri: I don't like these long methods, but I think the naming pretty much should reflect what is happening in the implementation, if its higher level abstraction is better, but should not be confusing. If sombody reads this, and checks the implementation, could think if the generation is higher all the previouse generations where processed, what is cleary not the case.

note that this method already just called from a higher abstraction the "skipBecauseOfGeneration". So when I read it:
ok we are skipping when this generation is not larger then we already seen before. So its reads quite well as it is now.

csviri: But clearly the implementation has nothing to do with processing or not, its about just the generation increased.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
attilameszaros-form3: ahh ok, then its good! thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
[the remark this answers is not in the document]
metacosm: Sure, but it doesn't change the fact that the behavior shouldn't depend on whether or not you optimize things by having a cache. A cache is an optimization, it's not a functional requirement and currently, it acts like one. This is the part that I don't like. The optimization should be removable and the code should still work when it is removed (slower, but still work). Right now, if you remove the "optimization" things don't work at all.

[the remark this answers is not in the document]
csviri: "Optimizations should be removable." - Why is this a requirement? We do optimizations not because we want to remove them.

It's already there, the whole process it's much more efficient this way. Making a K8S API request on every event, could be quite an issue. So does not makes sense to change this because, Especially when the event comes from an event source. 
So feel free that makes it nicer / more readable, if you think this is not that nice, this way. But without breaking the functionality.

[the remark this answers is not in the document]
csviri: Checking the operator sdk now, and this should work exactly the same. So the informer provides a cache about the objects.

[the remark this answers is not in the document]
metacosm: I'm not saying the optimization should be removed. I'm saying it should be removable without impacting the functionality, otherwise, it's not an optimization. Strictly speaking, there is no need for a cache, so the functionality should work even when there is no cache.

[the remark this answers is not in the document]
csviri: Imho the best would be if you would create a separate issue, why is an un optimized version is needed. So we can discuss the pros and cons and all the arguments behind it.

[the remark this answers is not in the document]
metacosm: An unoptimized version is not needed, strictly speaking. But the algorithm should work without optimization. The current logic doesn't work without it. It's a smell, it means the logic is not sound. In the shared informer, the cache is transparent, the calling code doesn't know there's a cache. The same should apply here. If there's a cache, great, but the code shouldn't break if the cache is not there.

Again, there is nothing that says that you can't retrieve the resources on each call. If we want to prevent that, we should have a logic that acts like it's retrieving the resource on each call but returns a cached version when available instead of having the cache drive the core of the logic. Otherwise the logic is based on fragile assumptions that things are cached in a certain way and that will lead to things breaking in subtle ways. The logic shouldn't even care about a cache being there or not.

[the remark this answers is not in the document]
csviri: I think I understand what you want to say:

> But the algorithm should work without optimization 

So I guess you are refereing to caching, so it should be more transparent (when asking for the lates custom resource). (This sentence can make sense in context of caches. Not in general - optimizes algorithms tends to be quite different from unoptimized ones).

So in this regard you are right, as I explained before, this was a simplest (quickest) approach that time to impement, and ment to be the first version, of this caching. 

I think the ideal would to contact the CustomResourceEventSource for the actual CustomResource - and not having the cache in this explicit way as it is now, it would be hidden behind the event source. 

In other words, as it is described here in caching section: https://csviri.medium.com/java-operator-sdk-introduction-to-event-sources-a1aab5af4b7b
The event source is responsible to make sure we have eventually  the fresh version of a resource. Currently this is true here, but in a not nice indirect way. So ideally where we are not reading the cache we would ask the eventsource for the latest resource, where the cr event source will decide if it takes it from local cache or calls the server. Eventually we can also replace the watches with informers there, so we delegate this caching to informer (but have to take a look deeper how fabric8 implements infomrers).

So in the cacing improvements issue I'm working on should reflect this. Would be better to discuss in detail in a call, if there are points which are not clear.

Just pls revert the changes in this PR (those ~4 lines), how the cache is propagted because it makes the implementation broken.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: To be clear, this example shows exactly what I mean: if you can't reason about the behavior of the code without having to take the caching into account then the behavior is fundamentally unsound, in my opinion. If the logic breaks because something wasn't cached, then the logic is wrong.

attilameszaros-form3: Look @metacosm , I think we discussed this already too much, we agree that this is not ideal, going to be improved/chaged. But it works. With this change it breaks. So we will improve this in a separate issue.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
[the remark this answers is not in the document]
metacosm: OK, I will revert the change for now but we need to make the cache not part of the behavioral logic: the actions that need to be taken based on the state shouldn't depend on what is cached and what is not. Querying the event source for the Custom Resource seems reasonable to me, and, indeed, could return a cached version or not. I was thinking also about using informers as well and it would make sense to make sure that the client's implementation is working properly when it comes to caching.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Context.java
```diff
@@ -2,13 +2,10 @@
 
 import io.fabric8.kubernetes.client.CustomResource;
 import io.javaoperatorsdk.operator.processing.event.EventList;
-import io.javaoperatorsdk.operator.processing.event.EventSourceManager;
 import java.util.Optional;
 
 public interface Context<T extends CustomResource> {
 
-  EventSourceManager getEventSourceManager();
```
csviri, line 10: Why is this removed from the context?

csviri: I think too that is not necessary there, just want to discuss the arguments.
So originally this looked a little bit different before the init method, the idea was that even dynamically would be possible to add event sources, or make some calls to event sources based on which resources is processed in which state. But users can set these instances directly to the class, so access those anyways possible. I basically agree with this.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
csviri: So what is the conclusion here?
Still think this new name is misleading.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Context.java
metacosm: I'm trying to clean up / simplify things: this wasn't used and it introduced an unnecessary dependency. We can always re-add it later if it fills a need.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: Inlined the method :)

csviri (approved):

PR #422, opened 2021-05-18
A fix for NPE when operator-framework-core is loaded dynamically.

gzukula:
I'm implementing k8s operator controller within a Keycloak plugin which is being loaded dynamically, and I'm getting NPE whenever I call "DefaultConfigurationService.instance()".
I assume because of this dynamic loading (and whatever thread model Keycloak is employing) the contextClassLoader is not able to find "version.properties" file, that eventually leads to NPE.

According to StackOverflow, using getContextClassLoader() is not reliable and is discouraged from using:
https://stackoverflow.com/questions/1771679/difference-between-threads-context-class-loader-and-normal-classloader

And a proper way to get a class loader in the static method, is to use an explicit reference to the class itself:
https://stackoverflow.com/questions/8275499/how-to-call-getclass-from-a-static-method-in-java

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
@@ -21,8 +21,7 @@ public class Utils {
    * @return a {@link Version} object encapsulating the version information
    */
   public static Version loadFromProperties() {
-    final var is =
-        Thread.currentThread().getContextClassLoader().getResourceAsStream("version.properties");
+    final var is = Utils.class.getClassLoader().getResourceAsStream("version.properties");
 
     final var properties = new Properties();
     if (is != null) {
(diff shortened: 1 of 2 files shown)


metacosm:
Strangely enough, these changes make the tests fail on my system. Not sure why they pass in CI. Another aspect is also the fact that this breaks the Quarkus extension (and I'm pretty sure this would also affect the Spring Boot integration, though I haven't checked).

Could you detail your use case a little bit more, using the `ThreadContext` `ClassLoader` is pretty standard when using frameworks so I'm a little surprised that this would cause an issue for you, since I usually have the reverse issue (i.e. using the class class loader resulting in the resources not being found).

gzukula:
I think this pull request could be discarded.

I have fixed my implementation by setting up a local class loader for the current thread, to avoid inheriting the one from the main thread. I can conclude that there is nothing to be fixed in the framework :)

metacosm:
That was going to be my next suggestion, I agree, though, that this is kind of subpar…

gzukula:
In retrospect, I think the change from the 1st commit (1e3877280) in my opinion should always be correct, as "version.properties" file should always belong to the same resource as the calling class (operator/api/config/Utils.java).

The tricky part, then I assume is to assume that the same class loader would be used for loading custom CRD controller and its resource classes as the class that is locating them (operator/config/runtime/ClassMappingProvider.java). I could imagine that there could be cases when that is not the case (when a dependency to java-operator-sdk is not explicitly embedded in the CRD controller jar), so I suspect that setting the right class loader in the controller thread seems to be a valid approach.

The whole classloading business is quite complicated to understand. Keycloak (that is loading my CRD controller) is using Wildfly/JBoss for locating its modules and loading them dynamically. I'm not entirely aware what is its impact on the classloading mechanism. Maybe there is a more solid explanation in the context of these technologies that are used there.

PR #423, opened 2021-05-18
fix: make test intent more explicit, disable wrong tests

metacosm:


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/EventDispatcherTest.java
```diff
@@ -121,6 +139,8 @@ void callsDeleteIfObjectHasFinalizerAndMarkedForDelete() {
 
   /** Note that there could be more finalizers. Out of our control. */
   @Test
+  @Disabled(
```
csviri, line 142: We should not disable these tests. Either should be fixed, or if does not make sense removed.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/EventDispatcherTest.java
```diff
@@ -89,29 +104,32 @@ void updatesBothResourceAndStatus() {
...
     eventDispatcher.handleExecution(
         executionScopeWithCREvent(Watcher.Action.MODIFIED, testCustomResource));
     verify(controller, times(1))
         .createOrUpdateResource(ArgumentMatchers.eq(testCustomResource), any());
   }
 
   @Test
-  void adsDefaultFinalizerOnCreateIfNotThere() {
+  @Disabled(
+      "This test is wrong, if the finalizer is not present, it is added, bypassing calling the controller")
```
csviri, line 118: Same here. probably should be removed.

csviri: This makes sense to address (other one probably not), but the controller actually should not be called. Untile there is no finalizer. So that condition is bad there. 
(The cleanup logic would be quite complex if we don't do it this way. Also could lead to undesired side effects.)

metacosm: The change that I'm making is that finalizer won't be required anymore (configurable) so the test will check that the controller is only called if the configuration says that there shouldn't be a finalizer.

csviri: That makes sense, I thing that would be a good feature. 
Just thinking if we should do some ADRs or at least issues to document the change and the arguments behind it.

PR #435, opened 2021-06-13
Don't install a shutdown hook by default

### LINKED ISSUE #434: Don't install a shutdown hook by default
When the Operator class is instantiated, it automatically registers a shutdown hook:

```java
Runtime.getRuntime().addShutdownHook(new Thread(this::close))
```

As the startup of the operator is responsibility of the operator developer, I think the shutdown should be left to the user otherwise there may be some conflict and undefined behavior because there may be the need to perform some logic before the operator shuts down.

I'm proposing to make this behavior explicit, like

```java
operator.installShutdownHook();
```

lburgazzoli:
Fixes #434

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -27,7 +27,9 @@ public class Operator implements AutoCloseable {
     this.k8sClient = k8sClient;
     this.configurationService = configurationService;
     this.closeables = new ArrayList<>();
+  }
 
+  public void installShutdownHook() {
     Runtime.getRuntime().addShutdownHook(new Thread(this::close));
   }
 


lburgazzoli:
@metacosm the failure here seems not related to he PR.

metacosm:
I'd rather keep the current behavior as the default so that `close` is automatically called when the operator is shut down. However, we could indeed have an `installShutdownHook` method on `Operator` that would take a lambda as parameter and wrap it so that `close` would also be called after whatever user logic needs to happen first.

lburgazzoli:
I wonder if that's correct, as example in quarkus, the k8s client is closed as part of the [lifecycle](https://github.com/quarkusio/quarkus/blob/main/extensions/kubernetes-client/runtime/src/main/java/io/quarkus/kubernetes/client/runtime/KubernetesClientProducer.java#L25-L30)  so I'm not sure if explicitly add something to the shutdown hook, may cause to invoke an operation on an already closed/disposed resource

metacosm:
On second thought, it makes sense to move ahead with this change, in particular so that frameworks can decide how to deal with stopping the operator in an idiomatic way. @lburgazzoli can you give me write access to your branch so that I can push updates before merging?

lburgazzoli:
@metacosm sent an invitation

metacosm:
Hmm, seems like I still can't push:  `! [remote rejected]   gh-434 -> gh-434 (refusing to allow an OAuth App to create or update workflow `.github/workflows/master-snapshot-release.yml` without `workflow` scope)`
I guess I need to do a proper fork as opposed to one from my IDE :)

lburgazzoli:
> Hmm, seems like I still can't push: `! [remote rejected] gh-434 -> gh-434 (refusing to allow an OAuth App to create or update workflow `.github/workflows/master-snapshot-release.yml`without`workflow` scope)`
> I guess I need to do a proper fork as opposed to one from my IDE :)

Since it is quite a small change, feel free to create a new PR, I really don't mind :)

metacosm:
Replaced by #498

PR #438, opened 2021-06-14
Add analytics to website.

amouat:
This PR adds Google Analytics to the website. I think it would be useful to get an idea of how many people are visiting the site. The analytics currently goes to Container Solutions.

This is just a proposal currently, please let me know if there are any concerns or issues.

I've not used the website generator before, so I may have made a mistake in there.

### DIFF
--- a/docs/_includes/analytics.html
+++ b/docs/_includes/analytics.html
@@ -1,0 +1,9 @@
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-TCB1LERV5K"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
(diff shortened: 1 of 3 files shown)


metacosm:
I don't really know how the site generator works either… but it would be nice if the project admins could easily access the data as well. How do you propose we do that?
If possible, as well, I'd rather use an open-source analytics solution but… 🤷🏼

amouat:
I can add people to the analytics "property" I think. I'm happy to use to open source or community alternative - were you thinking of any in particular?

amouat:
Ah, I think the main issue is we'd need to self host which is a bit of a pain :(

metacosm:
> Ah, I think the main issue is we'd need to self host which is a bit of a pain :(

Indeed… 😢

charlottemach:
LGTM

metacosm:
An interesting open-source alternative with hosted seemingly cheap model: https://plausible.io/

PR #441, opened 2021-06-21
Javadoc

teszes:
Putting it up as a draft as discussed, will continue working on it.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Context.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Context.java
@@ -4,9 +4,25 @@ import io.fabric8.kubernetes.client.CustomResource;
 import io.javaoperatorsdk.operator.processing.event.EventList;
 import java.util.Optional;
 
+/**
+ * Encapsulates information about the state of the cluster with respect to the provided custom
+ * resource class.
+ *
+ * @param <T> the {@link CustomResource} for which the context contains information
+ */
 public interface Context<T extends CustomResource> {
 
+  /**
+   * Gets the events related to the relevant custom resource
+   *
+   * @return the list of relevant {@link io.javaoperatorsdk.operator.processing.event.Event}s
+   */
   EventList getEvents();
 
+  /**
+   * Retrieves an object describing how to process retries.
+   *
+   * @return a {@link RetryInfo} object or null
+   */
   Optional<RetryInfo> getRetryInfo();
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Controller.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Controller.java
@@ -5,29 +5,52 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Interface to be implemented by user-provided controller classes.
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface Controller {
 
+  /**
+   * String representing no name provided.
+   */
   String NULL = "";
+
+  /**
+   * String representing config setting for custom resource controllers to watch the currently
+   * active namespace.
+   */
   String WATCH_CURRENT_NAMESPACE = "JOSDK_WATCH_CURRENT";
+
+  /**
+   * String representing explicitly that no finalizer should be added to custom resources managed by
+   * the controller.
+   */
   String NO_FINALIZER = "JOSDK_NO_FINALIZER";
 
+  /**
+   * Gets the name of the controller.
+   *
+   * @return the name of the controller
+   */
   String name() default NULL;
 
   /**
-   * Optional finalizer name, if it is not provided, one will be automatically generated. If the
-   * provided value is the value specified by {@link #NO_FINALIZER}, then no finalizer will be added
-   * to custom resources.
+   * Optional, the name of the finalizer attached to custom resources managed by the controller. If
+   * it is not provided one will be automatically generated. If the provided value is
+   * {@link #NO_FINALIZER}, then no finalizer will be attached to custom resources.
    *
    * @return the finalizer name
    */
   String finalizerName() default NULL;
 
   /**
-   * If true, will dispatch new event to the controller if generation increased since the last
-   * processing, otherwise will process all events. See generation meta attribute <a
-   * href="https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource">here</a>
+   * If true, will dispatch new event to the controller if
+   * <a href=
+   * https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource
+   * >generation</a>
+   * increased since the last processing, otherwise will process all events.
    *
    * @return whether the controller takes generation into account to process events
    */
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/DefaultContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/DefaultContext.java
@@ -4,21 +4,51 @@ import io.fabric8.kubernetes.client.CustomResource;
 import io.javaoperatorsdk.operator.processing.event.EventList;
 import java.util.Optional;
 
+/**
+ * Default implementation of the {@link Context} interface to be used by users when no custom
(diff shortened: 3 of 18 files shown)


csviri:
Hi Teszes,
I usually not fan of javadoc or comments from a code, while the intention is trivial from the code it self. 

(usually refering for the arguments found in clean code: https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882 which are very to the point).

But happy to hear the arguments why this would be benefical.

teszes:
My argument is basically either have javadoc and maintain it consistently as an API reference (not as "comments in the code", rather a separate hosted way of looking at the functions of the API, or not have it at all if we don't want to maintain it. This way we get to rely on it in some places, making the code itself less expressive (less "Clean Code"), but can't really use it in any other way.

Which approach should we take?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Context.java
```diff
@@ -4,9 +4,25 @@
...
+   * Gets the events related to the relevant custom resource
+   *
+   * @return the list of relevant {@link io.javaoperatorsdk.operator.processing.event.Event}s
+   */
   EventList getEvents();
 
+  /**
+   * Retrieves an object describing how to process retries.
+   *
+   * @return a {@link RetryInfo} object or null
```
metacosm, line 25: This is incorrect: the point of using `Optional` is that you never get `null` back…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Controller.java
```diff
@@ -5,29 +5,52 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Interface to be implemented by user-provided controller classes.
```
metacosm, line 9: Wrong, this is an annotation class, not an interface.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/Controller.java
```diff
@@ -5,29 +5,52 @@
...
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface Controller {
 
+  /**
+   * String representing no name provided.
```
metacosm, line 16: I'd rather use `Constant` than `String` either. People can see the type of the object without having it repeated in the doc.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/DefaultContext.java
```diff
@@ -4,21 +4,51 @@
...
+ * implementation of that interface is needed.
+ *
+ * @param <T> the {@link CustomResource} the context is built around
+ */
 public class DefaultContext<T extends CustomResource> implements Context<T> {
 
+  /**
+   * Object containing information on how retries are processed.
+   */
   private final RetryInfo retryInfo;
```
metacosm, line 18: I don't think we should be documenting any private details of classes.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/DefaultContext.java
```diff
@@ -4,21 +4,51 @@
...
+   * @param events the {@link EventList} containing the relevant events to the resource
+   * @param retryInfo the {@link RetryInfo} object encapsulating info about retries
+   */
   public DefaultContext(EventList events, RetryInfo retryInfo) {
     this.retryInfo = retryInfo;
     this.events = events;
   }
 
+  /**
+   * Gets the events associated with this context instance.
```
metacosm, line 38: We shouldn't be documenting implementations of interface / abstract class methods unless their implementation needs to the be specifically documented and/or we want to call out the users' attention on a particular detail.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/DeleteControl.java
```diff
@@ -1,5 +1,14 @@
 package io.javaoperatorsdk.operator.api;
 
+import io.fabric8.kubernetes.client.CustomResource;
+
+/**
+ * Encapsulates information about the logic executed when instances of the given custom resource are
+ * deleted. Returned from {@link ResourceController#createOrUpdateResource}. Informs the framework
```
metacosm, line 7: This is wrong: `DeleteControl` is returned from the `deleteResource` method.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -3,53 +3,47 @@
...
   /**
-   * The implementation should delete the associated component(s). Note that this is method is
-   * called when an object is marked for deletion. After it's executed the custom resource finalizer
-   * is automatically removed by the framework; unless the return value is {@link
-   * DeleteControl#NO_FINALIZER_REMOVAL}, which indicates that the controller has determined that
-   * the resource should not be deleted yet, in which case it is up to the controller to restore the
-   * resource's status so that it's not marked for deletion anymore.
+   * This method is called when a managed custom resource instance has been marked for deletion.
+   * After execution the custom resource finalizer is automatically removed unless the return value
+   * is {@link DeleteControl#NO_FINALIZER_REMOVAL}.
```
metacosm, line 17: As previously discussed, as a developer, I want to see that the method needs to be idempotent in the javadoc because that information cannot be communicated in any other way.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java
```diff
@@ -3,53 +3,47 @@
...
-   *     <b>However we will always call an update if there is no finalizer on object and it's not
-   *     marked for deletion.</b>
+   * @return either an {@link UpdateControl} object, or {@link UpdateControl#noUpdate()} if no
+   * updates need to be sent to the API server.
    */
   UpdateControl<R> createOrUpdateResource(R resource, Context<R> context);
 
   /**
-   * In init typically you might want to register event sources.
+   * Initialization function that registers event sources using a manager object.
```
metacosm, line 43: This description really isn't useful. That said, the previous one wasn't all that useful either. Ideally, we'd document the intent of the method, what it needs to be used for, not what it does.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/RetryInfo.java
```diff
@@ -1,8 +1,22 @@
 package io.javaoperatorsdk.operator.api;
 
+/**
+ * Interface to be implemented by user-defined classes encapsulating information about how many
```
metacosm, line 4: This isn't true: this isn't a class that users need to implement. It's simply providing a contract so that we don't expose internal implementation classes. I'm not even sure this class needs to be exposed to users at all, to be honest.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/Event.java
```diff
@@ -1,8 +1,21 @@
 package io.javaoperatorsdk.operator.processing.event;
 
+/**
+ * The interface all user-defined events must implement. Also implemented by {@link AbstractEvent}.
```
metacosm, line 4: Please do not explicitly list implementations. This is automatically handled by javadoc itself.

metacosm (changes requested):
I agree that we need more documentation but there is a balance to be found here.

No documentation is better than wrong documentation, the corollary being that the more documentation you have, the more likely it is to become wrong at some point. We therefore need to focus our documentation efforts and not try for quantity but rather strive for quality. 

Consequently:
- no need to document private details unless some non-obvious implementation details need to be pointed out to developers
- no need to document methods that override or implement already documented public methods unless there's an explicit need to do so because of their particular implementation
- no need to call out implementations / extensions of public classes/interfaces as this is handled by javadoc automatically
- document intent and not implementation as intent shouldn't change while implementation might

metacosm:
On a purely project maintenance side: please use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for your commit messages, no need to reference the file name. Also, please use rebase instead of merges to update your branches.

metacosm:
@teszes do you still intend to work on this issue?

metacosm:
@teszes ?

teszes:
Hi, sorry, got caught up in other stuff, I'll resume working on this now

teszes:
I'll close this for now and reopen when I have more time to dedicate to this.

PR #443, opened 2021-06-23
Change stream implementation to List.contains()

teszes:
This stream filtering implementation is basically the same as using List.contains(), and any performance differences might be negligible since we are just seeing if dependent classes are namespaced (there shouldn't be _that_ many of them).


teszes:
Could close #445

metacosm (approved):
LGTM

teszes:
Verified unit test coverage, merging.

PR #444, opened 2021-06-23
Change server version error to exception rather that system stop call

### LINKED ISSUE #408: Operator start behavior
The Operator.start method is implemented like:

```java
public void start() {
   ...
    try {
      ...
    } catch (Exception e) {
      log.error("Error retrieving the server version. Exiting!", e);
      System.exit(1);
    }
  }
```

I wonder if instead of doing a `System.exit(1)`, the start method should thrown an exception and let the caller decide what to do.

teszes:
One way to close #408

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -51,7 +51,7 @@ public class Operator implements AutoCloseable {
       }
     } catch (Exception e) {
       log.error("Error retrieving the server version. Exiting!", e);
-      throw new OperatorException("Error retrieving the server version");
+      throw new OperatorException("Error retrieving the server version", e);
     }
   }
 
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -51,7 +51,7 @@ public void start() {
       }
     } catch (Exception e) {
       log.error("Error retrieving the server version. Exiting!", e);
-      System.exit(1);
+      throw new OperatorException("Error retrieving the server version");
```
attilameszaros-form3, line 54: We should add the root cause exception, are we sure that this is always version?

metacosm: It most often is, but, strictly speaking, there might be other causes.

teszes: chained the thrown exception to the new one, would this work the way you expect?

metacosm:
Have you checked that throwing the exception properly crashes the app, resulting in a new pod being brought up when running in a cluster environment?

metacosm (approved):

PR #454, opened 2021-07-13
feat: use a predicate to select the custom resource for which a reconcile should be triggered

### LINKED ISSUE #430: event source: use a predicate to select the custom resource for which a reconcile should be triggered
If my understanding is correct, as today events triggers a reconcile on a CR matching by UID.

Sometimes is not possible to determine the UID because i.e. the reconcile should be triggered for an event outside kubernetes so it would be nice if events would be allowed to specify a predicate to filter out the CR for which a reconcile should ne triggered.

For backward compatibility an predicate based on UID can be created

Does this make sense ?
If it does I can work on some POC

lburgazzoli:
Fixes #430

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
@@ -19,11 +19,13 @@ import io.javaoperatorsdk.operator.processing.retry.RetryExecution;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -108,8 +110,19 @@ public class DefaultEventHandler implements EventHandler {
     try {
       lock.lock();
       log.debug("Received event: {}", event);
-      eventBuffer.addEvent(event);
-      executeBufferedEvents(event.getRelatedCustomResourceUid());
+
+      Predicate<CustomResource> selector = event.getCustomResourcesSelector();
+      if (selector == null) {
+        final String uid =
+            Objects.requireNonNull(event.getRelatedCustomResourceUid(), "CustomResource UID");
+
+        selector = customResource -> Objects.equals(uid, customResource.getMetadata().getUid());
+      }
+
+      for (String uid : eventSourceManager.getLatestResourceUids(selector)) {
+        eventBuffer.addEvent(uid, event);
+        executeBufferedEvents(uid);
+      }
     } finally {
       lock.unlock();
     }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -12,9 +12,11 @@ import io.javaoperatorsdk.operator.processing.DefaultEventHandler;
 import io.javaoperatorsdk.operator.processing.event.internal.CustomResourceEventSource;
 import io.javaoperatorsdk.operator.processing.event.internal.TimerEventSource;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
@@ -161,6 +163,16 @@ public class DefaultEventSourceManager implements EventSourceManager {
   // todo: remove
   public Optional<CustomResource> getLatestResource(String customResourceUid) {
     return getCache().getLatestResource(customResourceUid);
+  }
+
+  // todo: remove
+  public List<CustomResource> getLatestResources(Predicate<CustomResource> selector) {
+    return getCache().getLatestResources(selector);
+  }
+
+  // todo: remove
+  public Set<String> getLatestResourceUids(Predicate<CustomResource> selector) {
+    return getCache().getLatestResourcesUids(selector);
   }
 
   // todo: remove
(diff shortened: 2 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -108,13 +109,28 @@ public void handleEvent(Event event) {
...
+          eventBuffer.addEvent(uid, event);
+          executeBufferedEvents(uid);
+        }
+      }
     } finally {
       lock.unlock();
     }
   }
 
+  private void executeBufferedEvents(List<String> customResourceUids) {
```
metacosm, line 128: Both versions of the method should be merged into a single one, imo.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEvent.java
```diff
@@ -0,0 +1,16 @@
+package io.javaoperatorsdk.operator.processing.event;
+
+import io.fabric8.kubernetes.client.CustomResource;
+import java.util.function.Predicate;
+
+@SuppressWarnings("rawtypes")
+public class DefaultEvent extends AbstractEvent {
```
metacosm, line 7: Do we really need this class?

metacosm (changes requested):
Overall, looks OK, just some minor comments and improvements.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -108,13 +109,28 @@ public void handleEvent(Event event) {
     try {
       lock.lock();
       log.debug("Received event: {}", event);
-      eventBuffer.addEvent(event);
-      executeBufferedEvents(event.getRelatedCustomResourceUid());
+
+      if (event.getRelatedCustomResourceUid() != null) {
```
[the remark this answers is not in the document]
lburgazzoli: Absolutely, the reason I left is that it could be slightly more efficient as it would avoid an extra lookup but I will fix it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
lburgazzoli: this is a leftover, I'll remove it

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEvent.java
lburgazzoli: We don't but I found a little bit annoying that AbstractEvent is declared abstract but in fact it does not have any abstract method so if you want to create a generic event, you'd need an anonymous class.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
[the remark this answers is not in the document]
lburgazzoli: done

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
lburgazzoli: done

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEvent.java
lburgazzoli: removed

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/CustomResourceSelectorTest.java
```diff
@@ -0,0 +1,120 @@
...
+        ArgumentCaptor.forClass(ExecutionScope.class);
+
+    verify(eventDispatcherMock, timeout(SEPARATE_EXECUTION_TIMEOUT).times(2))
+        .handleExecution(executionScopeArgumentCaptor.capture());
+
+    assertThat(executionScopeArgumentCaptor.getAllValues())
+        .hasSize(2)
+        .allSatisfy(
+            s -> {
+              assertThat(s.getEvents()).isNotEmpty().hasOnlyElementsOfType(DefaultEvent.class);
```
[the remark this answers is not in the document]
lburgazzoli: done

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -163,6 +165,16 @@ public CustomResourceCache getCache() {
     return getCache().getLatestResource(customResourceUid);
   }
 
+  // todo: remove
```
metacosm, line 168: Why are these methods marked as `todo: remove`?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/CustomResourceSelectorTest.java
```diff
@@ -0,0 +1,125 @@
...
+          eventDispatcherMock,
+          "Test",
+          null,
+          ConfigurationService.DEFAULT_RECONCILIATION_THREADS_NUMBER);
+
+  @BeforeEach
+  public void setup() {
+    defaultEventHandler.setEventSourceManager(defaultEventSourceManagerMock);
+
+    // todo: remove
```
metacosm, line 47: Same comment as above…

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEvent.java
metacosm: I'll refactor `AbstractEvent` separately.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
metacosm: OK, I'll try to figure out what's going on there… :)

metacosm:
Thank you!

Issue #453, opened 2021-07-12
Option to limit CRs the operator watches


lburgazzoli:
We should be able to restrict what CRs an operator handles, as example, by using a selector:

```java
public @interface Controller(
    LabelSelector[] selectors() default {}
)
```

This could be required when a canary deployment of operators is required so at certain point in time, there will be more than one operator running in a cluster each one is taking into account a subset of the CRs.

lburgazzoli:
@metacosm I have some spare cycles this week so if you need any help on this feature , I'd be happy to take a stab

csviri:
@lburgazzoli This sounds awsome!

metacosm:
Ideally, however we implement this, it should boil down to generating a `Predicate`. Other aspects to keep in mind as well: 
- I don't think it will be possible (nor desirable, for that matter) to use `LabelSelector` on the `Controller` annotation
- Whatever configuration we add to the annotation, it should be possible to have similar configuration in application properties as well

lburgazzoli:
I don't know if a predicate would be the best approach as then it would filter events after we receive them, instead a label selector (or anything similar) would configure the watcher to only watch some resources

metacosm:
> I don't know if a predicate would be the best approach as then it would filter events after we receive them, instead a label selector (or anything similar) would configure the watcher to only watch some resources

Ah, right. Should we also filter on annotations?

> 
> I guess on quarkus a possible representation using properties could be:
> 
> ```
> quarkus.operator-sdk.controllers."foo".selectors[0] = "environment=production"
> quarkus.operator-sdk.controllers."foo".selectors[1] = "tier!=frontend"
> ```
> 
> or
> 
> ```
> quarkus.operator-sdk.controllers."foo".selectors = "environment=production,tier!=frontend"
> ```

I'd prefer the second option (which I think Quarkus supports out of the box, iirc).

lburgazzoli:
> > I don't know if a predicate would be the best approach as then it would filter events after we receive them, instead a label selector (or anything similar) would configure the watcher to only watch some resources
> 
> Ah, right. Should we also filter on annotations?
> 

As far as I know, you can't use annotation to select objects, only labels can be used.

csviri:
Yep annotations are not indexed in etcd AFK, so we cannot filtern on them.

[...]

PR #457, opened 2021-07-22
Custom event filter for controllers

### LINKED ISSUE #404: Custom event filter for controllers
By default the each controller implements a generation-based event filtering [1] thus a controller won't receive updates when i.,e. a status field is changed and if a controller need to react upon changes in the status field, the generation-based event filtering can be turned off i.e. by setting `quarkus.operator-sdk.controllers."${controller-name}".generation-aware = false` for quarkus based operators.

It would be nice to have some more options here so instead of having a simple property to turn on/of  generation-based filtering it would be nice to give to the operators developer a chance to provide a custom filter, i.e. a custom strategy could be injected upon controller initialization i.e. within the `ResourceController::init` method

[1] https://blog.container-solutions.com/a-deep-dive-into-the-java-operator-sdk

lburgazzoli:
Fixes #404


lburgazzoli:
@metacosm this PR is incomplete as it lacks some tests, however I'd like to discuss if it goes in the right direction because there are some behavioral changes:

- the custom resource cache did not perform a copy of the resource in the `cacheResource` method, which could lead unexpected results (but I don't know it this was expected)
- the latest generation value was cached as part of the CustomEventSource after a successful invocation of the event handler and I'm not sure about the reasoning behind that logic as I would assume that any retry logic should not be affected by the revision.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -128,46 +135,16 @@ public void eventReceived(Watcher.Action action, T customResource) {
...
-    if (lastGeneration == null) {
-      return false;
-    } else {
-      return customResource.getMetadata().getGeneration() <= lastGeneration;
+      return;
     }
-  }
 
-  @Override
-  public void eventSourceDeRegisteredForResource(String customResourceUid) {
```
metacosm, line 169: Why was this method removed?

lburgazzoli: because the method was only about to remove the generation from the local cache, but since the cache does not exists anymore the method is useless (and the interface provide a default no-op method)

lburgazzoli:
@csviri @metacosm what do you think about the PR ?

csviri:
> @csviri @metacosm what do you think about the PR (in general) ?

@lburgazzoli will try to take a look soon, like the idea of predicate very much! Will check deeper later.

lburgazzoli:
> > @csviri @metacosm what do you think about the PR (in general) ?
> 
> @lburgazzoli will try to take a look soon, like the idea of predicate very much! Will check deeper later.

At the moment it is like a POC so, lot of things may be wrong :)

lburgazzoli:
@csviri @metacosm I need to add some tests but it should be ready for review

metacosm:
How do you propose to configure the predicate?

lburgazzoli:
> How do you propose to configure the predicate?

This is an advanced functionality which I don't think it should be made configurable as it depends on the specific spec/status and behavior of the operator so I think this should not be exposed as a configuration option (IMHO, the "generation aware" configuration option should not be exposed too)

metacosm:
> > How do you propose to configure the predicate?
> 
> This is an advanced functionality which I don't think it should be made configurable as it depends on the specific spec/status and behavior of the operator so I think this should not be exposed as a configuration option (IMHO, the "generation aware" configuration option should not be exposed too)

I'm confused: I thought the whole point of this PR was to be able to configure the behavior with more than just a property or annotation value. Could you please explain the use case in greater details, then?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -109,38 +111,6 @@ public void eventReceived(Watcher.Action action, T customResource) {
     }
   }
 
-  private void markLastGenerationProcessed(T resource) {
```
csviri, line 112: sorry, not following why this was removed ? I mean the whole concept or generation awareness? Or am I missing something?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourcePredicates.java
```diff
@@ -0,0 +1,54 @@
+package io.javaoperatorsdk.operator.processing.event.internal;
+
+import io.fabric8.kubernetes.client.CustomResource;
+import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
+
+public final class CustomResourcePredicates {
+
+  private static final CustomResourcePredicate<CustomResource> GENERATION_AWARE =
```
csviri, line 8: I'm not sure this is equivalent to the old generation processing, is this aware o retries for example?

csviri:
hi @lburgazzoli @metacosm sorry busy days, will try to take a look today

lburgazzoli:
> > > How do you propose to configure the predicate?
> > 
> > 
> > This is an advanced functionality which I don't think it should be made configurable as it depends on the specific spec/status and behavior of the operator so I think this should not be exposed as a configuration option (IMHO, the "generation aware" configuration option should not be exposed too)
> 
> I'm confused: I thought the whole point of this PR was to be able to configure the behavior with more than just a property or annotation value. Could you please explain the use case in greater details, then?

Oh sorry I though that with `wow do you propose to configure the predicate` you meant how to use a property to configure a predicate. 

So far what I did is to use `ControllerConfiguration` upon registration so like 

```java
var operator = new Operator();
operator.register(new MyController(), new MyConfiguration() {
    public CustomResourcePredicate<MyResource> getPredicate() {
        return ...;
    }
});
```

metacosm:
OK, so that should part of the configuration overriding mechanism, then…

lburgazzoli:
> OK, so that should part of the configuration overriding mechanism, then…

will have a look later this week or beginning next week

lburgazzoli:
> > OK, so that should part of the configuration overriding mechanism, then…
> 
> will have a look later this week or beginning next week

I've added support to the conf override, now I wonder if it still make sense to expose the option to configure the generation awarness

metacosm (commented):
Please rebase your PR so that all commits follow the conventional commit message format: https://www.conventionalcommits.org/en/v1.0.0/

metacosm:
> > > OK, so that should part of the configuration overriding mechanism, then…
> > 
> > 
> > will have a look later this week or beginning next week
> 
> I've added support to the conf override, now I wonder if it still make sense to expose the option to configure the generation awarness

I'm not sure either but then again we can't break the API before v2. That said, the other question I have is how someone using Quarkus where the controller is automatically registered would be able to specify their predicates and have it taken into account…

lburgazzoli:
> I'm not sure either but then again we can't break the API before v2. That said, the other question I have is how someone using Quarkus where the controller is automatically registered would be able to specify their predicates and have it taken into account…

What about something like:

```java
@Target({ElementType.TYPE})
public @interface Controller {
    Class<CustomResource> eventFilter default GenrationAwareFilter.class;
}
```

```java
@Controller(
      eventFilter = MyCustomResourcePredicate.class
)
```

As alternative the controller should be able to override some config in its `init` method

csviri:
> > I'm not sure either but then again we can't break the API before v2. That said, the other question I have is how someone using Quarkus where the controller is automatically registered would be able to specify their predicates and have it taken into account…
> 
> What about something like:
> 
> ```java
> @Retention(RetentionPolicy.RUNTIME)
> @Target({ElementType.TYPE})
> public @interface Controller {
>     Class<CustomResource> eventFilter default GenrationAwareFilter.class;
> }
> ```
> 
> ```java
> @Controller(
>       eventFilter = MyCustomResourcePredicate.class
> )
> ```
> 
> As alternative the controller should be able to set some config option in its `init` method (that would be my preferred option)

@lburgazzoli  @metacosm just to add my two cents.  

- should not be this a list of predicates in the annotation? So it's easy to compose more predicates. 
- we could consider having the generation awareness as a boolean (as it is now ) and just add the predicate in the background. The only reason to do this is just to have this nice explicit api about generations, since those are quite central concepts of event processing. While the other predicates will usually solve some special cases I guess.

lburgazzoli:
> 
> @lburgazzoli @metacosm just to add my two cents.
> 
>     * should not be this a list of predicates in the annotation? So it's easy to compose more predicates.

I think it can be done, given that we assume the operators are evaluated as an AND right ?

> 
>     * we could consider having the generation awareness as a boolean (as it is now ) and just add the predicate in the background. The only reason to do this is just to have this nice explicit api about generations, since those are quite central concepts of event processing. While the other predicates will usually solve some special cases I guess.

The main issue I see by keeping both is the potential conflict between the generation aware option and the predicates. One option could be that if `generationAware=true`, then a predicate to check the generation will be registered in any case in addition to those defined in the `eventFilter` but not sure if that's a good UX.

csviri:
> I think it can be done, given that we assume the operators are evaluated as an AND right ?

Yes , that is an end. But I guess that will be true for most of the cases. (Like Combining the generation awareness and some other)

> The main issue I see by keeping both is the potential conflict between the generation aware option and the predicates. One option could be that if generationAware=true, then a predicate to check the generation will be registered in any case in addition to those defined in the eventFilter but not sure if that's a good UX.

Yes indeed, agree that can be a little bit confusing (not too much though, since there is always AND).  Also the generation awareness can be turned off, and then the custom predicates will receive all the events. 
Actually I'm fine with both ways.

lburgazzoli:
> 
> Yes indeed, agree that can be a little bit confusing (not too much though, since there is always AND). Also the generation awareness can be turned off, and then the custom predicates will receive all the events.
> Actually I'm fine with both ways.

@csviri @metacosm added an approach where the generation aware can be retained and used to build the final list of filters.

lburgazzoli:
@csviri @metacosm should be ready for review

metacosm:
Do you have examples of use cases apart from generation-based processing where this would be useful? I'm still not quite convinced that there's really a need for this feature…

lburgazzoli:
> Do you have examples of use cases apart from generation-based processing where this would be useful? I'm still not quite convinced that there's really a need for this feature…

This is mainly needed when you want to have a mix of generation aware and non generation aware, as example in my operator, I turned off the generation awareness because I want to use the status field as state machine, so when I change the "phase" of the status, I want the control loop to be triggered but I don't want to get an event when other fields of the status are updated so my logic is like:

```java
(conf, oldRes, newRes -> {
    return Objects.equals( oldRes.metadata.generation, newRes.metadata.generation )
        || Objects.equals( oldRes.status.phase, newRes.status.phase );
}
```

metacosm:
@lburgazzoli are you OK with the changes I made, btw?

lburgazzoli:
> @lburgazzoli are you OK with the changes I made, btw?

yes I am

csviri:
> > Do you have examples of use cases apart from generation-based processing where this would be useful? I'm still not quite convinced that there's really a need for this feature…
> 
> This is mainly needed when you want to have a mix of generation aware and non generation aware, as example in my operator, I turned off the generation awareness because I want to use the status field as state machine, so when I change the "phase" of the status, I want the control loop to be triggered but I don't want to get triggered when other fields of the status are updated so my logic is like:
> 
> ```java
> (conf, oldRes, newRes -> {
>     return Objects.equals( oldRes.metadata.generation, newRes.metadata.generation )
>         || Objects.equals( oldRes.status.phase, newRes.status.phase );
> }
> ```

Hi @lburgazzoli @metacosm 
IMHO having this predicates would make the even filtering very generic. On the other hand the use case mentioned about it's quite exotic, especially from the angle that the state should not be stored in status. Maybe it would be worth to discuss this with broader audience on community meetings and go through the arguments. What do you think?

What to check the code today or weekend also if you don't mind waiting few days still.

lburgazzoli:
> > > Do you have examples of use cases apart from generation-based processing where this would be useful? I'm still not quite convinced that there's really a need for this feature…
> > 
> > 
> > This is mainly needed when you want to have a mix of generation aware and non generation aware, as example in my operator, I turned off the generation awareness because I want to use the status field as state machine, so when I change the "phase" of the status, I want the control loop to be triggered but I don't want to get triggered when other fields of the status are updated so my logic is like:
> > ```java
> > (conf, oldRes, newRes -> {
> >     return Objects.equals( oldRes.metadata.generation, newRes.metadata.generation )
> >         || Objects.equals( oldRes.status.phase, newRes.status.phase );
> > }
> > ```
> 
> Hi @lburgazzoli @metacosm
> On the other hand the use case mentioned about it's quite exotic, especially from the angle that the state should not be stored in status. Maybe it would be worth to discuss this with broader audience on community meetings and go through the arguments. What do you think?

What do you mean with "the state should not be stored in status" ?

> 
> What to check the code today or weekend also if you don't mind waiting few days still.

I do not have any issue waiting some more time

csviri:
> What do you mean with "the state should not be stored in status" 
As described here a good practice is that if you want to store state it should not go to the status, rather into a configmap or other CR:
https://github.com/java-operator-sdk/java-operator-sdk/issues/525#issuecomment-911859202

> I do not have any issue waiting some more time
Thank you!

metacosm:
> Hi @lburgazzoli @metacosm
> IMHO having this predicates would make the even filtering very generic. On the other hand the use case mentioned about it's quite exotic, especially from the angle that the state should not be stored in status. Maybe it would be worth to discuss this with broader audience on community meetings and go through the arguments. What do you think?

I think that there are cases where you can't really avoid this, though, especially when you need to synchronize the work of several controllers or dealing with external resources (though I guess you could argue that external resources should be modeled via CRs as well with their own controller).

metacosm:
> > What do you mean with "the state should not be stored in status"
> 
> As described here a good practice is that if you want to store state it should not go to the status, rather into a configmap or other CR:
> [#525 (comment)](https://github.com/java-operator-sdk/java-operator-sdk/issues/525#issuecomment-911859202)

This makes sense but can lead to combinatorial explosion of CRs and Controllers so it might not always be practical. Putting the state in `ConfigMap` is just no better than using status, imo.

csviri:
> I think that there are cases where you can't really avoid this, though, especially when you need to synchronize the work of several controllers or dealing with external resources (though I guess you could argue that external resources should be modeled via CRs as well with their own controller).

Ye, basically, this is why I want to discuss it a little bit further. So again our interface is not confusing and does not lead the user to anti-patterns. I agree that this might be useful in some situations, but if understand correctly:

- it's still a performance optimization. 
- if the cases we would cover are exotic or very rare, the logic of filtering could be implemented in the controller itself.

metacosm:
>     * it's still a performance optimization.

I guess sometimes it's more than just optimization. If you need to synchronize several controllers and need to introduce another CR to do that and another controller to deal with it, then you increase your code complexity quite a bit. It might still be cleaner to do it that way, I guess it really is use-case dependent, though.

>     * if the cases we would cover are exotic or very rare, the logic of filtering could be implemented in the controller itself.

That's a good point, this implementation allows bypass calling the controller altogether though, so it would be more efficient that way. You could argue that generation-based processing should also be implemented in the controller as well… and yet we cover that use case, so to me, it makes sense to extend the filtering capability as long as it doesn't encourage bad patterns.

csviri:
> So I guess my point is that we should encourage people to use good patterns but also understand that external considerations (such as performance requirement) might lead to different patterns and make them possible. Most users won't bother getting into creating a custom filter like this for their events unless they really have to.

Completely agree with this.

So yes, I guess this is more about design principles of the framework. How hard we should push the users, so we don't give them other possibilities that we (or the community ) thinks as a good practice. Or how open should we be on other approaches, and give the users possibility to implement different patters, even those are quite exotic (not necessary this case). (Removing the events from context also related to these principles.)

This is a nice topic on it's own, also might worth to discuss on community meeting. 

So just to understand my view, I don't think PR in terms of principles bad at all, just we should clearly decide and understand what way we want to go. And if we should support the non-standard patters explicitly in SDK with these extension mechanisms. 
Until it does not bring complexity to the framework I'm totally fine with it btw.

csviri:
> You could argue that generation-based processing should also be implemented in the controller as well… and yet we cover that use case, so to me, it makes sense to extend the filtering capability as long as it doesn't encourage bad patterns.

That is true, althoug this explicitly defined by kubernetes itself (so by definition is a special case). That is why generations where created in first place to this filtering is easy on controller side.

lburgazzoli:
> > What do you mean with "the state should not be stored in status"
> 
> As described here a good practice is that if you want to store state it should not go to the status, rather into a configmap or other CR:
> [#525 (comment)](https://github.com/java-operator-sdk/java-operator-sdk/issues/525#issuecomment-911859202)
> 

I originally tried to go that path but I found it not very efficient at least for my use case as:

- you may have to hit the kubernetes api-server for each reconcile loop to get the dependant resource because as today there is no real caching mechanism in the operator sdk except the one for the primary custom resources handled by the operator
- in a single reconcile loop you may end up updating both the main and dependant resource so you end up doing at least two calls to the Kubernetes API server
- you may also need to set up a watch on the dependant resource (i.e. if the resource is changed outside the control of the operator you may need to re-write it) so you still need to keep track of the geenration or version of the dependant resourece in your status

So for my use case I ended up by abusing a little bit of the status (I'm aware of that) to implement a state machine because the other option were not really practical or they were making the code much more complex that adding a field of two in the status sub-resource.

lburgazzoli:
> > You could argue that generation-based processing should also be implemented in the controller as well… and yet we cover that use case, so to me, it makes sense to extend the filtering capability as long as it doesn't encourage bad patterns.
> 
> That is true, althoug this explicitly defined by kubernetes itself (so by definition is a special case). That is why generations where created in first place to this filtering is easy on controller side.

The problem here is that the reconcile loop is not aware of the "old resource" so it can't determine if it should process an event or not. The only point I found where the old resource can be retrieved is in the custom resource event source so that's why I opened the issue and the PR

csviri:
> The problem here is that the reconcile loop is not aware of the "old resource" so it can't determine if it should process an event or not. The only point I found where the old resource can be retrieved is in the custom resource event source so that's why I opened the issue and the PR

From old resource you need only the `phase`? I guess this should be stored in a state config map. Currently the old resource is available in the cache, but this will **not** be guaranteed in the future if we are going to improve caching.

> you may have to hit the kubernetes api-server for each reconcile loop to get the dependant resource because as today there is no real caching mechanism in the operator sdk except the one for the primary custom resources handled by the operator. 

This is not explicitly supported. But should be fairly easy to create a event source which wraps an informer. (We can maybe try to take a look on this together). So you can access the cached resource from the EventSource / Informer.

> in a single reconcile loop you may end up updating both the main and dependant resource so you end up doing at least two calls to the Kubernetes API server

This is true. So if you would update the status and the "state config map" it would be two api calls. 

> you may also need to set up a watch on the dependant resource (i.e. if the resource is changed outside the control of the operator you may need to re-write it) so you still need to keep track of the generation or version of the dependant resource in your status and you need to find a way to skip events related to the action you perform on the dependant resource

Yes you would need the event source for this. This is an interesting point how to handle you own changes to the dependent resources. Interesting thing here is that here you are interested in the status change only of the dependant resource in general. But actually in your case (for the phase) when you are just store the state there you are not necessairly interested on changes on that dependant resource? You just read it next time you would reconcile, or?

csviri:
> The problem here is that the reconcile loop is not aware of the "old resource" so it can't determine if it should process an event or not. The only point I found where the old resource can be retrieved is in the custom resource event source so that's why I opened the issue and the PR

From old resource you need only the `phase`? I guess this should be stored in a state config map. Currently the old resource is available in the cache, but this will **not** be guaranteed in the future if we are going to improve caching.

> you may have to hit the kubernetes api-server for each reconcile loop to get the dependant resource because as today there is no real caching mechanism in the operator sdk except the one for the primary custom resources handled by the operator. 

This is not explicitly supported. But should be fairly easy to create a event source which wraps an informer. (We can maybe try to take a look on this together). So you can access the cached resource from the EventSource / Informer.

> in a single reconcile loop you may end up updating both the main and dependant resource so you end up doing at least two calls to the Kubernetes API server

This is true. So if you would update the status and the "state config map" it would be two api calls. 

> you may also need to set up a watch on the dependant resource (i.e. if the resource is changed outside the control of the operator you may need to re-write it) so you still need to keep track of the generation or version of the dependant resource in your status and you need to find a way to skip events related to the action you perform on the dependant resource

Yes you would need the event source for this. This is an interesting point how to handle you own changes to the dependent resources. Interesting thing here is that here you are interested in the status change only of the dependant resource in general. But actually in your case (for the phase) when you are just store the state there you are not necessairly interested on changes on that dependant resource? You just read it next time you would reconcile, or?

lburgazzoli:
> > The problem here is that the reconcile loop is not aware of the "old resource" so it can't determine if it should process an event or not. The only point I found where the old resource can be retrieved is in the custom resource event source so that's why I opened the issue and the PR
> 
> From old resource you need only the `phase`? I guess this should be stored in a state config map. Currently the old resource is available in the cache, but this will **not** be guaranteed in the future if we are going to improve caching. So it should not be relied on.
> 

If the plan is to move to the fabric8 informer, then it provides a method where in case of an update, the old a new resource are provided so I guess, the old resource will still be available somehow. But in general if we are implementing a cache, then I don't really see how the old cached value can disappear :)

> > you may have to hit the kubernetes api-server for each reconcile loop to get the dependant resource because as today there is no real caching mechanism in the operator sdk except the one for the primary custom resources handled by the operator.
> 
> This is not explicitly supported. But should be fairly easy to create a event source which wraps an informer. (We can maybe try to take a look on this together). So you can access the cached resource from the EventSource / Informer.
> 

Yes but we need to be aware that by introducing another cache, then we impact also the memory profile of the operator. If you have a large set of resource to handle, having a side resource to track the status would mean you have doubled the number of objects you have in memory. It does not mean that you double the overall memory but still it something to take into account.

> > in a single reconcile loop you may end up updating both the main and dependant resource so you end up doing at least two calls to the Kubernetes API server
> 
> This is true. So if you would update the status and the "state config map" it would be two api calls.
> 
> > you may also need to set up a watch on the dependant resource (i.e. if the resource is changed outside the control of the operator you may need to re-write it) so you still need to keep track of the generation or version of the dependant resource in your status and you need to find a way to skip events related to the action you perform on the dependant resource
> 
> Yes you would need the event source for this. This is an interesting point how to handle you own changes to the dependent resources. Interesting thing here is that here you are interested in the status change only of the dependant resource in general. But actually in your case (for the phase) when you are just store the state there you are not necessairly interested on changes on that dependant resource? You just read it next time you would reconcile, or?
> 

I have to watch the dependant resource because that resource may be changed by someone else (i.e. kubectl edit configmap foo-bar) and in such case I have to implement corrective measures. Of course you can try to have some rbac/opa rules but you see that the whole setup may become much more complex that just letting me to add a simple predicate :)

> (To be honest, this storing the state in a config map or CR feels also little bit smelly for me :D but I guess this si how it is now on k8s world )

lburgazzoli:
> > > The problem here is that the reconcile loop is not aware of the "old resource" so it can't determine if it should process an event or not. The only point I found where the old resource can be retrieved is in the custom resource event source so that's why I opened the issue and the PR
> > 
> > 
> > From old resource you need only the `phase`? I guess this should be stored in a state config map. Currently the old resource is available in the cache, but this will **not** be guaranteed in the future if we are going to improve caching. So it should not be relied on.
> 
> If the plan is to move to the fabric8 informer, then it provides a method where in case of an update, the old a new resource are provided so I guess, the old resource will still be available somehow. But in general if we are implementing a cache, then I don't really see how the old cached value can disappear :)
> 
> > > you may have to hit the kubernetes api-server for each reconcile loop to get the dependant resource because as today there is no real caching mechanism in the operator sdk except the one for the primary custom resources handled by the operator.
> > 
> > 
> > This is not explicitly supported. But should be fairly easy to create a event source which wraps an informer. (We can maybe try to take a look on this together). So you can access the cached resource from the EventSource / Informer.
> 
> Yes but we need to be aware that by introducing another cache, then we impact also the memory profile of the operator. If you have a large set of resource to handle, having a side resource to track the status would mean you have doubled the number of objects you have in memory. It does not mean that you double the overall memory but still it something to take into account.
> 
> > > in a single reconcile loop you may end up updating both the main and dependant resource so you end up doing at least two calls to the Kubernetes API server
> > 
> > 
> > This is true. So if you would update the status and the "state config map" it would be two api calls.
> > > you may also need to set up a watch on the dependant resource (i.e. if the resource is changed outside the control of the operator you may need to re-write it) so you still need to keep track of the generation or version of the dependant resource in your status and you need to find a way to skip events related to the action you perform on the dependant resource
> > 
> > 
> > Yes you would need the event source for this. This is an interesting point how to handle you own changes to the dependent resources. Interesting thing here is that here you are interested in the status change only of the dependant resource in general. But actually in your case (for the phase) when you are just store the state there you are not necessairly interested on changes on that dependant resource? You just read it next time you would reconcile, or?
> 
> I have to watch the dependant resource because that resource may be changed by someone else (i.e. kubectl edit configmap foo-bar) and in such case I have to implement corrective measures. Of course you can try to have some rbac/opa rules but you see that the whole setup may become much more complex that just letting me to add a simple predicate :)
> 
> > (To be honest, this storing the state in a config map or CR feels also little bit smelly for me :D but I guess this si how it is now on k8s world )

And to add on this, I'm not saying that my way is better and using an external resource is bad but that each operator has its own characteristics and intrinsic complexity depending on the resources it has to deal with so having some flexibility at the sdk level would be - imho - the right thing to do

metacosm:
I don't buy the idea that putting the state in a ConfigMap is better than putting it in the status… At least, in the status, you can have conflict checking and the state lives where it's needed instead of some 3rd party object that needs to be managed.

csviri:
> Yes but we need to be aware that by introducing another cache, then we impact also the memory profile of the operator. If you have a large set of resource to handle, having a side resource to track the status would mean you have doubled the number of objects you have in memory. It does not mean that you double the overall memory but still it something to take into account.

Note that here you not caching necessarily the whole resource. It can be just the phase field in this case. Its an another object, but quite small.  

> If the plan is to move to the fabric8 informer, then it provides a method where in case of an update, the old a new resource are provided so I guess, the old resource will still be available somehow. But in general if we are implementing a cache, then I don't really see how the old cached value can disappear :)

This is an interesting thing yes, so in informers the objects are always cached forever (if understand correctly), that is why you get the old object. The old object is not accessible from k8s api in general. So it's actually a good question if we want to use these informers in this form, since as @metacosm rightfully raised that we don't necessarily want to cache the objects forever (as it is now).

csviri:
At the end I think it is good to have the event sources, which are then can be implemented as wrapper informers or maybe a more lightweight way, with some custom caching or without it.

Maybe we could also think in multiple flavor of event sources. One which actually caches the objects forever an other is not. So users can choose.

csviri:
> I don't buy the idea that putting the state in a ConfigMap is better than putting it in the status… At least, in the status, you can have conflict checking and the state lives where it's needed instead of some 3rd party object that needs to be managed.

Yes, it seems to be quite pragmatic to use status for this.

csviri:
> I don't buy the idea that putting the state in a ConfigMap is better than putting it in the status… At least, in the status, you can have conflict checking and the state lives where it's needed instead of some 3rd party object that needs to be managed.

Yep, reading now this book, https://www.oreilly.com/library/view/production-kubernetes/9781492092292/
and in the operator section they explicily mention that status is meant to record some state and reconcile based on that. They even have the same example , using `.status.phase` as @lburgazzoli

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -97,4 +99,17 @@ default void setConfigurationService(ConfigurationService service) {}
...
+  /**
+   * Allow controllers to filter events before they are provided to the
+   * {@link io.javaoperatorsdk.operator.processing.event.EventHandler}.
+   * </p>
+   * Note that the provided filter is combined with {@link #isGenerationAware()} to compute the
+   * final set of fiolters that should be applied;
+   *
+   * @return
+   */
+  default CustomResourceEventFilter<R> getEventFilter() {
```
csviri, line 112: This should be a list too, or?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -1,18 +1,19 @@
...
-import io.fabric8.kubernetes.client.CustomResource;
-
 public class ControllerConfigurationOverrider<R extends CustomResource<?, ?>> {
 
   private String finalizer;
   private boolean generationAware;
   private final Set<String> namespaces;
   private RetryConfiguration retry;
   private String labelSelector;
+  private CustomResourceEventFilter<R> customResourcePredicate;
```
csviri, line 16: Also a list of predicates. Also the name of the field should correlated with the type IMHO.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilters.java
```diff
@@ -0,0 +1,102 @@
+package io.javaoperatorsdk.operator.processing.event.internal;
+
+import io.fabric8.kubernetes.client.CustomResource;
+
+/**
+ * Convenience implementations of, and utility methods for, {@link CustomResourceEventFilter}.
+ */
+public final class CustomResourceEventFilters {
+
+  private static final CustomResourceEventFilter<CustomResource> USE_FINALIZER =
```
csviri, line 10: Again, this won't work in general. We won't have the old resource in all cases in the future. 
We have to clarify this, we might end up using actually reflectors instead of informers (or just ListWatcher?). So I'm not sure about this interface cotaining the oldObject.

csviri: For example this is why we store the lastGenerationProcessedSuccessfully.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilter.java
```diff
@@ -0,0 +1,58 @@
...
+   * Determines whether the change between the old version of the resource and the new one needs to
+   * be propagated to the controller or not.
+   *
+   * @param configuration the target controller's configuration
+   * @param oldResource the old version of the resource
+   * @param newResource the new version of the resource
+   * @return {@code true} if the change needs to be propagated to the controller, {@code false}
+   *         otherwise
+   */
+  boolean acceptChange(ControllerConfiguration<T> configuration, T oldResource, T newResource);
```
csviri, line 26: We have to discuss this @lburgazzoli @metacosm . The old object should not be here IMHO.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
```diff
@@ -98,49 +98,38 @@ public void eventReceived(Watcher.Action action, T customResource) {
...
-    if (controller.getConfiguration().isGenerationAware()
-        && resource.hasFinalizer(controller.getConfiguration().getFinalizer())) {
-      lastGenerationProcessedSuccessfully.put(
-          KubernetesResourceUtils.getUID(resource), resource.getMetadata().getGeneration());
-    }
-  }
+      // if the finalizer has not been set, the operator logic will add it so there will be
+      // events generated by the operator itself that won't lead to a new generation so such
+      // events should not be filtered
+      final CustomResourceEventFilter<T> finalizerPredicate =
```
csviri, line 110: Why are we checking this with predicate? It would be much simpler and easy to read with just checking directly the field with in a method.

csviri:
Back to the main topic of this PR:

In general I think if we could solve these cases elegantly with this predicates as now, it would be a nice way to implement things like generation awareness. On the other hand I see currently two issues:

- The oldObject (thus the previouse version of CR) is used in the filter api, which might not be awailable in the future. 
- I still don't see the clear vision, if somebody wan't to test a new version of an operator it's typically we can already do in a separate namespace. Remember that if we introduce this API for filters we will have to support it. But without clear vision it could make some confusion. Is there something like this supported in controller runtime?

lburgazzoli:
> Back to the main topic of this PR:
> 
> In general I think if we could solve these cases elegantly with this predicates as now, it would be a nice way to implement things like generation awareness. On the other hand I see currently two issues:
> 
>     * The oldObject (thus the previouse version of CR) is used in the filter api, which might not be awailable in the future.
> 

I'm not sure about this point but it can eventually be done or maybe we can clearly document that the `oldObject` may be null in case of a cacheless controller. Eventually also the generation generation awarness can be delegated to the user by using the `obeservedGeneration` patter (implemented  by a number of Kubernetes controllers) in which, in essence you have to store the generation the controller is act upon in the status.

As a side note, removing the `oldObject` would make the java sdk to drift from the design of the golang sdk (https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/)

>     * I still don't see the clear vision / use cases, if somebody wan't to test a new version of an operator it's typically we can already do in a separate namespace. Remember that if we introduce this API for filters we will have to support it. But without clear vision it could make some confusion. Is there something like this supported in controller runtime?

This issue/PR is not related to testing or canary deployment (that issue was the one about using label selector https://github.com/java-operator-sdk/java-operator-sdk/issues/453) but it is about letting the user be able to install advanced filter to reduce the reconcile events.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
lburgazzoli: Yes, the name here is a leftover from the previous iteration, will fix it

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
lburgazzoli: It can be a list but the it will become an AND so the rationale here is that if you have to chain multiple filers, then you can probably chain them using the `and` / `or` operations provided by the interface so you are free to compose them as you like.

But no problem to make this a list.

csviri:
> I'm not sure about this point but it can eventually be done or maybe we can clearly document that the oldObject may be null in case of a cacheless controller. Eventually also the generation generation awarness can be delegated to the user by using the obeservedGeneration patter (implemented by a number of Kubernetes controllers) in which, in essence you have to store the generation the controller is act upon in the status.

If we document that old object might be null, not sure if that would make sense even to put it there - Or we could also replicate the 3 methods in the filter (Create, Update, Delete)
 
This interface I guess comes from the fact that there is an informer used in the background and the old object is always cached.
(I mean this https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/ )
So not sure if these filters are such tightly coupled with informers, it could be added to rabric8 client, and we could just make an adaptor for it (can't see it now there).

The main problems with the `oldObject`. If the operator is restarted we don't have the old object anymore which is then a little bit smell, although also it's true that on restart we usually want to reconcile.
But on large clusters where there are thousands of custom resources we don't want to cache all the objects since it can happen that there might be just no related events coming for a very long time. In that case it might be more efficient to just read the CR from the k8s API. The plan for the future is that we will provide a configurable cache for this (like Caffein). So in this case the old object won't be available by definition.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
csviri: You right, is might be a too strong assumption to make. Thus that this would be usually an "AND" relationship in those filters.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilters.java
lburgazzoli: To be honest, I'm not sure if the operator should cache the last generation as it should be the controller that has the ultimate responsibility to define when it has successfully processed the latest generation (i.e. based on `.status.observedGeneration` field lot of controller are adding these days).

This is also one of the reason for which I think it would be very nice to add the option to define custom predicates.

lburgazzoli:
> > I'm not sure about this point but it can eventually be done or maybe we can clearly document that the oldObject may be null in case of a cacheless controller. Eventually also the generation generation awarness can be delegated to the user by using the obeservedGeneration patter (implemented by a number of Kubernetes controllers) in which, in essence you have to store the generation the controller is act upon in the status.
> 
> If we document that old object might be null, not sure if that would make sense even to put it there - Or we could also replicate the 3 methods in the filter (Create, Update, Delete), (if we are actually going to use informers in the background.)
> 
> This interface I guess comes from the fact that there is an informer used in the background and the old object is always cached.
> (I mean this [sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering](https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/) )
> So not sure if these filters are such tightly coupled with informers, it could be added to rabric8 client, and we could just make an adaptor for it (can't see it there atm).
> 
> The main problems with the `oldObject`. If the operator is restarted we don't have the old object anymore which is then a little bit smell, although also it's true that on restart we usually want to reconcile.
> 
> But on large clusters where there are thousands of custom resources **we don't want to cache all the objects** since it can happen that there might be just no related events coming for a very long time. In that case it might be more efficient to just read the CR from the k8s API. The plan for the future is that we will provide a configurable cache for this (like Caffein). So in this case the old object won't be available by definition.
> We should have actually a discussion about from above, also if we should have multiple options for caches etc, to cache or not cache the objects. So very interested in your opinion @lburgazzoli also. But the design of this predicates also depends on that outcome.

Fine with me, maybe we can have an ad-hoc meeting for this discussion ?

csviri:
> Fine with me, maybe we can have an ad-hoc meeting for this discussion ?

Definitely, that would be the best IMHO.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilters.java
csviri: I agree. With the new caching approach this should go away too, and have it in status.

Just maybe to think about this to do it seamlessly, so how the framework could do for the user transparently, with minimal effort. Best would be to create an issue for it, where we discuss the possibilities. (Like if there is an observedGeneration field in status we handle it automatically, or such)

lburgazzoli: Thinking about this aloud, I wonder if it would be better to move the filtering logic to the controller itself in the form of an interfaces, like:

```java
interface ResourceController<T extends CustomResource<Spec, Status>> {
    default boolean process(T resource) {
        return true;
    }
}
```

Then we can add some special `ResourceController` interface, like

```java
interface GenerationAwareResourceController<T extends GenerationAwareCustomResource<Spec, Status>> {
    @Override
    default boolean process(T resource) {
        return resource.status.observedGeneration < resource.meta.generation ;
    }
}
```

lburgazzoli:
Converted to a draft until we get consensus

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventSource.java
metacosm: So that the same logic is applied everywhere it's needed instead of having variations of the same code.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilter.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param <T> the type of custom resources handled by this filter
+ */
+@FunctionalInterface
+public interface CustomResourceEventFilter<T extends CustomResource> {
+
+  /**
+   * Determines whether the change between the old version of the resource and the new one needs to
+   * be propagated to the controller or not.
+   *
```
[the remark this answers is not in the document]
lburgazzoli: done

lburgazzoli:
@csviri @metacosm pr is now ready for review

csviri:
@lburgazzoli will take a look today, after finished my current issue

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/CustomResourceCache.java
```diff
@@ -44,19 +44,21 @@ public CustomResourceCache(ObjectMapper objectMapper, Metrics metrics) {
...
     try {
       lock.lock();
       final var uid = getUID(resource);
       if (predicate.test(resources.get(uid))) {
         if (passthrough != predicate) {
           log.trace("Update cache after condition is true: {}", getName(resource));
         }
-        resources.put(uid, resource);
+        // defensive copy
+        resources.put(getUID(resource), clone(resource));
```
csviri, line 61: why are we cloning here?

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/CustomResourceCache.java
lburgazzoli: The reason is that as example, if for some reason the process that applies the finalizer fails when updating the object, then the cache would still reporting the object as having the finalizer which is not correct.

it probably won't matter so if it should be removed, I ca do it in a subsequent pr

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/CustomResourceEventFilters.java
```diff
@@ -0,0 +1,163 @@
...
+      (configuration, oldResource, newResource) -> {
+        if (configuration.useFinalizer()) {
+          final var finalizer = configuration.getFinalizer();
+          boolean oldFinalizer = oldResource == null || oldResource.hasFinalizer(finalizer);
```
csviri, line 14: It would be good to write unit tests for these methods, since there is non-trivial logic.

Issue #64, opened 2020-03-02
Micrometer Integration


csviri:
Integrate micrometer to generate metrics about usage / execution.

csviri:
https://micrometer.io/

csviri:
Would be great to gather here the metrics we are interested in. I guess typical is:
- number of controller executions
- number of events processed
- controller execution durations
- number of custom resource managed

jmrodri:
@csviri How would something like this integrate with prometheus on kubernetes? Most operators output their metrics to be scraped by a prometheus service to expose their items to the console.

csviri:
@jmrodri The idea is that we would make it very easy to integrate with a metrics endpoint, thus to expose the metrics to http.
Prometheus is supported by micrometer so I guess that would not be a hassle. 

But ideally we would not be dependent on any framework (spring / quarkus / micronaut) but would make it very easy to integrate with them. So the controller is what exposes the metrics will be native to the desired framework.

csviri:
An important part of this would be extensibility. So anybody should be able to add own metrics. The simplest way to do that (the metrics about the Custom Resource processing ) is just expose the micrometer api to the controller. So the user can add custom metrics directly to the controller implementation. 
What we should provide out of the box is probably metrics that are not necessary available from the controllers. Like threading related metrics, maybe some event handling. But also some basic standard metrics mentioned above that is makes sense for everybody.

metacosm:
Another project to look into is https://opentelemetry.io/

jmrodri:
Another resources mentioned during community meeting: https://quarkus.io/guides/micrometer

metacosm:
Seems like Micrometer is indeed the framework of choice to deal with metrics in Quarkus. This shouldn't prevent us from looking at opentelemetry but it seems more focused on tracing than metrics.

laxmikantbpandhare:
Yes @metacosm - Micrometer is better choice as compare to OpenTelemetry. There are few differences like OpentTelemetry does not support dedicated times type, distribution summery type and also it does not have any naming conventions. OpenTelemetry is more focused on tracing  while micrometer is built specifically for getting the metrics/usage of an application. **OpenTelemetry would be useful to measure across many Micro-services as they hand off requests and you need to trace behavior, while Micrometer would be more useful for tracking a single application's performance and health.**

laxmikantbpandhare:
[...]

PR #486, opened 2021-08-06
PR for - Micrometer Integration - #64

laxmikantbpandhare:
Working on Timer and other remaining part. @metacosm and @csviri - Could you please verify the code for **Counter** and its implemented methods.


jmrodri:
@laxmikantbpandhare will this tie into the `deleteResource` method for decrementing the metric? https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java#L28-L30

How will we ensure the increment occurs for the metrics we plan to support? Or is that for another follow on PR?

jmrodri:
I think that we should create a metrics  package in the `operator-framework-core/src/main/java/io/javaoperatorsdk/operator` or in `operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api` to house the metrics stuff. And I think CounterDetails should be in implementation detail of what we're trying to do. The issue lists a set of metrics we would like to expose by default. 

- number of controller executions
- number of events processed
- controller execution durations
- number of custom resource managed
- retry related metrics

So I think we should have a `metrics.ControllerExecution` which is an instance of a `CounterDetails`. We can then increment it for the user whenever a reconcile occurs. Similar to how controller-runtime is doing it here:  https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/metrics/metrics.go#L30-L33

```go
	ReconcileTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
		Name: "controller_runtime_reconcile_total",
		Help: "Total number of reconciliations per controller",
	}, []string{"controller", "result"})
```

As you see in controller-runtime, there are a series of default structures that represent the default metrics they want to expose. I could see us having 5 classes representing the metrics we want.

Then in the `deleteResource` we would decrement the ControllerExecution.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/utils/CounterDetails.java
```diff
@@ -0,0 +1,40 @@
+package io.javaoperatorsdk.operator.utils;
+
+public class CounterDetails {
+
+    int counter;
```
adam-sandor, line 5: do these fields need to be package-private instead of private?

laxmikantbpandhare:
> @laxmikantbpandhare will this tie into the `deleteResource` method for decrementing the metric? https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java#L28-L30
> 
> How will we ensure the increment occurs for the metrics we plan to support? Or is that for another follow on PR?

@jmrodri - I am still looking into the code of `java-operator-sdk` for **delete resource** functionality. I will add those changes as part of this PR only.

laxmikantbpandhare:
> @laxmikantbpandhare will this tie into the `deleteResource` method for decrementing the metric? https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/ResourceController.java#L28-L30
> 
> How will we ensure the increment occurs for the metrics we plan to support? Or is that for another follow on PR?

@jmrodri - We will not decrement any counter as **operator-sdk** does not decrement anything as such. After delete event, the reconciliation process happens thats why need to increment controller execution counter. The **operator-sdk** has two counters, **succeeded** and **failed**. If reconciliation process fails then **failed** counter increases. We will follow the same logic for our metrics part in **java-operator-sdk**.

jmrodri:
> @jmrodri - We will not decrement any counter as **operator-sdk** does not decrement anything as such. After delete event, the reconciliation process happens thats why need to increment controller execution counter. The **operator-sdk** has two counters, **succeeded** and **failed**. If reconciliation process fails then **failed** counter increases. We will follow the same logic for our metrics part in **java-operator-sdk**.

You are correct we only increment. Thanks.

jmrodri:
@laxmikantbpandhare also, this code should live in a `metrics` package.

laxmikantbpandhare:
> @laxmikantbpandhare also, this code should live in a `metrics` package.

@jmrodri - Sure, I will create `metrics` package and add those changes in it.

jmrodri:
Also remember that PRs should follow the conventional commit: https://github.com/java-operator-sdk/java-operator-sdk/blob/master/CONTRIBUTING.md#pull-request-process

laxmikantbpandhare:
> Also remember that PRs should follow the conventional commit: https://github.com/java-operator-sdk/java-operator-sdk/blob/master/CONTRIBUTING.md#pull-request-process

Modified commit as a conventional commit.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/utils/CounterDetails.java
laxmikantbpandhare: checking on this, we can make those field as a private.

metacosm:
I've converted this PR to draft since it seems quite far from being finished. I cannot really review it in its current state because I need to see:
- which types of metrics will be provided
- how they will integrate with the SDK code
- which data will be recorded

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/utils/CounterDetails.java
metacosm: I don't currently see what's the point of this class since it doesn't appear to be used but I would assume that it should be immutable: all fields should be private and final and there shouldn't be any setters (i.e. it shouldn't be possible to modify an instance, after it's been created)

laxmikantbpandhare:
@metacosm - I showed demo of metrics on our weekly meeting. Attached screenprint for your reference. 

Screen Shot 2021-08-17 at 12 00 21 AM

The current changes are just utility and I am planning to call those methods from reconciliation process.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/utils/CounterDetails.java
laxmikantbpandhare: The counter details class will be used when we call/use these utility methods. You can refer to sample **memcached-quarkus-operator** where I used this - https://github.com/laxmikantbpandhare/memcached-quarkus-operator/blob/main/src/main/java/com/example/MemcachedController.java

metacosm:
On a design note: I'd like to see something similar to what is seen in https://github.com/ebullient/monster-combat/blob/main/core/src/main/java/dev/ebullient/dnd/combat/client/CombatMetrics.java. All metrics-related code is confined to a single class so that it doesn't "infect" the code and we can tweak what gets recorded and how in a single spot. The SDK code would call method of this class with application-specific results (e.g. `UpdateControl`) as opposed to things related to metrics, though we could add metrics-related information to these objects if needed. 
Also, we should only depend on `micrometer-core`, not `micrometer-registry-prometheus` so that we're not dependent on a given backend but users can choose whichever backend they want.

laxmikantbpandhare:
> On a design note: I'd like to see something similar to what is seen in https://github.com/ebullient/monster-combat/blob/main/core/src/main/java/dev/ebullient/dnd/combat/client/CombatMetrics.java. All metrics-related code is confined to a single class so that it doesn't "infect" the code and we can tweak what gets recorded and how in a single spot. The SDK code would call method of this class with application-specific results (e.g. `UpdateControl`) as opposed to things related to metrics, though we could add metrics-related information to these objects if needed.
> Also, we should only depend on `micrometer-core`, not `micrometer-registry-prometheus` so that we're not dependent on a given backend but users can choose whichever backend they want.

Yes, agree on this. I will update the code accordingly.

laxmikantbpandhare:
> On a design note: I'd like to see something similar to what is seen in https://github.com/ebullient/monster-combat/blob/main/core/src/main/java/dev/ebullient/dnd/combat/client/CombatMetrics.java. All metrics-related code is confined to a single class so that it doesn't "infect" the code and we can tweak what gets recorded and how in a single spot. The SDK code would call method of this class with application-specific results (e.g. `UpdateControl`) as opposed to things related to metrics, though we could add metrics-related information to these objects if needed.
> Also, we should only depend on `micrometer-core`, not `micrometer-registry-prometheus` so that we're not dependent on a given backend but users can choose whichever backend they want.

@metacosm - I completely removed CounterDetails class and added entire login in only one class. Please review the PR. About the second suggestion, `micrometer-core` contains only `SimpleMeterRegistry` and it is the simplest form of the registry. As per me, `MeterRegistry` is the best option. Below links explains more about it.

Links - 
https://micrometer.io/docs/installing
https://www.baeldung.com/micrometer#3-registry

metacosm:
We're not going to provide a registry implementation in the SDK: this will be up to users to decide which one to use. 
More precisely: the SDK will not be doing any actual metering, just provide the infrastructure for it to happen when a user selects a backend and activates it in their operator. `MeterRegistry` is the abstract class we should be coding against, not any implementation.

laxmikantbpandhare:
> We're not going to provide a registry implementation in the SDK: this will be up to users to decide which one to use.
> More precisely: the SDK will not be doing any actual metering, just provide the infrastructure for it to happen when a user selects a backend and activates it in their operator. `MeterRegistry` is the abstract class we should be coding against, not any implementation.

Ah! sorry it was my confusion. I updated dependency as said by you in `Quarkus Memcached Operator` and was getting `UnSatisfiedDepedency` for `micrometer-core`. It was my mistake. I updated the`pom.xml` file. Please review and let me know your comments.

adam-sandor:
Sorry @laxmikantbpandhare but I don't see where the actual functionality is supposed to be here... This is just an adding of a dependency and a simple counter class. Don't you want to actually use this in the code and count some things? Do you have a plan for what you want to measure?

laxmikantbpandhare:
> Sorry @laxmikantbpandhare but I don't see where the actual functionality is supposed to be here... This is just an adding of a dependency and a simple counter class. Don't you want to actually use this in the code and count some things? Do you have a plan for what you want to measure?

Exactly, as I said on the discord channel on operator-sdk-sync. Need some clarification on this.

metacosm:
Also note that your code is not properly formatted which is why CI fails.

metacosm:
@laxmikantbpandhare can you make your branch writable by maintainers, please? PRs should be opened in a way that maintainers can change stuff on it.

metacosm:
I've created #497 to show how I think we should go about the integration… Lots of work still to be done but should provide a basis for further discussions.

laxmikantbpandhare:
> @laxmikantbpandhare can you make your branch writable by maintainers, please? PRs should be opened in a way that maintainers can change stuff on it.

I already tick marked `Allow edits by maintainers`.

laxmikantbpandhare:
> I've created #497 to show how I think we should go about the integration… Lots of work still to be done but should provide a basis for further discussions.

@metacosm @adam-sandor @csviri - I created similar method for `deleteResource()`. Please review and let me know your comments.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Metrics.java
```diff
@@ -0,0 +1,152 @@
...
+      ResourceController controller,
+      ControllerConfiguration configuration,
+      CustomResource resource,
+      Context context) {
+    final var name = configuration.getName();
+    final var timer =
+        registry.timer("operator.sdk.controllers.execution.delete", "controller", name);
+    try {
+      final var result = timer.record(() -> controller.deleteResource(resource, context));
+      String successType = "cr";
```
metacosm, line 70: Do you really think that this success type makes sense for the delete operation?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -153,7 +158,12 @@ private PostExecutionControl handleDelete(R resource, Context<R> context) {
         getName(resource),
         getVersion(resource));
     // todo: this is be executed in a try-catch statement, in case this fails
```
metacosm, line 160: This comment is now obsolete.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -153,7 +158,12 @@ private PostExecutionControl handleDelete(R resource, Context<R> context) {
...
-    DeleteControl deleteControl = controller.deleteResource(resource, context);
+    //    DeleteControl deleteControl = controller.deleteResource(resource, context);
```
metacosm, line 161: Why keep this?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/EventDispatcherTest.java
```diff
@@ -141,7 +147,7 @@ void callDeleteOnControllerIfMarkedForDeletionWhenNoFinalizerIsConfigured() {
     eventDispatcher.handleExecution(
         executionScopeWithCREvent(Watcher.Action.MODIFIED, testCustomResource));
 
-    verify(controller).deleteResource(eq(testCustomResource), any());
+    verify(controller, never()).deleteResource(eq(testCustomResource), any());
```
metacosm, line 150: This isn't correct. Delete should be called.

metacosm:
On a strictly process-oriented basis: please do not request reviews for each commit. Finish the PR as you think it should be done and then we will be able to review it. Also, please leave the PR in draft mode until it's ready for review and merging, which isn't the case as this stage.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
laxmikantbpandhare: Removed this comment.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
laxmikantbpandhare: yes, removed this comment.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Metrics.java
laxmikantbpandhare:
```
      String successType = "notDelete";
      if (result == DeleteControl.DEFAULT_DELETE) {
        successType = "delete";
      }
 ```

I modified like this.

File: operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/EventDispatcherTest.java (hunk 5 above)
Comment (laxmikantbpandhare) [reply operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/EventDispatcherTest.java]: Yes @metacosm , this is the test case where we are ignoring finalizer configuration. I modified the `configureToNotUseFinalizer()` method as below mentioned setup was missing when `configureToNotUseFinalizer()` creates a new `configuration`.

```
    when(configuration.getName()).thenReturn("EventDispatcherTestController");
    when(configService.getMetrics()).thenReturn(Metrics.NOOP);
    when(configuration.getConfigurationService()).thenReturn(configService);
```

I updated the PR, please let me know if anything is missing here.

**Comment by laxmikantbpandhare** [comment]
> On a strictly process-oriented basis: please do not request reviews for each commit. Finish the PR as you think it should be done and then we will be able to review it. Also, please leave the PR in draft mode until it's ready for review and merging, which isn't the case as this stage.

@metacosm - This process is completely new to us. I never requested review for each commit. Moving this in draft is something I never did. Is there any document where we have this entire process mentioned or it will be better if we elaborate during our weekly meeting.

File: operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Metrics.java [hunk 6]
```diff
@@ -0,0 +1,152 @@
...
+import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
+import io.micrometer.core.instrument.*;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
+import io.micrometer.core.instrument.distribution.pause.PauseDetector;
+import io.micrometer.core.instrument.noop.*;
+import java.util.concurrent.TimeUnit;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToLongFunction;
+
+public class Metrics {
```
csviri, line 17: For me this whole construct while wrapping the logic in a Metrics class is little bit smelly, not sure if there is a better way. But will take a look soon.

laxmikantbpandhare: Initially, I wrote a separate class without wrapping it. @metacosm - Chris asked to have it like above wrapping the logic.

metacosm: What's wrong with wrapping the logic, @csviri? More than wrapping, it's about measuring things that make sense from the operator's perspective as opposed to calling more metrics-oriented methods. Another aspect is that all metrics code is in a single spot which makes it easier to figure out what's measured and change it if needed.

csviri: That the metrics code is in one place that is good. That we are mixing the application logic with metrics is not that nice. Those should be as loosely coupled as possible IMHO.

csviri: I don't say I have a better answer for this now, just mentioning that design principle. So something to think about.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Metrics.java
```diff
@@ -29,7 +29,10 @@ public Metrics(MeterRegistry registry) {
       Context<R> context) {
     final var name = configuration.getName();
     final var timer =
-        registry.timer("operator.sdk.controllers.execution.createOrUpdate", "controller", name);
+        Timer.builder("Controller Executions Timer")
```
metacosm, line 32: Why was only this timer changed? What happened to the name and the tags? Also, please strive to create atomic, functional commits. A commit should stand by itself with a single functionality added / implemented as much as possible. As this commits stands, it's missing the delete counterpart which should be part of the same commit.

laxmikantbpandhare: Yes, I will add same naming conventions and update it for Delete method as well.

PR #491, opened 2021-08-13
feat: delay the registration of controller till the operator is started

### LINKED ISSUE #488: Prevent Operator star/stop procedure be performed twice
As today the Operator start/stop methods do not have any barrier to avoid the methods to be invoked more than once and as example, since the introduction of the PR here https://github.com/quarkiverse/quarkus-operator-sdk/blob/main/runtime/src/main/java/io/quarkiverse/operatorsdk/runtime/StartupListener.java, I've observed that my operator did invoke start twice because I also had a startup listener to implement the same functionality.

I have now fixed my code so id does not happen anymore but I feel a barrier on the start/stop methods could be useful in any case.

### LINKED ISSUE #489: [discussion] perform Operator::register when invoking Operator::start
As today, once you call Operator::register, then watches are put in place and the reconcile loop may chick in before the Operator::start method is invoked. 

I wonder if it would be better to move the registration steps to the startup phase so the lifecycle of the operator and the controller would be predictable and consistent.

lburgazzoli:
@metacosm @csviri This is a POC of the discussion we had on the weekly meeting about the lifecycle of controllers vs operator' start/stop method.

Fixes #488 
Fixes #489

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -18,16 +18,21 @@ import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("rawtypes")
 public class Operator implements AutoCloseable {
-
   private static final Logger log = LoggerFactory.getLogger(Operator.class);
   private final KubernetesClient k8sClient;
   private final ConfigurationService configurationService;
   private final List<Closeable> closeables;
+  private final Object lock;
+  private final List<ControllerRef> controllers;
+  private volatile boolean started;
 
   public Operator(KubernetesClient k8sClient, ConfigurationService configurationService) {
     this.k8sClient = k8sClient;
     this.configurationService = configurationService;
     this.closeables = new ArrayList<>();
+    this.lock = new Object();
+    this.controllers = new ArrayList<>();
+    this.started = false;
 
     Runtime.getRuntime().addShutdownHook(new Thread(this::close));
   }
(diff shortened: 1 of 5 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -104,9 +157,10 @@ public void close() {
    * @param <R> the {@code CustomResource} type associated with the controller
    * @throws OperatorException if a problem occurred during the registration process
    */
-  public <R extends CustomResource> void register(
+  private <R extends CustomResource> void startController(
       ResourceController<R> controller, ControllerConfiguration<R> configuration)
       throws OperatorException {
+
     final var existing = configurationService.getConfigurationFor(controller);
     if (existing == null) {
```
lburgazzoli, line 165: @metacosm @csviri Not strictly related to this PR but is this check correct ? I'd expect this condition to be evaluated only if the given `configuration` parameter is null.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -92,6 +119,32 @@ public void close() {
...
+   * @throws OperatorException if a problem occurred during the registration process
+   */
+  public <R extends CustomResource> void register(
+      ResourceController<R> controller, ControllerConfiguration<R> configuration)
+      throws OperatorException {
+    synchronized (lock) {
+      if (!started) {
+        this.controllers.add(new ControllerRef(controller, configuration));
+      } else {
+        this.controllers.add(new ControllerRef(controller, configuration));
```
lburgazzoli, line 142: The `ControllerRef` object is added to the queue in case a scenario like start, stop, start would be supported.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: @lburgazzoli this check is correct: the intent is that the provided configuration is an override of the existing configuration. To that end, the controller must already been known to the configuration service, which is why we first check if we have an existing configuration which is used if no overridden version is provided.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -195,4 +249,14 @@ private void throwMissingCRDException(String crdName, String specVersion, String
     }
     return false;
   }
+
+  private static class ControllerRef {
```
metacosm, line 253: This should be parameterized by the CR type so that the `ResourceController` and `ControllerConfiguration` instances use the same type…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -45,10 +43,9 @@
     }
   }
 
+  @SuppressWarnings({"rawtypes", "unchecked"})
   public <R extends CustomResource<?, ?>> DefaultEventSourceManager(
-      ResourceController<R> controller,
-      ControllerConfiguration<R> configuration,
-      MixedOperation<R, KubernetesResourceList<R>, Resource<R>> client) {
+      ResourceController controller, ControllerConfiguration configuration, MixedOperation client) {
```
metacosm, line 48: I don't think that this change is needed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -45,10 +43,9 @@
     }
   }
 
+  @SuppressWarnings({"rawtypes", "unchecked"})
```
metacosm, line 46: Unneeded without the following change.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
lburgazzoli: I'm not sure how it should be done as the list of ControllerRef would need to be on the base type in any case. Unfortunately, the usage of generic in the project is a little bit confused an inconsistent so the only way I found to make all the methods happy were to remove generics. 

But I'm happy to change this if you help me understanding the right way of doing it.

metacosm: I agree that the generic usage is inconsistent in the project. Let's address this in a subsequent PR.

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
lburgazzoli: thx, can you re-trigger the CI ?

PR #499, opened 2021-08-19
feat: option to limit CRs the operator watches

### LINKED ISSUE #453: Option to limit CRs the operator watches

lburgazzoli:
Fixes #453


lburgazzoli:
@csviri @metacosm this is a POC for #453 and about using the mocked kubernetes server for testing

lburgazzoli:
@csviri @metacosm any feedback on this PR ?

metacosm:
I started looking into it. Will review today.

csviri:
Also will try to take a look, but little busy from now on today :-/

metacosm (approved):

PR #503, opened 2021-08-20
chore: refactor dependency management and settle on log4j2 as logging impl

lburgazzoli:
This PR aims to:
- move dependency management to the main pom
- use properties to define artifact versions
- explicitly set maven plugins version instead of rely on defaults (i.e. compiler plugin was 3.1, now 3.8.1)
- use log4j2 as the default logging impl and get rid of logback and sl4j-simple


metacosm (approved):

csviri (approved):

metacosm:
Replaced by #509 to be able to fix conflicts.

PR #506, opened 2021-08-20
chore: add maven wrapper for bettere reproducibility

lburgazzoli:
This is to be sure the build is independent from the maven version provided by the GitHub Action environment

### DIFF
--- a/.github/workflows/master-snapshot-release.yml
+++ b/.github/workflows/master-snapshot-release.yml
@@ -1,4 +1,8 @@
 name: Test & Release Snapshot to Maven Central
+
+env:
+  MAVEN_ARGS: -V -ntp -e
+
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
   cancel-in-progress: true
@@ -28,7 +32,7 @@ jobs:
           distribution: ${{ matrix.distribution }}
           java-version: ${{ matrix.java }}
       - name: Run unit tests
-        run: mvn -B test --file pom.xml
+        run: ./mvnw ${MAVEN_ARGS} -B test --file pom.xml
       - name: Set up Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
         with:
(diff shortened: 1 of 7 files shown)


.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,2 @@
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip
```
metacosm, line 1: Why use this old Maven version?

.github/workflows/master-snapshot-release.yml
```diff
@@ -1,4 +1,8 @@
 name: Test & Release Snapshot to Maven Central
+
+env:
+  MAVEN_ARGS: -V -ntp -Dhttp.keepAlive=false -e
```
metacosm, line 4: What's the reason for using these args specifically?

.mvn/wrapper/maven-wrapper.properties
lburgazzoli: Oh, I think this is a copy an paste from my history

.github/workflows/master-snapshot-release.yml
lburgazzoli: To have better troubleshooting info:
-V is about dumping information about maven and java
-ntp is to avoid printing the download of artifacts
-e is about having some better error message if maven is failing

-Dhttp.keepAlive=false is a copy and paste but was used to fix some issue we had sometimes with the way the GitHub Actions environment is set up (I can remove it)

.mvn/wrapper/maven-wrapper.properties
lburgazzoli: fixed

.github/workflows/master-snapshot-release.yml
lburgazzoli: `-Dhttp.keepAlive=false` removed

metacosm (approved):

Issue #496, opened 2021-08-18
Use `ControllerRef` internally instead of `ResourceController`


metacosm:
The idea is that there are many spots where we would like to access the controller's configuration without having to first inject the `ConfigurationService` to be able to retrieve it. An alternative would be to add a `getConfiguration` method to `ResourceController` but this idea has already been discussed and rejected on the ground that this would put undue burden on the `ResourceController`'s implementer.

PR #507, opened 2021-08-20
feat: create a junit5 extension

lburgazzoli:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -253,13 +258,21 @@ private void throwMissingCRDException(String crdName, String specVersion, String
     return false;
   }
 
-  private static class ControllerRef {
+  public static class ControllerRef {
```
metacosm, line 261: I plan on making this class more widely used, cf #496

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -253,13 +258,21 @@ private void throwMissingCRDException(String crdName, String specVersion, String
...
     public final ResourceController controller;
     public final ControllerConfiguration configuration;
 
     public ControllerRef(ResourceController controller, ControllerConfiguration configuration) {
       this.controller = controller;
       this.configuration = configuration;
     }
+
+    public ResourceController getController() {
```
metacosm, line 270: Why add getters if the instance variables are public final… Use one or the other :)

csviri:
> @metacosm @csviri I had some spare time and I have started hacking a JUnit 5 extension to make it easy to run integration tests.
> 
> It is quire basic at the moment but my goal is:
> 
> * to make it possible to either use a real Kubernetes cluster or a mocked one provided by the fabric8
> * if possible run test in parallel as each test runs in its own namespace
> 
> Let me know if you think it is work to create such JUnit 5 extensions

Maybe we should discuss this issue, running integration/functional tests in parallel would boost the speed of the test. However there is a benefit to have tests isolated completely (the runs), it's already quite hard analyse and find the problem in those tests in case of failure.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
lburgazzoli: I think this is a leftover when using the `ResourceController` as part of some stream operation so getter/setter where require for method references

lburgazzoli:
> > @metacosm @csviri I had some spare time and I have started hacking a JUnit 5 extension to make it easy to run integration tests.
> > It is quire basic at the moment but my goal is:
> > 
> > * to make it possible to either use a real Kubernetes cluster or a mocked one provided by the fabric8
> > * if possible run test in parallel as each test runs in its own namespace
> > 
> > Let me know if you think it is work to create such JUnit 5 extensions
> 
> Maybe we should discuss this issue, running integration/functional tests in parallel would boost the speed of the test. However there is a benefit to have tests isolated completely (the runs), it's already quite hard analyse and find the problem in those tests in case of failure.

The tests are isolated as each test runs it its own namespace so there should not be any interference by many tests running in parallel. We can make the parallel run something we enable only in CI

lburgazzoli:
superseded by https://github.com/java-operator-sdk/java-operator-sdk/pull/545

PR #517, opened 2021-08-26
Introduce `ConfiguredController` concept, decouple Metrics code from business logic

### LINKED ISSUE #496: Use `ControllerRef` internally instead of `ResourceController`

metacosm:
The idea is to expand on the `ControllerRef` class to associate a controller with its configuration and propagate this entity instead of passing the controller, the configuration and very often the kube client to each object that needs that information. 
See commit message on https://github.com/java-operator-sdk/java-operator-sdk/commit/1c47ecf61f150bd241980ffff0c20afcfb003c57 for more details.

I've also taken the opportunity to clean up things from the Metrics class so that it's better decoupled from the business logic.

Fixes #496.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -15,6 +15,9 @@ import io.javaoperatorsdk.operator.api.ResourceController;
 import io.javaoperatorsdk.operator.api.config.ConfigurationService;
 import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
 import io.javaoperatorsdk.operator.processing.ConfiguredController;
+import io.javaoperatorsdk.operator.processing.DefaultEventHandler;
+import io.javaoperatorsdk.operator.processing.DefaultEventHandler.EventMonitor;
+import io.javaoperatorsdk.operator.processing.event.Event;
 
 @SuppressWarnings("rawtypes")
 public class Operator implements AutoCloseable {
@@ -24,25 +27,29 @@ public class Operator implements AutoCloseable {
   private final Object lock;
   private final List<ConfiguredController> controllers;
   private volatile boolean started;
-  private final Metrics metrics;
 
-  public Operator(
-      KubernetesClient k8sClient, ConfigurationService configurationService, Metrics metrics) {
+  public Operator(KubernetesClient k8sClient, ConfigurationService configurationService) {
     this.k8sClient = k8sClient;
     this.configurationService = configurationService;
     this.lock = new Object();
     this.controllers = new ArrayList<>();
     this.started = false;
-    this.metrics = metrics;
+    DefaultEventHandler.setEventMonitor(new EventMonitor() {
+      @Override
+      public void processedEvent(String uid, Event event) {
+        configurationService.getMetrics().timeControllerEvents();
+      }
+
+      @Override
+      public void failedEvent(String uid, Event event) {
+        configurationService.getMetrics().timeControllerRetry();
+      }
+    });
   }
 
   /** Adds a shutdown hook that automatically calls {@link #close()} when the app shuts down. */
   public void installShutdownHook() {
     Runtime.getRuntime().addShutdownHook(new Thread(this::close));
-  }
-
-  public Operator(KubernetesClient k8sClient, ConfigurationService configurationService) {
-    this(k8sClient, configurationService, Metrics.NOOP);
   }
 
   public KubernetesClient getKubernetesClient() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -25,7 +25,6 @@ public class Operator implements AutoCloseable {
   private static final Logger log = LoggerFactory.getLogger(Operator.class);
   private final KubernetesClient k8sClient;
   private final ConfigurationService configurationService;
-  private final ReentrantLock lock = new ReentrantLock();
   private final List<ConfiguredController> controllers = new LinkedList<>();
   private volatile boolean started = false;
 
@@ -65,44 +64,39 @@ public class Operator implements AutoCloseable {
    */
   @SuppressWarnings("unchecked")
   public void start() {
-    try {
-      lock.lock();
-      if (started) {
-        return;
-      }
-      if (controllers.isEmpty()) {
-        throw new OperatorException("No ResourceController exists. Exiting!");
-      }
-
-      final var version = configurationService.getVersion();
-      log.info(
-          "Operator SDK {} (commit: {}) built on {} starting...",
-          version.getSdkVersion(),
-          version.getCommit(),
-          version.getBuiltTime());
-
-      log.info("Client version: {}", Version.clientVersion());
-      try {
-        final var k8sVersion = k8sClient.getVersion();
-        if (k8sVersion != null) {
-          log.info("Server version: {}.{}", k8sVersion.getMajor(), k8sVersion.getMinor());
-        }
-      } catch (Exception e) {
-        final String error;
-        if (e.getCause() instanceof ConnectException) {
-          error = "Cannot connect to cluster";
-        } else {
-          error = "Error retrieving the server version";
-        }
-        log.error(error, e);
-        throw new OperatorException(error, e);
-      }
-
-      controllers.parallelStream().forEach(ConfiguredController::start);
-      started = true;
-    } finally {
-      lock.unlock();
+    if (started) {
+      return;
     }
+    if (controllers.isEmpty()) {
+      throw new OperatorException("No ResourceController exists. Exiting!");
+    }
+
+    final var version = configurationService.getVersion();
+    log.info(
+        "Operator SDK {} (commit: {}) built on {} starting...",
+        version.getSdkVersion(),
+        version.getCommit(),
+        version.getBuiltTime());
+
+    log.info("Client version: {}", Version.clientVersion());
+    try {
+      final var k8sVersion = k8sClient.getVersion();
+      if (k8sVersion != null) {
+        log.info("Server version: {}.{}", k8sVersion.getMajor(), k8sVersion.getMinor());
+      }
+    } catch (Exception e) {
+      final String error;
+      if (e.getCause() instanceof ConnectException) {
+        error = "Cannot connect to cluster";
+      } else {
+        error = "Error retrieving the server version";
+      }
+      log.error(error, e);
+      throw new OperatorException(error, e);
+    }
+
+    controllers.parallelStream().forEach(ConfiguredController::start);
+    started = true;
   }
 
   /** Stop the operator. */
@@ -111,25 +105,20 @@ public class Operator implements AutoCloseable {
     log.info(
         "Operator SDK {} is shutting down...", configurationService.getVersion().getSdkVersion());
 
-    try {
-      lock.lock();
-      if (!started) {
-        return;
-      }
-
-      this.controllers.parallelStream().forEach(closeable -> {
-        try {
-          log.debug("closing {}", closeable);
-          closeable.close();
-        } catch (IOException e) {
-          log.warn("Error closing {}", closeable, e);
-        }
-      });
-
-      started = false;
-    } finally {
-      lock.unlock();
+    if (!started) {
+      return;
     }
+
+    this.controllers.parallelStream().forEach(closeable -> {
+      try {
+        log.debug("closing {}", closeable);
+        closeable.close();
+      } catch (IOException e) {
+        log.warn("Error closing {}", closeable, e);
+      }
+    });
+
+    started = false;
   }
 
   /**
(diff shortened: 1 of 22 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -62,10 +65,14 @@ public ConfigurationService getConfigurationService() {
    */
   @SuppressWarnings("unchecked")
   public void start() {
-    synchronized (lock) {
+    try {
+      lock.lock();
```
csviri, line 69: Why is this method synchronized at all? Do we except this to be called from a multi-threaded environment? I't not enough just to say in docs that this is not thread safe, call it from single thread?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -188,39 +173,13 @@ public void close() {
...
-      final var crdName = configuration.getCRDName();
-      final var specVersion = "v1";
-
-      // check that the custom resource is known by the cluster if configured that way
-      final CustomResourceDefinition crd; // todo: check proper CRD spec version based on config
-      if (configurationService.checkCRDAndValidateLocalModel()) {
-        crd = k8sClient.apiextensions().v1().customResourceDefinitions().withName(crdName).get();
-        if (crd == null) {
-          throwMissingCRDException(crdName, specVersion, controllerName);
+      synchronized (lock) {
```
csviri, line 176: Here we still use synchronized keyword as far I can see.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
```diff
@@ -0,0 +1,169 @@
...
+import io.javaoperatorsdk.operator.OperatorException;
+import io.javaoperatorsdk.operator.api.Context;
+import io.javaoperatorsdk.operator.api.DeleteControl;
+import io.javaoperatorsdk.operator.api.ResourceController;
+import io.javaoperatorsdk.operator.api.UpdateControl;
+import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
+import io.javaoperatorsdk.operator.processing.event.DefaultEventSourceManager;
+import io.javaoperatorsdk.operator.processing.event.EventSourceManager;
+
+public class ConfiguredController<R extends CustomResource<?, ?>> implements ResourceController<R>,
```
csviri, line 24: I don't think this should implement ResourceController, couldn't we just get the controller when it's needed instead of delegating and implementing this interface. I would rather see this class which just couples these objects. Implementing this interface does not seems right.

In other words this principle loosly applies (in regard of interfaces):
https://en.wikipedia.org/wiki/Composition_over_inheritance

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -1,52 +1,55 @@
...
-      KubernetesClient k8sClient, ConfigurationService configurationService, Metrics metrics) {
+  public Operator(KubernetesClient k8sClient, ConfigurationService configurationService) {
     this.k8sClient = k8sClient;
     this.configurationService = configurationService;
-    this.closeables = new ArrayList<>();
-    this.lock = new Object();
-    this.controllers = new ArrayList<>();
-    this.started = false;
-    this.metrics = metrics;
+    DefaultEventHandler.setEventMonitor(new EventMonitor() {
```
csviri, line 35: I think this would be better to have these metrics per controller. (Something to think about in the future)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -43,12 +45,14 @@
...
   }
 
-  @SuppressWarnings({"rawtypes", "unchecked"})
-  public <R extends CustomResource<?, ?>> DefaultEventSourceManager(
-      ResourceController controller, ControllerConfiguration configuration, MixedOperation client) {
-    this(new DefaultEventHandler(controller, configuration, client), true);
-    registerEventSource(
-        CUSTOM_RESOURCE_EVENT_SOURCE_NAME, new CustomResourceEventSource<>(client, configuration));
+  public DefaultEventSourceManager(ConfiguredController<R> controller) {
+    this(new DefaultEventHandler<>(controller), true);
```
csviri, line 49: This is just a remark for the future, maybe to discuss separately. It's little bit messy how we wire up and/or instantiate these object DefaultEventHandler here and the DefaultEventSourceManager in the Configured controller. Those should be probable more explicit somewhere, making the code more readable, since these two classes are quite fundamental. Now it's quite hard to trace back, what is instantiated where.

csviri:
Made some comments, but in general I like the idea of this ConfiguredController, so we break down the large chunk of code from Operator class to separate parts. These classes in it are feels cohesive.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: It's actually something I've wondered about. It's from a different PR. I don't think that it's called from multiple threads since it's just called when the app starts.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
metacosm: We're using composition here and to me, it makes sense to have that class implement `ResourceController`. We're not inheriting anything, just stating that this class behaves for all intents and purposes like a `ResourceController`. What are the disadvantages of having the class extend the interface? Implementing the interface allows to pass the object around anywhere a `ResourceController` is used currently which simplifies the current code, imo. The whole idea of this class is that it **is** a `ResourceController` that also manages its configuration (instead of asking a `ConfigurationService` for it).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
metacosm: Yes, this still needs to be cleaned up but this PR is already doing too much as it stands so that work should be done in subsequent PRs.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: Might make sense, indeed, though we can side-step the issue by adding the controller's name to the counter tags so that we can only have one handler and still be able to filter per controller. Having a different monitor per controller would only make sense if we wanted to track different metrics per controller (which we indeed might).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: Yep, seems like I forgot about that part. Let's remove synchronization for now and re-add it when/if needed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
csviri: This is probably more a philosphical question, so also a point of view how we are looking at this class. So if we take a look on it as a "proxy" class then yes you are right there is no harm to implement it. If we take a look on this as a wrapper that just collects some related classes then it is a smell. 
So it's all about how we express notions. But here actually you are probably right that we can look on this as a "proxy" class.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -62,63 +64,61 @@ public ConfigurationService getConfigurationService() {
    */
   @SuppressWarnings("unchecked")
   public void start() {
-    synchronized (lock) {
```
lburgazzoli, line 65: don't we need to keep this lock ? if sop is invoked on a separate thread it may not perform any cleanup but controller may be running

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractControllerConfiguration.java
```diff
@@ -1,10 +1,11 @@
 package io.javaoperatorsdk.operator.api.config;
 
-import io.fabric8.kubernetes.client.CustomResource;
 import java.util.Collections;
 import java.util.Set;
 
-public abstract class AbstractControllerConfiguration<R extends CustomResource>
+import io.fabric8.kubernetes.client.CustomResource;
+
+public abstract class AbstractControllerConfiguration<R extends CustomResource<?, ?>>
```
lburgazzoli, line 8: wonder if this need to be an abstract class or just have some more params to set the crd class name and the configuration service

metacosm: Maybe… the reason why it's abstract is that it was needed like that at some point for the Quarkus extension. Maybe it doesn't need to be abstract anymore. Let me check.

metacosm: It appears it's not needed anymore. I have local changes to this PR to account for changes made for #499 that I intend to merge first. I will make the changes to this PR after the other one is merged.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: Do you have examples / use cases where `close` might be invoked asynchronously?

metacosm: 🤔 Actually, thinking about it some more, if a shutdown hook is installed, then, indeed, `close` will be called from a different thread so we do need to re-instate synchronization.

metacosm:
On a more general note, the integration tests seem to be particularly flaky with this PR so there may something more to it than just flakiness. 😢

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
attilameszaros-form3: But that would be still just 1 thread. Is there still a situation where can be a race condition? By definition this should not be called by multiple threads.

lburgazzoli: it depends who invokes the shutdown process, i.e. in spring & quarkus this is done by the underlying runtime ad we don't know if this happen in the same thread or not.

metacosm: If the operator app is stopped during the start process then it's possible that there would two different threads accessing the `started` and `controllers` fields at the same time.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractControllerConfiguration.java
metacosm: 🤔 Cleaning this up properly, though, would require a breaking API change because, obviously, the class shouldn't be still named `AbstractControllerConfiguration` if it's not abstract anymore.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
csviri: That is true.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractControllerConfiguration.java
metacosm: I fixed the backwards compatibility issue but name has to stay as-is for now.

metacosm:
@csviri can we merge this?

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -219,7 +219,7 @@ public R updateStatus(R resource) {
       return resourceOperation
           .inNamespace(resource.getMetadata().getNamespace())
           .withName(getName(resource))
-          .updateStatus(resource);
+          .patchStatus(resource);
```
attilameszaros-form3, line 222: @metacosm I think we should discuss this more in depth under an issue. I think we need both, patch and update status without optimistic locking. Will create an issue and describe the use cases.

attilameszaros-form3: Can we make this a separate PR from the metric and refactoring?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/EventDispatcher.java
```diff
@@ -227,7 +219,7 @@ public R updateStatus(R resource) {
...
```
csviri, line 222: @metacosm I think we should discuss this more in depth under an issue. I think we need both, patch and update status without optimistic locking. Will create an issue and describe the use cases.

Can we make this a separate PR from the metric and refactoring?

metacosm: Done.

PR #518, opened 2021-08-30
chore: cleanup operator-framework-core pom

lburgazzoli:


operator-framework-core/pom.xml
```diff
@@ -61,12 +53,16 @@
   <dependencies>
     <dependency>
       <groupId>io.fabric8</groupId>
-      <artifactId>openshift-client</artifactId>
+      <artifactId>kubernetes-client</artifactId>
```
metacosm, line 56: We should keep the OpenShift client here so that apps can get the more featured version if needed.

metacosm (changes requested):

operator-framework-core/pom.xml
lburgazzoli: fixed

metacosm (approved):

operator-framework-core/pom.xml
csviri: I don't really get this argument.  Is there more features in openshift client? 
This could be confusing for the users, at least shuold be documented. Also could be replaces in operators with maven if somebody want to use openshift client.

metacosm: There's more feature in the OpenShift client because it provides access to OpenShift-only resources in addition to the plain Kubernetes one. I don't think it causes any confusion as people can interact with this client the exact same way they would with the plain version. Replacing the client in not necessarily that trivial depending on how you do it and this is the easiest/simplest way to have broader support at this time.
Personally, I'd rather see the OpenShift client implemented as an add-on on top of the plain Kubernetes one but I'm not sure that's feasible at this point.

csviri: Can we put this there in form of a comment above this dependency for the users which are not necessary familiar with the relation of these clients?

lburgazzoli: I can include a comment in this pr, what would you like me to add ?

csviri: Basically what @metacosm  mentioned. That "We use openshift client, because functionally it a superset of kubernetes client.". Thank you!

csviri (approved):

operator-framework-core/pom.xml
lburgazzoli: done

operator-framework-core/pom.xml
```diff
@@ -59,14 +51,20 @@
   </build>
 
   <dependencies>
+    <!-- We use openshift client, because functionally it a superset of kubernetes client -->
```
metacosm, line 54:
```suggestion
    <!-- We use the OpenShift client, because functionally it is a superset of the Kubernetes client -->
```

metacosm (changes requested):

PR #527, opened 2021-09-03
chore(ci): drop actions/cache@v2

sullis:
We no longer need to use actions/cache.

Maven dependencies are cached via actions/setup-java

https://github.blog/changelog/2021-08-30-github-actions-setup-java-now-supports-dependency-caching/


metacosm:
I did wonder about that when I activated the caching in the other PR but didn't check. Thanks for undertaking this.

PR #545, opened 2021-09-16
feat: introduce JUnit5 extension

lburgazzoli:


lburgazzoli:
@metacosm @csviri what do you think about this extension ?
I think it could be a basis for a testing framework for operator written using the java-operator-sdk

metacosm:
I'm planning on looking at this today.

csviri:
I will take a look tomorrow morning

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
```diff
@@ -60,12 +54,14 @@ protected void throwExceptionOnNameCollision(
     final var key = keyFor(controller);
     final var configuration = configurations.get(key);
     if (configuration == null) {
-      log.warn(
-          "Configuration for controller '{}' was not found. {}", key, getControllersNameMessage());
+      logWarning(key, getControllersNameMessage());
```
lburgazzoli, line 57: wonder if we need to log things here or should be a responsibility of the caller

metacosm: I added that logging feature a while back because missing controller configuration could lead to rather unclear subsequent error messages. It's definitely not needed but it helps the user figure out what might be wrong.

metacosm:
Do you intend to make more changes, @lburgazzoli?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
```diff
@@ -0,0 +1,28 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BaseConfigurationService extends AbstractConfigurationService {
```
csviri, line 6: I don't get this class. Why are we creating loggers this way?

csviri:
> @metacosm @csviri what do you think about this extension ?
> I think it could be a basis for a testing framework for operator written using the java-operator-sdk

Starting to look. Would it be an idea to add description to the PR's such this, where we explain some problem statement, so what is we are solving? What do you think?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
```diff
@@ -60,12 +54,14 @@ protected void throwExceptionOnNameCollision(
     final var key = keyFor(controller);
     final var configuration = configurations.get(key);
     if (configuration == null) {
-      log.warn(
-          "Configuration for controller '{}' was not found. {}", key, getControllersNameMessage());
+      logMissingControllerWarning(key, getControllersNameMessage());
```
csviri, line 57: Just to my understanding, why is this better?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
metacosm: This class is just used to be able to have a non-abstract implementation that can be used in the junit extension. I'm debating whether to just put the `DefaultConfigurationService` into core itself and get rid of this class completely but I'm not too sure of the impact on the quarkus extension…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
metacosm: The benefit is minimal, indeed, just to be able to not pull SLF4J when using `AbstractConfigurationService`.

csviri: Hmm, isn't this a little bit over-engineering? We use SLF4J everywhere, and it's quite an industry standard

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
csviri: oki, 
(will take a look on the quarkus side to understand these aspects better)

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -0,0 +1,272 @@
...
+  private final String namespace;
+  private final Operator operator;
+  private final boolean preserveNamespaceOnError;
+  private final boolean waitForNamespaceDeletion;
+
+  private OperatorExtension(
+      ConfigurationService configurationService, boolean preserveNamespaceOnError,
+      boolean waitForNamespaceDeletion) {
+    this.kubernetesClient = new DefaultKubernetesClient();
+    this.namespace = UUID.randomUUID().toString();
```
csviri, line 59: Cannot we add more semantic name? Like name of the test method?

lburgazzoli: yes, that's possible, let me try

lburgazzoli: done

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -0,0 +1,274 @@
...
+    this.waitForNamespaceDeletion = waitForNamespaceDeletion;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  @Override
+  public void beforeAll(ExtensionContext context) throws Exception {
+    before(context);
```
csviri, line 74: Why do we execute this before all? Should be this either before each or before all but not both?

lburgazzoli: the invocation depend on how you instantiate the extension:
- if static --> beforeAll 
- if non static --> beforeEach

(at least that is my understanding)

csviri: ahh, ok, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
metacosm: The other issue is also that the previous version hardcoded the name of the logger, which might not have been appropriate depending on the actual implementation being executed (for example, even the Quarkus implementation would have logged something with `Default ConfigurationService implementation` logger name).

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -135,6 +135,7 @@ protected void before(ExtensionContext context) {
     namespace += "-";
     namespace += context.getRequiredTestMethod().getName();
     namespace = KubernetesResourceUtil.sanitizeName(namespace).toLowerCase(Locale.US);
+    namespace = namespace.substring(0, Math.min(namespace.length(), 63));
```
metacosm, line 138: So the length is not taken into account in `sanitizeName`? That sounds like a fabric8 bug to me…

lburgazzoli: Correct, what it does is:

```java
name.replaceAll("[^A-Za-z0-9]+", "-")
```

csviri:
In general I like this idea, looks good, makes the tests more readable. Hopefully won't make it more complex to understand the framework/tests for newcomers.

csviri (approved):

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
metacosm: Yep, just saw that, will fix it now :)

metacosm (approved):

PR #546, opened 2021-09-17
feat: use only one, configurable ExecutorService

### LINKED ISSUE #540: Using only one instance of ScheduledExecutorService
Currently we are using `ScheduledThreadPoolExecutor` per controller. It would be better to control threading and related configurations if it would be shared between controllers. So there are no more such services created.

metacosm:
Fixes #540
cc @lburgazzoli


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -146,7 +124,13 @@ private void executeBufferedEvents(String customResourceUid) {
               latestCustomResource.get(),
               retryInfo(customResourceUid));
       log.debug("Executing events for custom resource. Scope: {}", executionScope);
-      executor.execute(new ExecutionConsumer(executionScope, eventDispatcher, this));
+      executor.execute(() -> {
```
csviri, line 127: Nice!

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
metacosm: Actually, I'm probably going to revert that change so that we can override `toString` to get more precise debugging information on the Runnable… 🥲

csviri: What I meant eliminating the ExecutionConsumer is elegant here

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -146,7 +124,13 @@ private void executeBufferedEvents(String customResourceUid) {
...
+        // change thread name for easier debugging
+        Thread.currentThread().setName("EventHandler-" + controllerName);
```
csviri, line 129: Is this needed? It looks a bit strange that we change the thread name on every execution.

lburgazzoli:
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
metacosm: Yes, I understood that but like I said, I actually need to re-introduce some equivalent form of it :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
metacosm: Well, that was previously done when initializing the thread pool by specifying the thread factory. We can't do that anymore so the only alternative I could think of at that point was to rename the thread when it executes.

PR #553, opened 2021-09-22
fix: misleading MissingCRDException exception

### LINKED ISSUE #552: Misleading MissingCRDException exception
If an event source fails to start and the error code is 404, the `DefaultEventSourceManager` wraps it with a `MissingCRDException` [1] which the `ConfiguredController` re-wrap and include details about the CRD of the controller resulting in a misleading message as the exception will report that the CRD of the controller is missing but it may not be the case.

[1] https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java#L92-L97
[2] https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L177-L182

lburgazzoli:
Fixes #552

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -15,8 +15,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.client.CustomResource;
-import io.fabric8.kubernetes.client.KubernetesClientException;
-import io.javaoperatorsdk.operator.MissingCRDException;
 import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.processing.ConfiguredController;
 import io.javaoperatorsdk.operator.processing.CustomResourceCache;
@@ -88,12 +86,6 @@ public class DefaultEventSourceManager<R extends CustomResource<?, ?>>
       if (e instanceof IllegalStateException) {
         // leave untouched
         throw e;
-      }
-      if (e instanceof KubernetesClientException) {
-        KubernetesClientException ke = (KubernetesClientException) e;
-        if (404 == ke.getCode()) {
-          throw new MissingCRDException(null, null);
-        }
       }
       throw new OperatorException("Couldn't register event source named '" + name + "'", e);
     } finally {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
@@ -174,12 +174,8 @@ public class ConfiguredController<R extends CustomResource<?, ?>> implements Res
       CustomResourceUtils.assertCustomResource(resClass, crd);
     }
 
-    try {
-      DefaultEventSourceManager<R> eventSourceManager = new DefaultEventSourceManager<>(this);
-      controller.init(eventSourceManager);
-    } catch (MissingCRDException e) {
-      throwMissingCRDException(crdName, specVersion, controllerName);
-    }
+    DefaultEventSourceManager<R> eventSourceManager = new DefaultEventSourceManager<>(this);
+    controller.init(eventSourceManager);
 
     if (failOnMissingCurrentNS()) {
       throw new OperatorException(


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -89,12 +87,6 @@ public final void registerEventSource(String name, EventSource eventSource)
         // leave untouched
         throw e;
       }
-      if (e instanceof KubernetesClientException) {
-        KubernetesClientException ke = (KubernetesClientException) e;
-        if (404 == ke.getCode()) {
-          throw new MissingCRDException(null, null);
-        }
-      }
```
lburgazzoli, line 97: @csviri @metacosm removed this because a 404 may also be the result of watching a resource by name or in a namespace that does not exists so assuming 404 is only when a CRD is missing seems wrong

metacosm:
We need to find a better way to check whether the 404 occurs because of a missing CRD or for some other reason. The `MissingCRDException` mechanism is central to how the Quarkus automatically applies missing CRD so we need that exception to be thrown there in this situation.

lburgazzoli:
The check for the operator CRD is performed before (https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L165-L175) so I guess this is what Quarkus leverages. 

The issue I'm facing is when an event source is registered and i.e. it may be that a CRD is missing but could be an external CRD so Quarkus won't be able to apply them (I guess)

metacosm:
> The check for the presence of the controller CRD is performed before (https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L165-L175) so I guess this is what Quarkus leverages.

Unfortunately, this isn't enough, which is why I added that code in the first place. However, I agree that it's subpar… I'm working on an hopefully better solution.

lburgazzoli:
> > The check for the presence of the controller CRD is performed before ([https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L165-L175](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java?rgh-link-date=2021-09-22T15%3A49%3A26Z#L165-L175)) so I guess this is what Quarkus leverages.
> 
> Unfortunately, this isn't enough, which is why I added that code in the first place. However, I agree that it's subpar… I'm working on an hopefully better solution.

Wonder if it would be simpler if Quarkus should just apply the CR if it is configured to do so without checking, what would be the downside ?

metacosm:
> > > The check for the presence of the controller CRD is performed before ([https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L165-L175](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java?rgh-link-date=2021-09-22T15%3A49%3A26Z#L165-L175)) so I guess this is what Quarkus leverages.
> > 
> > 
> > Unfortunately, this isn't enough, which is why I added that code in the first place. However, I agree that it's subpar… I'm working on an hopefully better solution.
> 
> Wonder if it would be simpler if Quarkus should just apply the CR if it is configured to do so without checking, what would be the downside ?

That's a good point and I actually think that's the direction things are going now… The exception was introduced when the extension didn't have the ability to re-apply the CRD on changes but it now has that ability so we might indeed simplify things here… I will have to check. I've been working on improving the exceptions in the fabric8 client as well so will first finish on that path to validate the approach. 
That said, the `MissingCRDException` should be interesting for more than just the Quarkus extension as a diagnostic tool for users as well (because that's quite a common occurrence for beginners with the SDK to forget to apply the CRD to the cluster and, previously, you used to get a cryptic 404 error without any details).

lburgazzoli:
> > > > The check for the presence of the controller CRD is performed before ([https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L165-L175](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java?rgh-link-date=2021-09-22T18%3A31%3A00Z#L165-L175)) so I guess this is what Quarkus leverages.
> > > 
> > > 
> > > Unfortunately, this isn't enough, which is why I added that code in the first place. However, I agree that it's subpar… I'm working on an hopefully better solution.
> > 
> > 
> > Wonder if it would be simpler if Quarkus should just apply the CR if it is configured to do so without checking, what would be the downside ?
> 
> I've been working on improving the exceptions in the fabric8 client as well so will first finish on that path to validate the approach.

+1

> That said, the `MissingCRDException` should be interesting for more than just the Quarkus extension as a diagnostic tool for users as well (because that's quite a common occurrence for beginners with the SDK to forget to apply the CRD to the cluster and, previously, you used to get a cryptic 404 error without any details).

Agree, better exception from fabric8 client would be very useful.

lburgazzoli:
I guess I can close this one then ?

metacosm:
> I guess I can close this one then ?

I think so, yes. Just let me know if the fix in #558 works for you. :)

lburgazzoli:
LGTM

metacosm:
Superseded by #558. Thanks for bringing this up!

PR #558, opened 2021-09-23
fix: only throw MissingCRDException if we get a 404 on the target CRD

### LINKED ISSUE #552: Misleading MissingCRDException exception

metacosm:
Note that this relies on fabric8io/kubernetes-client#3492 so won't build
until this PR is merged and available in a release.
Fixes #552

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
@@ -35,18 +35,21 @@ public class DefaultEventSourceManager<R extends CustomResource<?, ?>>
   private final Map<String, EventSource> eventSources = new ConcurrentHashMap<>();
   private final DefaultEventHandler<R> defaultEventHandler;
   private TimerEventSource<R> retryTimerEventSource;
+  private final String targetCRDName;
 
-  DefaultEventSourceManager(DefaultEventHandler<R> defaultEventHandler, boolean supportRetry) {
+  DefaultEventSourceManager(DefaultEventHandler<R> defaultEventHandler, boolean supportRetry,
+      String targetCRDName) {
     this.defaultEventHandler = defaultEventHandler;
     defaultEventHandler.setEventSourceManager(this);
     if (supportRetry) {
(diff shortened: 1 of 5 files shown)


lburgazzoli:
LGTM

wonder if the MissingCRDException should only be thrown by the CustomResourceEventHanderl instead of any event source

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -92,7 +95,10 @@ public final void registerEventSource(String name, EventSource eventSource)
       if (e instanceof KubernetesClientException) {
```
csviri, line 95: Should not we catch this exception and make these checks in the CustomResourceEventSource? 
(And just throw the exception further if this is not the exception we want to trasnlate)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/DefaultEventSourceManager.java
```diff
@@ -92,7 +95,10 @@ public final void registerEventSource(String name, EventSource eventSource)
...
         KubernetesClientException ke = (KubernetesClientException) e;
         if (404 == ke.getCode()) {
```
csviri, line 97: I mean this logic. This is quite specific to CustomResourceEventSource. Cannot be moved there?

csviri (approved):
LGTM!

Nice late night PR :)

PR #559, opened 2021-09-23
chore(ci): use Java 17

metacosm:


metacosm:
2 options:
- Either only run on Temurin JDKs
- Wait for OpenJ9 to release a Java 17 JDK

metacosm:
Let's go ahead and focus solely on Temurin for now… Would be nice if we could have more control over the distribution x version matrix so that we could say use version 11 & 17 for Temurin and only 11 for OpenJ9 but 🤷🏼

PR #561, opened 2021-09-24
Remove spring boot version - not used anymore

csviri:


csviri:
Nevermind still in the example

PR #562, opened 2021-09-24
fix: make it clear that `ConfiguredController.init` should not be called

### LINKED ISSUE #554: ConfiguredController behavior
I see that the ConfiguredController has an [init](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L106-L110) method in which does:

```java
@Override
public void init(EventSourceManager eventSourceManager) {
  this.manager = eventSourceManager;
  controller.init(eventSourceManager);
}
```

So the underlying controller is initialized with the manager that is passed as an argument. But then the [start](https://github.com/java-operator-sdk/java-operator-sdk/blob/master/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java#L159) method does something similar:

```java
try {
  DefaultEventSourceManager<R> eventSourceManager = new DefaultEventSourceManager<>(this);
  controller.init(eventSourceManager);
} catch (MissingCRDException e) {
  throwMissingCRDException(crdName, specVersion, controllerName);
}
```

Wonder if this is correct or if this lead the controller to be initialized with two different event source manager.

metacosm:
Fixes #554

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ConfiguredController.java
@@ -105,8 +105,7 @@ public class ConfiguredController<R extends CustomResource<?, ?>> implements Res
 
   @Override
   public void init(EventSourceManager eventSourceManager) {
-    this.manager = eventSourceManager;
-    controller.init(eventSourceManager);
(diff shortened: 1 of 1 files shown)


csviri (approved):

lburgazzoli:
LGTM

Issue #564, opened 2021-09-27
Support Informers by Event Sources


csviri:
In order to manage dependent k8s resources easily, ideally we would utilize Informers. Since there is already a implementation for informers in fabric8 client, we can provide an EventSource implementation for them. Effectively this will be Informer adaptor to the event source architecture.

See also:
https://github.com/fabric8io/kubernetes-client/tree/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/informers

PR #565, opened 2021-09-27
Informer event source

csviri:
see: https://github.com/java-operator-sdk/java-operator-sdk/issues/564

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
@@ -1,0 +1,98 @@
+package io.javaoperatorsdk.operator.processing.event.internal;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
+import io.fabric8.kubernetes.client.informers.SharedInformer;
+import io.fabric8.kubernetes.client.informers.cache.Store;
+import io.javaoperatorsdk.operator.processing.event.AbstractEventSource;
+
+public class InformerEventSource<T extends HasMetadata> extends AbstractEventSource {
+
+  private final SharedInformer<T> sharedInformer;
+  private final ResourceToRelatedCustomResourceUIDMapper<T> mapper;
+  private final boolean skipUpdateEventPropagationIfNoChange;
+
+  public InformerEventSource(SharedInformer<T> sharedInformer,
+      ResourceToRelatedCustomResourceUIDMapper<T> mapper) {
+    this(sharedInformer, mapper, true);
+  }
+
+  InformerEventSource(KubernetesClient client, Class<T> type,
+      ResourceToRelatedCustomResourceUIDMapper<T> mapper) {
+    this(client, type, mapper, false);
+  }
+
+  InformerEventSource(KubernetesClient client, Class<T> type,
+      ResourceToRelatedCustomResourceUIDMapper<T> mapper,
+      boolean skipUpdateEventPropagationIfNoChange) {
+    this(client.informers().sharedIndexInformerFor(type, 0), mapper,
+        skipUpdateEventPropagationIfNoChange);
+  }
+
+  public InformerEventSource(SharedInformer<T> sharedInformer,
+      ResourceToRelatedCustomResourceUIDMapper<T> mapper,
+      boolean skipUpdateEventPropagationIfNoChange) {
+    this.sharedInformer = sharedInformer;
+    this.mapper = mapper;
+    this.skipUpdateEventPropagationIfNoChange = skipUpdateEventPropagationIfNoChange;
+
+    sharedInformer.addEventHandler(new ResourceEventHandler<T>() {
+      @Override
+      public void onAdd(T t) {
+        propagateEvent(InformerEvent.Action.ADD, t, null);
+      }
+
+      @Override
+      public void onUpdate(T oldObject, T newObject) {
+        if (InformerEventSource.this.skipUpdateEventPropagationIfNoChange &&
+            oldObject.getMetadata().getResourceVersion()
+                .equals(newObject.getMetadata().getResourceVersion())) {
+          return;
+        }
+        propagateEvent(InformerEvent.Action.UPDATE, newObject, oldObject);
+      }
+
+      @Override
+      public void onDelete(T t, boolean b) {
+        propagateEvent(InformerEvent.Action.DELETE, t, null);
+      }
+    });
+  }
+
+  private void propagateEvent(InformerEvent.Action action, T object, T oldObject) {
+    var uid = mapper.map(object);
+    if (uid.isEmpty()) {
+      return;
+    }
+    InformerEvent event = new InformerEvent(uid.get(), this, action, object, oldObject);
+    this.eventHandler.handleEvent(event);
+  }
+
+  @Override
+  public void start() {
+    sharedInformer.run();
+  }
+
+  @Override
+  public void close() throws IOException {
+    sharedInformer.close();
+  }
+
+  public Store<T> getStore() {
+    return sharedInformer.getStore();
+  }
+
+  public SharedInformer<T> getSharedInformer() {
+    return sharedInformer;
+  }
+
+  public interface ResourceToRelatedCustomResourceUIDMapper<T> {
+    // in case cannot map to the related CR uid, skip the event processing
+    Optional<String> map(T resource);
+  }
+
+}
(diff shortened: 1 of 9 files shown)


csviri:
The PR is ready for review. Was thinking to add an additional filter option for event coming from informer. However filtering can be now done with the mapper (what maps the event to UID now), and can be added also later. Skipped this for now.

Plan to work on the docs, related to this. Will have those as separate PR.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
```diff
@@ -0,0 +1,85 @@
...
+import io.javaoperatorsdk.operator.processing.event.AbstractEventSource;
+
+public class InformerEventSource<T extends HasMetadata> extends AbstractEventSource {
+
+  private final SharedInformer<T> sharedInformer;
+  private final ResourceToRelatedCustomResourceUIDMapper<T> mapper;
+  private final boolean skipUpdateEventPropagationIfNoChange;
+
+  public InformerEventSource(SharedInformer<T> sharedInformer,
+      ResourceToRelatedCustomResourceUIDMapper<T> mapper) {
```
lburgazzoli, line 19: Wonder if we should use `Predicate<CustomResource>` instead of adding another mapping

csviri: Not sure how that would work, we really need to produce an UID here. So we can associate the event with the related custom resource.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
+import io.fabric8.kubernetes.client.informers.SharedInformer;
+import io.fabric8.kubernetes.client.informers.cache.Store;
+import io.javaoperatorsdk.operator.processing.event.AbstractEventSource;
+
+public class InformerEventSource<T extends HasMetadata> extends AbstractEventSource {
```
lburgazzoli, line 12: Wonder if we should have a simple constructor that hides the details about creating an informer

csviri: Not sure if that is desirable, fabric8 now supports shared informers, so it might be also common case that those are shared between multiple EventSource. 
Also I don't see how that would look like, based on the current fabric8 api for informers.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
lburgazzoli: My understanding is that the `InformerEvent` that is generated by the event source relies on the `Predicate<CustomResource>` behind the scenes as when it is constructed with an UID it invokes `new UIDMatchingPredicate(relatedCustomResourceUid)` so I guess it could be made generic and just pass the `Predicate<CustomResource>` from the `InformerEventSource` constructor.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
lburgazzoli: Correction: Wonder if we should **ALSO** have a simple constructor that hides the details about creating an informer, something like:

```java
InformerEventSource(KubernetesClient client, Class<T> type) {
    this.informer = client.informers().sharedIndexInformerFor(type.class, 0);
}
```

lburgazzoli:
LGTM, I had 2 minor notes that could be solved/postponed/discussed also after this PR gets merged

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
csviri: Added additional constructors.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
csviri: I'm checking this now, and discovered one thing, how this predicate is used:
```
  public Set<String> getLatestResourcesUids(Predicate<CustomResource> selector) {
    try {
      lock.lock();
      return resources.values().stream()
          .filter(selector)
          .map(r -> r.getMetadata().getUid())
          .collect(Collectors.toSet());
    } finally {
      lock.unlock();
    }
  }
```
But this is quite un-efficient, before it was just getting the resource from a hashmap using an the UID. Now this is traversing a possible hube map.

csviri: We will have to change this anyways in V2. I mean how we are addressing the resources (will have to be namespace+name). So, hmm, not sure, for now I would rather stick with the current implementation, so that change is not even more confusing in the future.

lburgazzoli: Yes we need to improve this area i.e. 

I think if we use and indexed informer, then we can improve the performances here. 

The issue withe the UID is that is not always possible to use it, in my case as example a change to a monitored resource could lead to a reconcile loop to be triggered on a number of custom resources thus the need to have something like that.

I guess in the future this selector could be defined with something like: 

```java
List<CustomResource> apply(ReadOnlyStore store);
```

So then the selector can implement an optimized version if the resource can be addressed by a key.

csviri: The problem with this that, the custom resource itself is not accessible at this moment. So this as in the example applies for a config map, we just know the owner reference or something what we explicitly add to the resource (UID for example).

Also why would it be a list?

In v2 the referencing of the CustomResource will  be always namespace+name (known already known). So reading this from owner reference would be possible.

lburgazzoli: The issues is that you may want to watch resources that are not owned by you (i.e. created by another operator) so you can't have a 1-1 relation and you can't rely on the owner ref but instead you add information to your custom resources about i.e. the name of the external resources.

In such case, when you receive an event related to external resource, you need to scan the operator cache and find any custom resource for which the metadata matches and since there no guarantee that the relation is always 1-1, then you need to return a list.

csviri: That is a good point. But in that case we have to make the cache available for the custom resource event source. What makes these changes much more complex. Especially, since shortly we will make the `CustomResourceEventSource` based on informer. Where then we will have completely different API.

csviri: > The issues is that you may want to watch resources that are not owned by you (i.e. created by another operator) so you can't have a 1-1 relation and you can't rely on the owner ref but instead you add information to your custom resources about i.e. the name of the external resources.

On the other hand don't think this applies within one controller. So the that a resource is a dependent resource of multiple custom resources is probably not the case. Although it can happen that a resource change can have impact to multiple custom resource.

lburgazzoli: Lots of good point in this discussion, sdk v2.0 will be awesome :)

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/informereventsource/InformerEventSourceTestCustomResourceController.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  public UpdateControl<InformerEventSourceTestCustomResource> createOrUpdateResource(
+      InformerEventSourceTestCustomResource resource,
+      Context<InformerEventSourceTestCustomResource> context) {
+
+    // Reading the config map from the informer not from the API
+    // name of the config map same as custom resource for sake of simplicity
+    ConfigMap configMap =
```
metacosm, line 61: I think we should make that stuff simpler but we can do that in a separate PR.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/informereventsource/InformerEventSourceTestCustomResourceController.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private KubernetesClient kubernetesClient;
+  private SharedInformer<ConfigMap> informer;
+
+  @Override
+  public void init(EventSourceManager eventSourceManager) {
+    SharedInformerFactory sharedInformerFactory = kubernetesClient.informers();
+    informer = sharedInformerFactory.sharedIndexInformerFor(ConfigMap.class, 0);
+    eventSourceManager.registerEventSource("configmap", new InformerEventSource<>(informer,
+        resource -> {
```
metacosm, line 46: We should provide default mappers to make this use case easier to implement.

metacosm (approved):
LGTM, comments should be discussed / addressed in subsequent PRs

PR #572, opened 2021-09-29
Improvements on Test Flakyness

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/retry/GenericRetryExecution.java
```diff
@@ -6,8 +6,8 @@
 
   private final GenericRetry genericRetry;
 
-  private int lastAttemptIndex = 0;
-  private long currentInterval;
+  private volatile int lastAttemptIndex = 0;
```
csviri, line 9: This might cause one of the tests fails. It does not do any harm otherwise since we might access those from different threads

lburgazzoli (approved):

PR #577, opened 2021-10-04
feat: extract micrometer support into its own module

### LINKED ISSUE #576: Provide a no-dependency `Metrics` class implementation
We should separate the `Metrics` class from micrometer so that users who are not interested in monitoring don't carry the dependency. This would mean creating a separate module for the micrometer-based implementation.

metacosm:
Fixes #576. This allows getting rid of mandatory micrometer dependency
if users are not interested in having metrics.

### DIFF
--- a/micrometer-support/pom.xml
+++ b/micrometer-support/pom.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>java-operator-sdk</artifactId>
+    <groupId>io.javaoperatorsdk</groupId>
+    <version>1.9.8-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>micrometer-support</artifactId>
+  <name>Operator SDK - Micrometer Support</name>
+
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.micrometer</groupId>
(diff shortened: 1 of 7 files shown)


csviri:
I'm not against this, it makes sense. Just wonder if has that much added benefit, while we introduce an additional module. But again I'm not against this.

lburgazzoli:
I think the split makes sense however I think we should reason a little bit more about how metrics related bits are added into the code as example for what concerns `DefaultEventHandler` we don't to a great job:

https://github.com/java-operator-sdk/java-operator-sdk/blob/4d36922c90c49767a355c44006fc5586fe91a967/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java#L35-L44
https://github.com/java-operator-sdk/java-operator-sdk/blob/4d36922c90c49767a355c44006fc5586fe91a967/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java#L35-L46

In the code above we are saying:
1. `DefaultEventHandler` is one of the possible implementation of `EventHandler`
2. `DefaultEventHandler` has a static field that is used to configure metrics
3. well, having the `EventHandler` interface in not very useful as we rely on a static field on `DefaultEventHandler` to set-up metrics.

So I wonder if instead, if metrics are enabled, then we should instantiate a `InstrumentedEventHandler` maybe provided by the micrometer support which IMHO, would make the addition of this new module even more meaningful (the `InstrumentedEventHandler` can be eventually loaded using the `ServiceLoader` mechanism, Quarkus could optimize even optimize it by pre-loading and configuring things at build time)

lburgazzoli (approved):

csviri:
Definitely, will take a look on that also soon. Maybe we can discuss that under a new issue, and we can discuss the design there.

What I don't like on micrometer core that is has bunch of dependencies. That is a good argument to not have it on the classpath by default.
On the other hand it's de facto industry standard, and every production system needs to have metrics monitored. So eventually everybody will use this module. And more importantly they should.

So if you insist to merge this I'm fine. Just let's think in the future if it really has the benefit, since we have to then create additional abstractions for this in core.

csviri (approved):

metacosm:
@csviri @lburgazzoli I changed how we deal with `EventMonitor`, hopefully it makes more sense now… I can't really do the full clean-up I'd like to until v2, though… Let me know what you think!

csviri:
LGTM

Issue #538, opened 2021-09-16
Removing events from Context


csviri:
Currently the event(s) triggered the controller execution is accessible from the Conext object. There is already a consensus that using these events is an antipattern in operators. Controller Runtime does not expose these event either.

Removing them will have two benefits:

- The intended usage is more obvious, so we won't even allow users to use events, so the mentioned anti-patter is not even possible to implement
- We can simplify the DefaultEventHandler so we don't store events, just a flag is the controller should be executed or not. This will reduce memory consumption.

lburgazzoli:
Would fix also: https://github.com/java-operator-sdk/java-operator-sdk/issues/469

PR #581, opened 2021-10-05
Informer based CustomResourceEventSource and caching

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/DefaultEventHandler.java
```diff
@@ -119,24 +125,17 @@ private EventMonitor monitor() {
...
-
-      final Predicate<CustomResource> selector = event.getCustomResourcesSelector();
       final var monitor = monitor();
-      for (String uid : eventSourceManager.getLatestResourceUids(selector)) {
-        eventBuffer.addEvent(uid, event);
-        monitor.processedEvent(uid, event);
-        executeBufferedEvents(uid);
-      }
+      eventBuffer.addEvent(event.getRelatedCustomResourceID(), event);
+      monitor.processedEvent(event.getRelatedCustomResourceID(), event);
```
lburgazzoli, line 137: I know that it was like that even before but wonder if that's correct as in fact here the event is added to a buffer for late processing. In case of failure, the event may also be put back to the buffer so I guess, `processedEvent` should be invoked only when the event has been delivered to the controller. Eventually we can also add metrics about how long it takes for an event from the generation to the delivery. 

@metacosm am I missing something ?

csviri: That is probably a good point. The event is not processed at this time. We should probably take a look on this in a separate PR.
Note that we plan to remove the event from the context:
https://github.com/java-operator-sdk/java-operator-sdk/issues/538

so not sure how this part will look like after this issue. We might still receive events, but the whole event handling should be simplified.

csviri:
@lburgazzoli @metacosm  pls review this

metacosm:
LGTM apart from the last annotation issue. Could you please rebase and squash before merging as well, making sure all the commit messages follow the conventional format?

PR #613, opened 2021-10-19
feat: extract version logging to make testing easier

metacosm:
Also removed expectation setting in CustomResourceSelectorTest since it
actually wasn't working as we cannot use the mock server in both CRUD
and expectations mode.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -26,14 +26,21 @@ public class Operator implements AutoCloseable {
   private final KubernetesClient kubernetesClient;
   private final ConfigurationService configurationService;
   private final ControllerManager controllers = new ControllerManager();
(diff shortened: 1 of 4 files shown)


csviri:
I'm not sure if I'm getting this PR, so we have a version logger interface, but here there is no implementation of it. Is this related to quarkus extension?

metacosm:
> I'm not sure if I'm getting this PR, so we have a version logger interface, but here there is no implementation of it.

There is an implementation, see: https://github.com/java-operator-sdk/java-operator-sdk/pull/613/files#diff-d5424f5942d7550b1840bcff251f386eda527a265178425e0f468630bffe937dR4. We could also encapsulate the current behavior into an implementation but that doesn't really bring much.

This actually has no impact on the Quarkus extension because the only way to configure which logger implementation to use at this time is to pass it to the Operator's constructor. The goal here is to be able to not log versions during tests. Not convinced that we need this, we could simply tell people not to upgrade to 5.9.

metacosm:
Closing this as obsolete at this point.

PR #642, opened 2021-10-29
feat: contextual information about processing for loging using MDC

### LINKED ISSUE #37: Using MDC (Mapped Diagnostic Context) to improve logging
In order to make it easy to work with logs, thus correlate log messages within an operator (1 process) we could add "correlation ids" or data (note this is not the same as we add correlation IDs in distributed systems) or more like the mapped diagnostic context:
see: https://www.baeldung.com/mdc-in-log4j-2-logback
to the logs. So we can correlate logs for the custom resource, but also for a reconciliation execution path. Note that this involves multiple threads, so its not just using directly MDC.

So we can add to for every log message with MDC: 
- resource uuid 
- resource version
- generated uuid for every reconciliation (so we can see which log messages are related to a particular reconciliation)

csviri:
closes #37

### DIFF
--- a/docs/documentation/features.md
+++ b/docs/documentation/features.md
@@ -33,6 +33,24 @@ permalink: /docs/features
 
 ## Monitoring with Micrometer
 
+## Contextual Info for Logging with MDC
+
+Logging is enhanced with additional contextual information using [MDC](http://www.slf4j.org/manual.html#mdc). 
+When it is following contextual information is added to logging, it is available in most parts of reconciliation logic
+and during the execution of the controller:
+
+| MDC Key      | Value added from Custom Resource |
+| :---        |    :---   | 
+| `resource.apiVersion`   | `.apiVersion` |
+| `resource.kind`   | `.kind` |
+| `resource.name`      | `.metadata.name` | 
+| `resource.namespace`   | `.metadata.namespace` |
+| `resource.resourceVersion`   | `.metadata.resourceVersion` |
+| `resource.generation`   | `.metadata.generation` |
+| `resource.uid`   | `.metadata.uid` |
+
+For more information about MDC see this [link](https://www.baeldung.com/mdc-in-log4j-2-logback).
+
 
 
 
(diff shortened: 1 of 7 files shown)


lburgazzoli:
LGTM, however I wonder if we should make such things something a user can opt in

csviri:
> LGTM, however I wonder if we should make such things something a user can opt in

Hmm good question, so if the user should be able to specify the keys and values interested in. Wondering what would be the scope of that, just the custom resource?

Doing the refactor on the other branch, so would rather merge this and create a separate issue and see if there is an elegant way to do it that way.

lburgazzoli:
> > LGTM, however I wonder if we should make such things something a user can opt in
> 
> Hmm good question, so if the user should be able to specify the keys and values interested in. Wondering what would be the scope of that, just the custom resource?
> 

I meant more a flag to enable/disable MDC entirely (or a dedicated module that enable MDC logging if on the classpath)

csviri:
@lburgazzoli  what would be the benefit of that? 
You configure the logging pattern not to log it. SLF4J contains it by default.

lburgazzoli:
There is usually a cost associated MDC  (thread local, maps and other allocations) so if you don't require MDC, then you can turn it off and save some memory

csviri:
yes, but that part usually doable on the impl level, 
for log4j2 at least (on the bottom) https://logging.apache.org/log4j/2.x/manual/thread-context.html

So not sure if it makes that much sense to solve on this level.  In case the logger implementation would not handle that sure.

lburgazzoli:
I don't have any strong opinion, so it is fine with me but as personal preference, I always like more when I have the option to opt in to something that to opt out :)

csviri:
> I don't have any strong opinion, so it is fine with me but as personal preference, I always like more when I have the option to opt in to something that to opt out :)

A completely agree in general. Want to do a round to review for v2 how we are configuring the operator and controller. Will take in mind also this and put it under a feature flag if dobale nicely.

PR #645, opened 2021-11-02
fix: use `close` as destroyMethod of Bean

KimDoubleB:
`Operator` does not have a `stop` method.
In accordance with the meaning, instead of the `stop` method, the `close` method was modified to operate as `destroyMethod `.


csviri:
Thx @KimDoubleB . good catch!

PR #659, opened 2021-11-05
Add tomcat operator sample

adam-sandor:
Move the Tomcat Operator sample from the samples repo and add end to end tests in the process.


sample-operators/tomcat-operator/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
csviri, line 1: This does not seems to be different from the one in the main resources. Shouldn't we remove the whole file?

adam-sandor: I think it's ok for an application to define it's own logging format, but of course we can leave it out too.

csviri: I mean, if it's necessary to have an other logging format for tests then for the application, seems to be redundant. What I propose is to delete the format settings from the test resources.

csviri:
Maybe we should rename the `samples` module to samples-simple and the current one to samples, what do you think?

its quite uglu to have `sample-operators` and `samples` in the same folder, or move the current simple samples under `simple` module in `samle-operators `

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/TomcatController.java
```diff
@@ -0,0 +1,188 @@
...
+
+  private DeploymentEventSource deploymentEventSource;
+
+  public TomcatController(KubernetesClient client) {
+    this.kubernetesClient = client;
+  }
+
+  @Override
+  public void init(EventSourceManager eventSourceManager) {
+    this.deploymentEventSource = DeploymentEventSource.createAndRegisterWatch(kubernetesClient);
```
csviri, line 42: @adam-sandor could we migrate this to `InformerEventSource` ?

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/TomcatController.java
```diff
@@ -0,0 +1,188 @@
...
+    eventSourceManager.registerEventSource("deployment-event-source", this.deploymentEventSource);
+  }
+
+  @Override
+  public UpdateControl<Tomcat> createOrUpdateResource(Tomcat tomcat, Context<Tomcat> context) {
+    Optional<CustomResourceEvent> latestCREvent =
+        context.getEvents().getLatestOfType(CustomResourceEvent.class);
+    if (latestCREvent.isPresent()) {
```
csviri, line 50: Using event is kinda anti-pattern, would be better to not use it in our samples, could we remove it from this controller? (2 places) ,

adam-sandor:
> Maybe we should rename the `samples` module to `samples-simpl`e and the current one to samples, what do you think?
> 
> its quite ugly to have `sample-operators` and `samples` in the same folder, or move the current simple samples under `simple` module in `samle-operators `
> 
> @adam-sandor @metacosm what do you think?

How about integration-test-samples instead of simple-samples? Just an idea

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/TomcatController.java
adam-sandor: Yeah I'd love to see how that would look like.

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/TomcatController.java
adam-sandor: Agreed, I thought about changing this but then decided to have the e2e tests first and do it a subsequent PR. But no reason we can't have that done here, especially with the tests ready.

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/TomcatController.java
```diff
@@ -0,0 +1,188 @@
...
+      createOrUpdateDeployment(tomcat);
+      createOrUpdateService(tomcat);
+    }
+
+    Optional<DeploymentEvent> latestDeploymentEvent =
+        context.getEvents().getLatestOfType(DeploymentEvent.class);
+    if (latestDeploymentEvent.isPresent()) {
+      Tomcat updatedTomcat =
+          updateTomcatStatus(tomcat, latestDeploymentEvent.get().getDeployment());
```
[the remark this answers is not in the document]
adam-sandor: great

metacosm:
> > Maybe we should rename the `samples` module to `samples-simpl`e and the current one to samples, what do you think?
> > its quite ugly to have `sample-operators` and `samples` in the same folder, or move the current simple samples under `simple` module in `samle-operators `
> > @adam-sandor @metacosm what do you think?
> 
> How about integration-test-samples instead of simple-samples? Just an idea

It would make more sense, yes and I was about to suggest something like this. It would make it clearer that the "simple" samples are meant for testing purposes and not necessarily display good practices / patterns.

csviri:
> > Maybe we should rename the `samples` module to `samples-simpl`e and the current one to samples, what do you think?
> > its quite ugly to have `sample-operators` and `samples` in the same folder, or move the current simple samples under `simple` module in `samle-operators `
> > @adam-sandor @metacosm what do you think?
> 
> How about integration-test-samples instead of simple-samples? Just an idea

yep, or `smoke-test-samples`, since we have integration tests already. But either way is good for me.

csviri:
@adam-sandor @metacosm moved the sample to use informers, also some minor changes to make it easy to write. (those are already in v2, or in different way)
renamed the `samples` to `smoke-test-samples`.

csviri (approved):

sample-operators/tomcat-operator/README.md
```diff
@@ -0,0 +1,67 @@
...
+You can build the sample using `mvn install jib:dockerBuild` this will produce a Docker image you can push to the 
+registry of your choice. The JAR file is built using your local Maven and JDK and then copied into the Docker image.
+
+## Install Operator into cluster
+
+Run `kubectl apply -f k8s/crd.yaml` if you haven't already, then run `kubectl apply -f k8s/operator.yaml`. 
+Now you can create Tomcat instances with CRs (see examples above).
+
+## EventSources
+The TomcatController is listening to events about Deployments created by the TomcatOperator by registering a
```
adam-sandor, line 56: This needs to be updated for InformerEventSources too

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/WebappController.java
```diff
@@ -0,0 +1,203 @@
...
+
+  public WebappController(KubernetesClient kubernetesClient) {
+    this.kubernetesClient = kubernetesClient;
+  }
+
+  @Override
+  public void init(EventSourceManager eventSourceManager) {
+    InformerEventSource<Tomcat> tomcatEventSource =
+        new InformerEventSource<>(kubernetesClient, Tomcat.class, t -> {
+          var cache =
```
adam-sandor, line 44: What is this doing here?

sample-operators/tomcat-operator/src/test/resources/log4j2.xml
adam-sandor: We can delete I don't mind

sample-operators/tomcat-operator/README.md
csviri: updated.

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/WebappController.java
csviri: added explanation in a comment:

To create an event to a related WebApp resource and trigger the reconciliation
we need to find which WebApp this Tomcat custom resource is related to.
To find the related UID of the WebApp resource we traverse the cache to
and identify it based on naming convention.

As a remember this was a api call before, with cache this is more efficient, in v2 we address the resources differently there this will be trivial without the cache. But we are looking now into how to make this able to do efficiently in general.

metacosm (approved):

metacosm:
I think we can still make improvements but let's get this merged first and address any remaining issue / improvements in subsequent PRs.

adam-sandor:
🎉

PR #662, opened 2021-11-08
fix: cover informer automatic start case

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
```diff
@@ -82,7 +82,14 @@ private void propagateEvent(T object) {
     }
     uids.forEach(uid -> {
       Event event = new Event(CustomResourceID.fromResource(object));
-      this.eventHandler.handleEvent(event);
+      /*
+       * In fabric8 client for certain cases informers can be created on in a way that they are
+       * automatically started, what would cause a NullPointerException here, since an event might
+       * be received between creation and registration.
+       */
+      if (eventHandler != null) {
```
lburgazzoli, line 90: I think this need some sort of protection as there may be a race conditions between the event being fired and the `eventHandler` being set in the `AbstractEventSource`

csviri: Normally it is `LifecycleAware` , and we "`run`" the informer only on `start`. But there are some cases where the informer is run automtically, and is created outside our scope (what we want, like it is passed as a param). Will address this in an issue for fabric8 client.

lburgazzoli: Yes but my understanding int hat case is that `propagateEvent` may run as per the invocation of the constructor so while the `propagateEvent` in invoked by an informer thread another thread my set the `eventHandler`. 

Since the `eventHandler` is neither volatile nor protected by a lock then the two thread my see a different value for `eventHandler`, right ?

csviri: ahh, good point, changed it to `volatile`. 

But I don't understand the second point:

> Also, since there is a list of events, some events may get propagated and some other not.

All the events are propagated from the list from one thread, so it's either sees the variable set or not, or?

lburgazzoli: assuming that `esourceToUIDs.apply(object)` produces 3 elements, and the `eventHandler` is set once the `uids.forEach` is processing the 2nd element, then the 3rd UID will trigger a reconcile, the first two don't. 

It may be ok but I don't know.

csviri: The thing is with volatile (now set, thx) and `LifecycleAware` all should be covered, if the Informer is created and did not run. Now it can happen that if created in a certain way Informer.run method called automatically in fabric8 - will address this in fabric8 client and a subsequent issue here. 

For now this workaround should mitigate this issues, there can be missed events on startup, but since this is event source and not a the main custom resource event source, it's not that big deal IMHO. Hopefully we can address this before we release v2. But actually will be just a docs from our side to not create an Informer event source when set as paramter. (Or check if the passed event source is running, in case yes log a warning)

lburgazzoli (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/internal/InformerEventSource.java
csviri: Ok it's doable this way:
https://github.com/java-operator-sdk/java-operator-sdk/pull/659/commits/7eb3aa7f26e72ab59be04d87a8937440935d3945

Added warn log message if user passes a running informer.

PR #685, opened 2021-11-16
ErrorStatusHandler - support for better error reporting in status 

csviri:
See linked issue.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ErrorStatusHandler.java
```diff
@@ -0,0 +1,27 @@
+package io.javaoperatorsdk.operator.api.reconciler;
+
+import io.fabric8.kubernetes.client.CustomResource;
+
+public interface ErrorStatusHandler<T extends CustomResource<?, ?>> {
```
metacosm, line 5: Once we have the `HasStatus` (or similar) interface, this probably should use that instead of `CustomResource`

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ErrorStatusHandler.java
csviri: It will be moved to HasMetada if the other PR is merged.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ReconciliationDispatcher.java
```diff
@@ -109,27 +104,76 @@ private boolean shouldNotDispatchToDelete(R resource) {
...
+    }
+  }
+
+  /**
+   * Resource make sense only to clone for the ErrorStatusHandler. Otherwise, this operation can be
+   * skipped since it can be memory and time-consuming. However, it needs to be cloned since it's
+   * common that the custom resource is changed during an execution, and it's much cleaner to have
+   * to original resource in place for status update.
+   */
+  private R cloneResourceIfErrorStatusHandlerIfCouldBeCalled(R resource, Context context) {
```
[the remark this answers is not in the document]
csviri: :D 
will rename it :)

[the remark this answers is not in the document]
csviri: done. :) not sure if much better

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ReconciliationDispatcher.java
```diff
@@ -109,27 +110,76 @@ private boolean shouldNotDispatchToDelete(R resource) {
...
+    }
+  }
+
+  /**
+   * Resource make sense only to clone for the ErrorStatusHandler. Otherwise, this operation can be
+   * skipped since it can be memory and time-consuming. However, it needs to be cloned since it's
+   * common that the custom resource is changed during an execution, and it's much cleaner to have
+   * to original resource in place for status update.
+   */
+  private R cloneResourceIfErrorStatusHandlerIfNeeded(R resource, Context context) {
```
metacosm, line 130: The more I think about it, the less I'm sure we need to clone before updating the status. Presumably, if the resource has changed, the reconciliation will happen again and this update should be cancelled, anyway?

csviri: as mentioned in the docs, it is for the case when the reconciliation executed but at some point an exception thrown, meannwhile in the logic of `reconcile` the status could be updated on the CR. I don't think we should use that, rather the original version of the resource. Since also info could be deleted, it's probably better to work with the fresh copy.

metacosm: Not convinced but I'll trust your judgement on this. :)

metacosm (approved):

PR #693, opened 2021-11-22
Add Webpage Sample to V2 + fix for observed generation

### LINKED ISSUE #683: Add automatic observed generation handling into the samples
Observed generation is not supported out of the bix by the framework, its a nice optimization, should be added to -at least some of the samples, but ideally all.

csviri:
(it's without e2e test yet)

closes #683


sample-operators/webpage/k8s/crd.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apiextensions.k8s.io/v1
```
metacosm, line 1: This shouldn't be needed: the generator should be used for that and what we want people to use instead of manually writing CRDs.

csviri: fixed. removed crd.yaml addd dependency.

metacosm (approved):

PR #697, opened 2021-11-23
Update observed gen on no-update

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ReconciliationDispatcher.java
```diff
@@ -184,14 +186,27 @@ private boolean isLastAttemptOfRetryAndErrorStatusHandlerPresent(Context context
...
+        var observedGen = ((ObservedGenerationAware) status).getObservedGeneration();
+        var currentGen = resource.getMetadata().getGeneration();
+        return !currentGen.equals(observedGen);
+      }
+    }
+    return false;
   }
 
   private void updateStatusObservedGenerationIfRequired(R resource) {
-    // todo: change this to check for HasStatus (or similar) when
```
metacosm, line 193: Why did you remove this comment?

csviri: don't like comments in general but actually though about it and the HasStatus resources would make sense to us only if the standard kuberenetes resources would also implement observed generation aware too. What is quite a big change again in fabric8 client. Also not sure if makes sense for a controller which handles the standard kubernetes objects side by side with it's standard controller to manage observed generations. So IMHO we should support this only for custom resources. But wanted to discuss this with the fabtic8 team too, maybe today.

metacosm: Makes sense. That said, the `todo` comment shows up in IDEA so it's interesting to keep these until we're sure we've addressed what they're referring to, generally.

metacosm (approved):
LGTM

PR #705, opened 2021-11-24
fix: v1-release-process-to-test

csviri:


.github/workflows/release.yml
```diff
@@ -20,20 +25,25 @@ jobs:
...
   # This is separate job because there were issues with git after release step, was not able to commit changes. See history.
   update-working-version:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        if: ${{ startsWith(github.event.release.tag_name, 'v1.' ) }}
+        with:
+          ref: "v1"
+      - uses: actions/checkout@v2
+        if: ${{ startsWith(github.event.release.tag_name, 'v2.') }}
```
metacosm, line 46: Do we really need that part? I mean v2 is now master so we should only have a special for v1 now.

csviri: good question, this forces v2 but don't have strong opinion in this. Can remove it.

csviri: Ahh you mean the if part? Yes that is needed so its no double checkout for v1

Issue #655, opened 2021-11-04
Summary of Proposed Changes for Naming and Related Refactoring for V2


csviri:
Based on comments and other remarks, but also from the usage experience, for v2 we plan to do changes in the public and Internal APIs, mainly some structural and naming changes on the main classes. Some of these changes are also result of the functional changes of v2 (like removing events from context). This issues is a summary of current planned renamings and the related open questions. (see also the linked PR)

We would be more then happy to receive some feedback, if the new model and naming is intuitive enough.

**Public API Changes:**

1. `ResourceController` -> `Reconciler`    - as mentioned on related issues the Controller is a higher level concept and it's defintelly higher level in controller-runtime for example (https://book.kubebuilder.io/architecture.html ), what in our case ResourceController was doing in v1 is implementing reconciliation logic (and defining event sources). So we renamed it to `Reconciler`. 

2. Removed `init(EventSourceManager)` method from `Reconciler`. The original intention was to allow to do some initializations and mainly register event sources by implementing this method. However, it is only used to register event sources and access the `CustomResourceEventSource` - this is required in some corner cases in our experience. So we removed the `init` method and created a new interace `EventSourceInitializer` with `prepareEventSources(EventSourceRegistry<T> eventSourceRegistry)` to implement. So if Reconciler want's to register event sources needs to implement this method. Hopefully this better describes the domain. (We might further evaluate event source registration approach, since Reconciler might not be the best place to do that.)

3. `@Controller` -> ???  - the Controller annotation name made more sense on the `ResourceController`. What it really does it provides configuration values for the controller (not really to the `Reconciler`), ideally it would have the name `@ControllerConfiguration` but we already have an interface elsewhere with that name. So it is still open how to rename this annotation if rename at all.

4. `EventSourceManager` to `EventSourceRegistry` - see below.

**Internal Changes:**

1. `ConfiguredController` -> `Controller`  - after also a  bit of refactoring the `ConfiguredController` is a high level aggregate that mostly directly (some indirectly) encapsulates all the classes regarding to one registered controller. Also now that the `ResourceController` is renamed to `Reconciler` the `ConfiguredController` with it's original meaning does not make that much sense. Note that also after refactoring the controller is responsible of starting and stoping most of the major components from one place, what is now much cleaner.

[...]

PR #716, opened 2021-11-25
refactor: rename internal package to source, moving LifecycleAware

csviri:
package re-naming and moving lifecycle aware


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -88,11 +89,14 @@ private EventProcessor(ResourceCache<R> resourceCache, ExecutorService executor,
     this.retry = retry;
     this.resourceCache = resourceCache;
     this.metrics = metrics != null ? metrics : Metrics.NOOP;
-    this.eventMarker = eventMarker;
+    this.eventMarker = new EventMarker();
+    this.retryAndRescheduleTimerEventSource = new TimerEventSource<>();
```
csviri, line 93: I don't know, but the thinking was that EventSourceManager encapsulates all the event sources, so manages also the lifecycle etc of the. This breaks that logic.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -20,15 +19,14 @@
...
-import io.javaoperatorsdk.operator.processing.event.EventSourceRegistry;
+import io.javaoperatorsdk.operator.processing.event.source.EventSourceRegistry;
 
 public class Controller<R extends HasMetadata> implements Reconciler<R>,
     LifecycleAware, EventSourceInitializer<R> {
   private final Reconciler<R> reconciler;
   private final ControllerConfiguration<R> configuration;
   private final KubernetesClient kubernetesClient;
   private EventSourceManager<R> eventSourceManager;
-  private EventProcessor<R> eventProcessor;
```
csviri, line 31: The idea as desribed here is that the controller maneges the lifecycle of every other component top level component:
https://github.com/java-operator-sdk/java-operator-sdk/issues/655

So it is the highest level agregate, therefore will be no multiple layers for managing LifecycleAware

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
metacosm: I understand but that makes the code more complex. I tend to think that things that are only needed in one spot should be managed there.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: The problem is that it doesn't really do anything with it, it's just needed by the event manager so I don't think we should expose it since the lifecycle of the `EventProcessor` is tied to the `EventManager` (you cannot have one without the other). Making `EventManager` handle the `EventProcessor` life cycle also allows to avoid the rather ugly setters and prevent forgetting to start/stop one or the other. 
If the `Controller` was using the `EventProcessor` directly, it would make sense to keep there. At the very least, if we wanted to keep the top-level aggregate concept, the `EventManager` should retrieve the `EventProcessor` from the `Controller` (or vice-versa) and that's actually what I tried doing yesterday without complicating things even more…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -25,38 +25,25 @@
...
+  EventSourceManager(EventProcessor<R> eventProcessor) {
+    this.eventProcessor = eventProcessor;
+    controller = null;
   }
 
   public EventSourceManager(Controller<R> controller) {
-    init();
+    this.controller = controller;
     controllerResourceEventSource = new ControllerResourceEventSource<>(controller);
+    this.eventProcessor = new EventProcessor<>(this);
```
csviri, line 40: The event source manager should not instantiate the processor. I that the controller manages the lifecycle is much cleaner. This is what was moved out as part of the design changes.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
csviri: This is kinda a `double-edged sword`, since on the other hand it brings complexity in terms of managing the life cycle of an event source in the processor, which is kinda smell if we have already a manager for them. So those are managed in one place.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
metacosm: This doesn't make things cleaner: look at how ugly the `setEventHandler` / `setEventManager` mess is.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
csviri: But in general the Controller does not do anything just instantiates and starts/stops the major components. The circular dependency is ugly, but the EventManager is just a class to encapsulate a list of EventSources. For me that it creates a central component, probably the most complex part is quite a smell.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
csviri: except for those :D

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
metacosm: The problem is that `EventManager` and `EventProcessor` should really be a single object as is evidenced by their tied lifecycle and the need to set a reciprocal relations between them. If we insist on splitting them, then they should each manage the parts that only they use.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: Except that `EventSourceManager` and `EventProcessor` are not really separate components. Again, one cannot exist without the other. If we want to keep them separate, we should make it as cleanly as possible and I think the current solution is cleaner because it avoids the mess of having incomplete objects for no good reason. Maybe we should combine `EventSourceManager` and `EventProcessor`, renaming them to `EventManager`?

csviri: The problem with that is even now the EventProcessor is too big, with bunch of logic, what little bugs me. Actually wanted to somehow break it down to more components.

metacosm: It is indeed quite complex, but some things are 😄
It doesn't make sense to split things that are not coherent if we end up with more complexity to manage the different pieces. I could see the `TimerEventSource` moving back to `EventSourceManager` but I really think that we should tie the lifecycle of `EventSourceManager` and `EventProcessor` by making `EventSourceManager` handle the processor's lifecycle because, again, the manager cannot work without the processor.

csviri: Then the event source manager will handle the lifecycle of event processor? 
ok that is fine by me, we can evaluate further later.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
metacosm: Reverted.

PR #722, opened 2021-11-30
refactor: sample code refactoring

heesuk-ahn:
### Description

I refactored the sample`tomcat-operator` to make it easier to see. :) 

I think it would be good to add more detailed actions & workflow to the README later after refactoring.

cf) after refactoring, the operation was verified by running the e2e test.

### Proposal

* separate tomcat and webapp packages
* separate EventInitializer from Reconciler into abstract class


csviri:
thx @heesuk-ahn will take a look asap

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/webapp/WebappReconciler.java
```diff
@@ -20,42 +19,26 @@
...
-import io.javaoperatorsdk.operator.processing.event.source.InformerEventSource;
+import io.javaoperatorsdk.operator.sample.tomcat.resource.Tomcat;
+import io.javaoperatorsdk.operator.sample.webapp.resource.Webapp;
+import io.javaoperatorsdk.operator.sample.webapp.resource.WebappStatus;
 
 import okhttp3.Response;
 
 @ControllerConfiguration
-public class WebappReconciler implements Reconciler<Webapp>, EventSourceInitializer<Webapp> {
+public class WebappReconciler extends WebappEventSourceInitializer implements Reconciler<Webapp> {
```
csviri, line 31: I'm kinda against inheritance in service layer, (it's harder to unit test test mainly in general). Can be make this an composition rather? SO implement the Initializer and maybe just instatiate a class there where you pass the event registry?

csviri (changes requested):

csviri:
pls just build the project before you push, it fixes the formatting issues

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/webapp/WebappReconciler.java
heesuk-ahn: Thanks for the nice comment!

That's right, it is advantageous for unit tests later to solve it by combination rather than inheritance.

I think what if we proceed with EventInitializer in Controller ?

To make that possible, we'll probably need some change in some APIs.

for example, users can inject their own custom event sources if they wish. Using the API like below.

```
final var operator = new Operator(client, DefaultConfigurationService.instance());
final List<EventSourceInitializer<?>> eventSourceInitialzers = ...

operator.register(reconciler, eventSourceInitialzers);
```

Then, `io.javaoperatorsdk.operator.processing.Controller` receives EventSourceInitializer parameter and passes `eventSourceManager` to process it.

With this improvement, I think that Reconciler can de-couple with EventSourceInitilizer.

heesuk-ahn: If we proceed with this improvement, it seems that we can make an issue separately and apply it to the sample code after processing.

If you agree with this change, I'll create a separate issue and send a pull-request again. :)

heesuk-ahn: Maybe with this improvement we can get rid of this code?

https://github.com/java-operator-sdk/java-operator-sdk/blob/efc0f058c6b0b93a06327a9f47aadb308b67dd18/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java#L171-L173

csviri: You can just pass the registry to an another class currently. 

I agree that is not the nicest way have as it is currently, but note that usually you want to reference the instances of the event source from the reconciler to access the cache during the reconciliation. With just passing them separately we have to solve that part too. That why the current approach is kind a compromise.

csviri: Another problem with this sample is that owner references in it. That could be fixes too :)

csviri: Maybe discuss this problem on Thursday community meeting more in depth?

heesuk-ahn: I have an appointment this week, so it's difficult to attend, but I'll try to attend the next meeting. :) 

In this pull-request, I think about the idea and apply it so that it can be used in a composition way for de-coupling `reconciler` and `eventsoruce` :)

csviri: Not sure if I follow the last sentence. So what I see here, is just extracting a method to a subclass, not sure if this has any strong benefit.

heesuk-ahn: yeq, ok First, it was changed to simply inject and use eventInitilzer. :)

heesuk-ahn: https://github.com/java-operator-sdk/java-operator-sdk/pull/722/commits/2d81fe9359264876649ff30a632caa0fc508ce61

However, this refactoring feels like a Delegate, since we currently have to implement EventSourceInitializer in our Reconciler.
 There don't seem to be any major strengths. so, we can also rollback this part to a previous version. 
 what do you think? 🤔

csviri: yes, that is neither a very useful simplification, having just the method implemented as was originally is good enough with this architecture I think.

heesuk-ahn: I agree. I deleted the part separated by EventSourceInitilizer and rolled back to the originally way.

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/webapp/WebappReconciler.java
```diff
@@ -48,11 +52,20 @@ public void prepareEventSources(EventSourceRegistry<Webapp> eventSourceRegistry)
           // we need to find which WebApp this Tomcat custom resource is related to.
           // To find the related customResourceId of the WebApp resource we traverse the cache to
           // and identify it based on naming convention.
-          return eventSourceRegistry.getControllerResourceEventSource().getResourceCache()
```
csviri, line 51: Are you up to date with master? I think this part should not be changed

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/webapp/WebappReconciler.java
```diff
@@ -48,11 +52,20 @@ public void prepareEventSources(EventSourceRegistry<Webapp> eventSourceRegistry)
...
-              .list(webApp -> webApp.getSpec().getTomcat().equals(t.getMetadata().getName()))
-              .map(ResourceID::fromResource)
+          var webAppInformer =
```
csviri, line 55: Are you up to date with master, this should  not be changed probably.

heesuk-ahn: Is it okay to close and reopen this pull-request? The refactoring is a little different from what I originally intended, so I think it would be better to re-pullrequest it again.

csviri: sure, np

PR #723, opened 2021-11-30
fix: Update examples in README to have consistent class names

katheris:
Update the example in the README.md file so that class names used in the Controller/Reconciler class are consistent with the class names used in the other example classes.

I noticed this inconsistency when trying out the java operator-sdk for the first time and found it confusing.

Signed-off-by: Katherine Stanley <11195226+katheris@users.noreply.github.com>


csviri:
@katheris thank you!!

PR #731, opened 2021-12-02
feature: update observed generation on updateResource

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -157,6 +157,9 @@ private R cloneResourceForErrorStatusHandlerIfNeeded(R resource, Context context
       updatedCustomResource = updateStatusGenerationAware(updateControl.getResource());
     } else if (updateControl.isUpdateResource()) {
```
metacosm, line 158: Can't that `if/else` be simplified?

csviri: Good question, could try magically transform this, but what I tried to achieve here is just break it down based on the return cases from the controller. So it's obvious what happens on those cases. If you have any suggestion pls let me know.

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
heesuk-ahn: If `Enum` is used to express Status in UpdateControl, it seems that if/else can be refactored  🤔 .

```
public class UpdateControl<T extends HasMetadata> extends BaseControl<UpdateControl<T>> {

  private final T resource;
  private final boolean updateStatus;
  private final boolean updateResource;
  private final UpdateControlStatusEnum controlStatusEnum; 
  ...
```

```
public enum UpdateControlStatusEnum {

  UPDATE_RESOURCE_AND_STATUS {
    @Override
    public <R extends HasMetadata> R toUpdatedResource(....) {
            return new UpdateResourceAndStatus(...);
        }
  },
  UPDATE_RESOURCE {
    @Override
    public <R extends HasMetadata> R toUpdatedResource(....) {
            return new UpdateResource(...);
        }
  },
  UPDATE_STATUS {
    @Override
    public <R extends HasMetadata> R toUpdatedResource(....) {
            return new UpdateResource(...);
        }
  }
  
  // abstract method for business logic per status..
  // public abstract <R extends HasMetadata> R toUpdatedResource(...)
  
  // private static class UpdateResourceAndStatus implements UpdateCustomResource { .. }
  // private static class UpdateResource implements UpdateCustomResource { .. }
  // private static class UpdateStatus implements UpdateCustomResource { .. }
  // ...
```

Then in that class, you can use it like this:

```
class ReconciliationDispatcher<R extends HasMetadata> {
.
.
  private PostExecutionControl<R> reconcileExecution(ExecutionScope<R> executionScope,
      R resourceForExecution, R originalResource, Context context) {
   
    // REMOVED IF/ELSE...
    var updateResource = updateControl.getControlStatusEnum.toUpdateResoruce(...);
    
    return createPostExecutionControl(updateResource, updateControl);
}

```

If refactoring is needed for this part, I think it can be improved in other issues and PullRequest. :)

metacosm: I'd be interested in seeing how this would work in a more fleshed-out PR, yes. Is that something you could work on, @heesuk-ahn?

heesuk-ahn: @metacosm Yes, I will make an issue about this and work on it separately and upload the pr. :)

PR #736, opened 2021-12-06
fix: issue with event source start ordering

csviri:


metacosm (commented):
This is addressed in the dependent source PR

csviri:
Ok since it's a bug fix of the current scope will merge it, we overwrite it with that eventually.

PR #742, opened 2021-12-08
Error status handler improvements

csviri:


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcherTest.java
```diff
@@ -56,9 +58,12 @@ void setup() {
     when(configuration.getName()).thenReturn("EventDispatcherTestController");
     when(configService.getMetrics()).thenReturn(Metrics.NOOP);
     when(configuration.getConfigurationService()).thenReturn(configService);
-    when(configService.getResourceCloner()).thenReturn(ConfigurationService.DEFAULT_CLONER);
-    when(reconciler.reconcile(eq(customResource), any()))
-        .thenReturn(UpdateControl.updateResource(customResource));
+    when(configService.getResourceCloner()).thenReturn(new Cloner() {
```
metacosm, line 61: Why was this changed?

csviri: To make sure to make easier to do `when(..).thenReturn()`  with mockito. Since otherwise its a different instance.

metacosm: Yes, but then you're not really testing the actual behavior?

csviri: This can be understood as a form of a mocking of that clone method / cloner. The goal here is not to test the cloner but only if it is called in case it's relevant.

metacosm: I understand but the problem is what if the behavior changes when the instance is not actually cloned?

csviri: Ahh, I think I see what you mean, will try to improve it.

csviri: Actually not, is just what instance we move around, for some special cases.

docs/documentation/features.md
```diff
@@ -166,19 +166,21 @@ A successful execution resets the retry.
...
 
-    T updateErrorStatus(T resource, RuntimeException e);
+    Optional<T> updateErrorStatus(T resource, RetryInfo retryInfo, RuntimeException e);
 
 }
 ```
 
-The `updateErrorStatus` resource is called when it's the last retry attempt according the retry configuration and the
-reconciler execution still resulted in a runtime exception.
+The `updateErrorStatus` is called in case an exception is thrown from the reconciler. It is also called when
```
metacosm, line 180:
```suggestion
The `updateErrorStatus` method is called in case an exception is thrown from the reconciler. It is also called when
```

metacosm (approved):

PR #758, opened 2021-12-16
fix: resource cache interface for InformerEventSource

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/CachingEventSource.java
```diff
@@ -56,8 +50,8 @@ protected void handleEvent(T value, ResourceID relatedResourceID) {
     }
   }
 
-  public Cache<ResourceID, T> getCache() {
-    return cache;
+  public Map<ResourceID, T> getCache() {
```
metacosm, line 53: I think this is too restrictive… I think we should probably simplify `ResourceCache` and return that instead maybe?

csviri: Not sure if following, we return all the values, can it be less restrictive? 

On the other hand the resource cache is now about K8S resource, and this class is dedicated to non k8s resources. So not sure if we should unify those. Those have totally different requirements. And approach how we are handling them.

csviri: Or maybe just a very simple way. 

Like having 
` getValue(ResourceID id)` ?

metacosm: This will be addressed in a subsequent PR, let's merge this as-is for now.

metacosm (approved):

PR #760, opened 2021-12-16
API changes proposal

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
```diff
@@ -1,10 +1,11 @@
...
 
+import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.processing.LifecycleAware;
-import io.javaoperatorsdk.operator.processing.event.EventHandler;
 
-public interface EventSource extends LifecycleAware {
+public interface EventSource<P extends HasMetadata> extends LifecycleAware {
 
-  void setEventHandler(EventHandler eventHandler);
+  void setEventRegistry(EventSourceRegistry<P> registry);
```
csviri, line 8: Here the method naming is not good. But don't get this change, why should an event source be aware of the whole registry?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/UpdatableResourceCache.java
```diff
@@ -0,0 +1,10 @@
+package io.javaoperatorsdk.operator.processing.event.source.controller;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+public interface UpdatableResourceCache<T extends HasMetadata> extends ResourceCache<T> {
```
csviri, line 6: This is not used anywhere. Where it will be?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSourceRegistry.java
```diff
@@ -16,11 +17,12 @@
...
-  void registerEventSource(EventSource eventSource)
+  void registerEventSource(EventSource<T> eventSource)
       throws IllegalStateException, OperatorException;
 
-  Set<EventSource> getRegisteredEventSources();
+  Set<EventSource<T>> getRegisteredEventSources();
 
   ControllerResourceEventSource<T> getControllerResourceEventSource();
 
+  EventHandler getEventHandler();
```
csviri, line 27: Why should be the registry aware of the handler?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/CachingEventSource.java
```diff
@@ -69,4 +67,9 @@ public void stop() throws OperatorException {
     super.stop();
     cache.close();
   }
+
+  @Override
+  public T getAssociated(P primary) {
```
csviri, line 72: Why it needs the whole resource? I think this should be ideally as simple as possible API and just use the ResourceID

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/AbstractEventSourceTest.java
```diff
@@ -0,0 +1,46 @@
+package io.javaoperatorsdk.operator.processing.event.source;
+
+import org.junit.jupiter.api.AfterEach;
+
+import io.javaoperatorsdk.operator.processing.event.EventHandler;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class AbstractEventSourceTest<S extends EventSource, T extends EventHandler> {
```
csviri, line 10: This should be a Junit extension, as for Integration tests. I think there should be no inheritance in unit tests.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
metacosm: Good catch on the name. Why wouldn't an event source be aware of where it's registered? The idea here is that knowing the registry allows an `EventSource` to query the registry for other resources without having to hold a reference to the other event sources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/UpdatableResourceCache.java
metacosm: Ah, that shouldn't have been committed! 😅

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSourceRegistry.java
metacosm: All `EventSources` registered with the same `EventRegistry` should use the same `EventHandler` so that's how they get which `EventHandler` to use. This makes the link between the registry, sources and event handler slightly more explicit. I do agree that this is kind of not strictly related to the pure registry aspects of the interface. The problem is that the abstraction is not really appropriate to reflect the reality of the implementation.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/CachingEventSource.java
metacosm: It's just generalizing the method that's already present on `InformerEventSource`. This method will always be called in a context where the primary is available anyway so why restrict it to only providing the `ResourceID`?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/AbstractEventSourceTest.java
metacosm: Feel free to change that.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/EventSourceInitializer.java
```diff
@@ -1,17 +1,25 @@
...
-   * CustomResourceEventSource, what might be useful for some edge cases.
+   * Prepares a list of {@link EventSource} implementations to be registered by the SDK.
    * 
-   * @param eventSourceRegistry the {@link EventSourceRegistry} where event sources can be
-   *        registered.
+   * @param primaryCache a cache providing direct access to primary resources so that event sources
+   *        can extract relevant information from primary resources as needed
    */
-  void prepareEventSources(EventSourceRegistry<T> eventSourceRegistry);
+  List<EventSource> prepareEventSources(ResourceCache<P> primaryCache);
```
csviri, line 23: Shouldn't we wrap the cache into some context? So its easily and backwards compatibly extendable in future?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/CachingEventSource.java
csviri: not sure how we agreed on this, I still think it should be ResourceID :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/EventSourceInitializer.java
metacosm: Done.

csviri (approved):
Let's merge this, will work on followup polishing

PR #775, opened 2021-12-23
fixes and api improvements

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -6,5 +6,9 @@
 
   Optional<RetryInfo> getRetryInfo();
 
-  <T> T getSecondaryResource(Class<T> expectedType, String... qualifier);
+  default <T> T getSecondaryResource(Class<T> expectedType) {
```
csviri, line 9: @metacosm is this ok with you, kind would like it to limit it in this scope. Since more qualifiers does not make sense, and if dependent resources will use the previous way it will be backwards compatible anyway. So it can be changed back without breaking the api.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventSourceManagerTest.java
```diff
@@ -94,16 +95,16 @@ void retrievingEventSourceForClassShouldWork() {
   }
 
   @Test
-  void shouldNotBePossibleToAddEventSourcesForSameTypeAndQualifier() {
+  void shouldNotBePossibleToAddEventSourcesForSameTypeAndName() {
     EventSourceManager manager = initManager();
 
     CachingEventSource eventSource = mock(CachingEventSource.class);
-    when(eventSource.getResourceClass()).thenReturn(String.class);
+    when(eventSource.getResourceClass()).thenReturn(InformerEventSource.class);
```
metacosm, line 102: That's not correct. `getResourceClass` is supposed to return the class handled by the event source. So if you specify that the resource class is `InformerEventSource`, this means that your event source deals with `InformerEventSources` which doesn't really make sense. Granted, `String` doesn't make a lot more sense either but it's a little bit more realistic and doesn't risk clashing with the `EventSource` class itself (which should be reserved for cases where there is no associated type with an `EventSource`).

metacosm (changes requested):

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventSourceManagerTest.java
csviri: Ohh, yes, sorry my fault, will change it!

csviri: fixed

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/sample/simple/TestCustomResource.java
```diff
@@ -2,10 +2,12 @@
 
 import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Kind;
 import io.fabric8.kubernetes.model.annotation.Version;
 
 @Group("sample.javaoperatorsdk.io")
 @Version("v1")
+@Kind("TestCustomResource")
```
metacosm, line 10: Why is this needed?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/sample/simple/TestCustomResourceV2.java
```diff
@@ -2,10 +2,12 @@
 
 import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Kind;
 import io.fabric8.kubernetes.model.annotation.Version;
 
 @Group("sample.javaoperatorsdk.io")
 @Version("v2")
+@Kind("TestCustomResource")
```
metacosm, line 10: OK, I guess you added it to the V1 version for symmetry?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/sample/simple/TestCustomResource.java
csviri: This is needed to compare resources when we register a new reonciler. There was a tests for the registration of the same resource with multiple versions - what should fail. It did not fail without this, could not determine the kind. It was not failing until now because there were junit4 annotations used on test class

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/sample/simple/TestCustomResourceV2.java
csviri: same as above

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/sample/simple/TestCustomResource.java
csviri: see `ControllerManagerTest.java` changes

csviri: If that functionality should work without this annotation, that should be a separate issues IMHO.

metacosm: It shouldn't be needed here. Just for the v2.

metacosm (approved):

PR #782, opened 2022-01-03
Secondary resource access with Optional

csviri:
This PR should be reviewed first:
https://github.com/java-operator-sdk/java-operator-sdk/pull/775

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
@@ -6,9 +6,9 @@ public interface Context {
 
   Optional<RetryInfo> getRetryInfo();
 
-  default <T> T getSecondaryResource(Class<T> expectedType) {
+  default <T> Optional<T> getSecondaryResource(Class<T> expectedType) {
     return getSecondaryResource(expectedType, null);
   }
 
-  <T> T getSecondaryResource(Class<T> expectedType, String eventSourceName);
+  <T> Optional<T> getSecondaryResource(Class<T> expectedType, String eventSourceName);
 }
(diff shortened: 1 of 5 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -6,5 +6,9 @@
 
   Optional<RetryInfo> getRetryInfo();
 
-  <T> T getSecondaryResource(Class<T> expectedType, String... qualifier);
+  default <T> Optional<T> getSecondaryResource(Class<T> expectedType) {
```
metacosm, line 9: Not sure I like this change…

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/WebappReconciler.java
```diff
@@ -70,13 +71,13 @@ public WebappReconciler(KubernetesClient kubernetesClient) {
...
-    Tomcat tomcat = context.getSecondaryResource(Tomcat.class);
-    if (tomcat == null) {
+    Optional<Tomcat> tomcatValue = context.getSecondaryResource(Tomcat.class);
+    if (tomcatValue.isEmpty()) {
       throw new IllegalStateException("Cannot find Tomcat " + webapp.getSpec().getTomcat()
           + " for Webapp " + webapp.getMetadata().getName() + " in namespace "
           + webapp.getMetadata().getNamespace());
     }
-
+    Tomcat tomcat = tomcatValue.get();
```
metacosm, line 80: This makes the usage a little bit more complex, though I guess it forces users to handle the `null` case. This is something I often wonder about actually when designing an API: should we use `Optional` in similar cases or follow a `Map`-like convention of returning `null` if the element is not present?

csviri: IMHO, map like convention was there before `Optional` was present in Java. It make sense and happens in multiple examples that the secondary resources is checked for null. It should be, since it's not necessary present, if not created yet for example. For that case it's better to use `Optional` I think. Exactly as you say, that it forces users to check, also to eliminate `null` checks

metacosm (approved):

PR #786, opened 2022-01-04
mysql e2e

csviri:


csviri:
There seems to be a small issue with `PerResourceEventSource` when deleting the CR, will take a look on a separate PR since not closely related

metacosm (approved):
LGTM, not familiar with this example so just took a cursory look. Trusting you and the tests passing on this! 😄

PR #796, opened 2022-01-05
feat: provide KubernetesClient access from context

metacosm:


csviri:
@metacosm are you sure this is needed? This is for the case when the client would be build by the SDK? (not happening now) Or just a convinietn feature?
Not enought to pass it to the constructor?
Also should we put it also to the context?

metacosm:
This is to enable use cases like this: https://github.com/java-operator-sdk/java-operator-sdk/pull/785/files#diff-b2e8a6070866bd0d784849eb8fd8fc075c14c78b91ec07b502c339a4d45a7056 where basically the client is not available where the object is created.

csviri (approved):

PR #801, opened 2022-01-07
feature: sonar for builds

csviri:


csviri:
@metacosm what about merging this, in case we will introduce new issues we will se it. And will take a look on the major ones now.

PR #809, opened 2022-01-11
feature: flag to allow to not close k8s client on stop

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -24,16 +24,24 @@
...
   private static final Logger log = LoggerFactory.getLogger(Operator.class);
   private final KubernetesClient kubernetesClient;
+  private final boolean closeKubernetesClientOnStop;
   private final ConfigurationService configurationService;
   private final ControllerManager controllers = new ControllerManager();
 
+
   public Operator(ConfigurationService configurationService) {
-    this(new DefaultKubernetesClient(), configurationService);
+    this(new DefaultKubernetesClient(), configurationService, true);
```
metacosm, line 33: Maybe the default should be `false`, no?

csviri: Mentioned it in the issues (linked) IMHO usually (at least in our samples) it is used this way, that the registered hook shuts down the operator and also the client.

metacosm: Ok, we need to document it, I guess :)

metacosm: Actually, can we put this configuration option on the `ConfigurationService`? I think it makes sense to put it there and would avoid creating another constructor.

csviri: good point, will change it a look how it would look like

csviri: updated

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -118,4 +118,8 @@ default Metrics getMetrics() {
   default ExecutorService getExecutorService() {
     return Executors.newFixedThreadPool(concurrentReconciliationThreads());
   }
+
+  default boolean getCloseClientOnStop() {
```
metacosm, line 122: How about just `closeClientOnStop`? I always find it weird to see `getDoSomething` methods :)

csviri: agree, changed

metacosm (approved):

PR #811, opened 2022-01-11
fix: only check for CRD if we're dealing with CustomResources

### LINKED ISSUE #810: CRD checks should be bypassed when reconciling native Kubernetes resources
Right now, the SDK might check that the CRD associated with resources handled by a given `Reconciler` are indeed present on the cluster. This, however, only makes sense if we're reconciling custom resources as it will fail for native Kubernetes resources.

metacosm:
Fixes #810

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
@@ -6,6 +6,7 @@ import java.util.Objects;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
+import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.client.KubernetesClient;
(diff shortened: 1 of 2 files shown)


csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -167,10 +168,10 @@ public void start() throws OperatorException {
     try {
       // check that the custom resource is known by the cluster if configured that way
       final CustomResourceDefinition crd; // todo: check proper CRD spec version based on config
-      if (configurationService().checkCRDAndValidateLocalModel()) {
-        crd =
-            kubernetesClient.apiextensions().v1().customResourceDefinitions().withName(crdName)
-                .get();
+      if (configurationService().checkCRDAndValidateLocalModel()
```
csviri, line 171: Can we add a unit test?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: Done.

csviri (approved):

PR #814, opened 2022-01-11
fix: generate proper resource name & finalizer for group-less resources

### LINKED ISSUE #812: Resource names are not currently properly generated
## Bug Report

Resource names for historic resources such as `Pods` or `Secrets` are not currently properly generated. This is fixed in fabric8 client 5.12 but we need a workaround until we can use that version.

#### What did you do?

Creating a reconciler for historic resources leads to failures because the associated resource name is not properly generated. This occurs in particular for automatically generated finalizers where something like `secrets./finalizer` is generated but is an invalid format.

Proper resource names.

N/A

#### Environment

**Kubernetes cluster type:**

`$ java -version`

`$ kubectl version`

metacosm:
Fixes #812

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
@@ -10,9 +10,15 @@ import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 public class ReconcilerUtils {
 
   private static final String FINALIZER_NAME_SUFFIX = "/finalizer";
+  protected static final String MISSING_GROUP_SUFFIX = ".javaoperatorsdk.io";
 
-  public static String getDefaultFinalizerName(String crdName) {
-    return crdName + FINALIZER_NAME_SUFFIX;
+  public static String getDefaultFinalizerName(String resourceName) {
+    // resource names for historic resources such as Pods are missing periods and therefore do not
+    // constitute valid domain names as mandated by Kubernetes so generate one that does
+    if (resourceName.indexOf('.') < 0) {
+      resourceName = resourceName + MISSING_GROUP_SUFFIX;
+    }
+    return resourceName + FINALIZER_NAME_SUFFIX;
   }
 
   public static String getNameFor(Class<? extends Reconciler> reconcilerClass) {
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
```diff
@@ -10,9 +10,15 @@
 public class ReconcilerUtils {
 
   private static final String FINALIZER_NAME_SUFFIX = "/finalizer";
+  protected static final String MISSING_GROUP_SUFFIX = ".javaoperatorsdk.io";
 
-  public static String getDefaultFinalizerName(String crdName) {
-    return crdName + FINALIZER_NAME_SUFFIX;
+  public static String getDefaultFinalizerName(String resourceName) {
```
csviri, line 15: Could we add a unit test pls?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -17,7 +17,15 @@ default String getName() {
   }
 
   default String getResourceTypeName() {
-    return HasMetadata.getFullResourceName(getResourceClass());
+    return getResourceTypeName(getResourceClass());
+  }
+
+  static <T extends HasMetadata> String getResourceTypeName(Class<T> resourceClass) {
```
csviri, line 23: Same here, pls add unit test.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
csviri, line 15: Pls add unit test

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
metacosm: This is a workaround for a fix that will be in fabric8 5.12 where it's already pretty well tested.

csviri: But if too much effort, I'm ok to skip it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
metacosm: Done.

csviri (approved):

PR #828, opened 2022-01-13
feature: move to one or all namespaces WIP

csviri:


csviri:
For now we going to support multiple namespaces in configs.

PR #841, opened 2022-01-17
feature: ResourceID from first owner reference

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
```diff
@@ -13,6 +13,15 @@ public static ResourceID fromResource(HasMetadata resource) {
         resource.getMetadata().getNamespace());
   }
 
+  public static Optional<ResourceID> fromFirstOwnerReference(HasMetadata dependentResource) {
```
metacosm, line 16: It'd be better if you'd use the implementation from the dependent resources branch: https://github.com/java-operator-sdk/java-operator-sdk/pull/785/files#diff-5166f926d491fa8bc37664176eb65ecb8dc051b8db0e3d3771fb817a1fbd462dR70-R80 so that merging later will be easier.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
csviri: good point, changed

metacosm (approved):

PR #842, opened 2022-01-17
fix: remove deprecated api

csviri:
A really small PR


metacosm:
We've tried this before, it has never worked and we probably should look into why

csviri:
> We've tried this before, it has never worked and we probably should look into why

Fixed now, it's a strange new behavior with the client. But it's not bad, in terms that the status should be always the result of the last reconciliation.

metacosm (approved):
LGTM

PR #846, opened 2022-01-18
feature: add from first reference to ResourceID too

csviri:
see: https://github.com/strimzi/kafka-access-operator/pull/2/files#r786855066
It is useful situations like this.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
@@ -13,6 +13,16 @@ public class ResourceID implements Serializable {
         resource.getMetadata().getNamespace());
   }
 
+  public static Optional<ResourceID> fromFirstOwnerReference(HasMetadata resource) {
+    var ownerReferences = resource.getMetadata().getOwnerReferences();
+    if (!ownerReferences.isEmpty()) {
+      return Optional.of(new ResourceID(ownerReferences.get(0).getName(),
+          resource.getMetadata().getNamespace()));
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
```diff
@@ -13,6 +13,16 @@ public static ResourceID fromResource(HasMetadata resource) {
         resource.getMetadata().getNamespace());
   }
 
+  public static Optional<ResourceID> fromFirstOwnerReference(HasMetadata resource) {
```
metacosm, line 16: I would still put this method on the `Mappers` class because it's not easily findable here, imo.

csviri: Based on the current situation it would fit also there. But `ResourceID` **already has** `fromResource` static initializer. A lean usually to have those methods directly in the class, basically an utility class is always harder to find (actually already a problem if there is need to search for it). 
Currently Mapper contains static init methods for `PrimaryResourceRetriever`, these might be also directly on that interface, so its obviouse how to create them and not search for `Mappers`.

So I kinda lean towards this, to have it as it is now.

metacosm: Ah yes, and the method doesn't have the same exact semantics so yes, it makes kind of sense to keep it on `ResourceID`. Diff reviews sometimes make you miss context… 🤦🏼

metacosm (approved):

PR #852, opened 2022-01-20
More principled way to wait for the CRD

andreaTP:
> this fixes an issue with CRD registration

The underlying problem is that with a small/new cluster `etcd` takes some time to propagate the state.

Unfortunately I couldn't reproduce on my machine(any suggestion on how to reproduce is welcome!), but this should be a more "kubernetes-client" idiomatic way to obtain the same result. 🙂


csviri:
Yes, I saw this before as well.

andreaTP:
Sonar is not going to succeed @csviri :

```
Error:  Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.9.1.2184:sonar (default-cli) on project java-operator-sdk: You're not authorized to run analysis. No sonar.login or SONAR_TOKEN env variable was set
```

csviri:
np, this is known issue, will take a look, thx! will merge this

PR #859, opened 2022-01-21
chore(ci): rename snapshots workflow, add next, only keep unit tests

metacosm:
This should be cherry-picked to `main` as well.


csviri (approved):
LGTM

Issue #848, opened 2022-01-19
Periodic Reconciliation - Able to Set Max Interval Between Reconciliations


csviri:
In some cases, like when managing external resources, or just having a simple controller that we want to periodically reschedule it would be handy to be able to define that  "I want to have maximum delay [X time] between reconciliations".
This would be much better than fixed rate, since that can easily lead to very close executions after each other, what is not the intention in these use cases.

One way that this is already doable is to make sure that the UpdateControl returns with `UpdateControl.<MyCustomResource>noUpdate().rescheduleAfter(8, TimeUnit.HOURS);`. Where it would return just `noUpdate()` - (or alternaives like `updateStatus()` )

But this is not that nice, since basically puts kinda noise in the code.

This could be also a new annoation `@PeriodicReconciliation(maxDelay=8, timeUnit=HOURS)` on the reconciler
or a property of `@ControllerConfuguration(periodicReconiliationMaxDelay=8, periodicReconiliationTimeUnit = HOURS )` 

please let me know what do you think. 

## Implementation notes
Implementation wise regarding the execution, it is very simple to do, we just always put a reschedule if there is missing from the `UpdateControl`.

andreaTP:
I think that `@ControllerConfuguration(periodicReconiliationMaxDelay=8 ...` with sensible defaults works perfectly for this use-case.

metacosm:
I'd favor a configuration option as well. It makes more sense that way than having to remember to call a reschedule on `UpdateControl`…

csviri:
see also: https://kubernetes.slack.com/archives/CAW0GV7A5/p1642594910160600

For the record, I think discussed before, if you have informers in place and correct code, this is not needed. It's really a fail safe.

sclorng:
That would help for testing. Actually, and especialy true on quarkus side, tests are more integration tests than unit testing the Reconciler. Then, it is not easy to test if a reschedule has been set.

Informers can help detecting change but for time based behaviour, I guess we still need to use either TimedEventSource or reschedule.

PR #871, opened 2022-01-26
feature: controller reconciliation max delay

csviri:


docs/documentation/features.md
```diff
@@ -223,6 +222,27 @@ public class DeploymentReconciler
   }
 ```
 
+## Max Interval Between Reconciliations
+
+In case informers are all in place and reconciler is implemented correctly the reconciliation is triggered when needed,
```
andreaTP, line 227: Maybe something like:

```
The reconciliation is supposed to be self-sufficient to cover all the use-cases, however ...
```

andreaTP (approved):
LGTM

docs/documentation/features.md
csviri: will improve the wording, but this does not mean the same, also little bit abstract imho.

csviri: The point is that event sources are needed to have a correct implementation.

docs/documentation/features.md
```diff
@@ -198,8 +196,9 @@ in status) to handle generation filtering in memory. Thus, if an event is receiv
 resource is compared with the resource in the cache.
 
 Note that the **first approach has significant benefits** in the situation when the operator is restarted and there is
-no cached resource anymore. In case two this leads to a reconciliation of every resource, event if the resource is not
-changed while the operator was not running.
+no cached resource anymore. In case two this leads to a reconciliation of every resource in all cases,
```
metacosm, line 199: Would have been better to do these changes in a separate PR.

docs/documentation/features.md
```diff
@@ -223,6 +222,28 @@ public class DeploymentReconciler
   }
 ```
 
+## Max Interval Between Reconciliations
+
+In case informers are all in place and reconciler is implemented correctly, there is no need for additional triggers. 
+However, it's a [common practice](https://github.com/java-operator-sdk/java-operator-sdk/issues/848#issuecomment-1016419966) 
+having a failsafe periodic trigger in place,
```
metacosm, line 229:
```suggestion
to have a failsafe periodic trigger in place,
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -114,4 +115,12 @@ default boolean useFinalizer() {
   default ResourceEventFilter<R> getEventFilter() {
     return ResourceEventFilters.passthrough();
   }
+
+  default long reconciliationMaxInterval() {
```
metacosm, line 119: I think it will be better to introduce a separate annotation if we split the duration and its unit to make the name simpler and make it more coherent because it doesn't make sense to provide one without the other.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
csviri: I'm a little worried that it would unnecessary complicate things; also there is always a default value. Also both start with same prefix, I don't think this will be confusing.

csviri: An alternative is to have the time as a string, like `"10h"` or "140m". But we use this separation in other API already.

docs/documentation/features.md
csviri: it would...will extract it.

metacosm: No need, just for future reference ;)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
metacosm: I don't think it would complicate things and it would make things cleaner to group related things in a single part. Otherwise, you could have annotated reconcilers with just `reconciliationMaxIntervalTimeUnit` and that would look really weird. Currently, the only hint that these things are related are their names. It's better to make it explicit, in my opinion. It also avoids polluting the main configuration with things that will rarely be used. Embedding the unit in the String is a recipe for disaster, imo, and I think it's much cleaner to separate the configuration into a separate annotation.

csviri: So just to see it clear. The reason to separate this into a new annotation because there are two related fields? 
I'm not sure that makes the effort, especially since we everything in one (for now mandatory) annotation what is a good thing. 
We would like to have this feature by default turned on, so what if that annotation is not present? Actually I guess should be turned off? It's easy to miss that it even exists.

csviri: Agree that it nicer, but also has these side effects..so I lean little bit more to the current way. But if you insist we can change it. Just lets clarify what should be the behavior.

metacosm: Which side effects?

csviri: As mentioned, wanted to have this behavior turned on by default. So the developers should turn it off if they are absolutely sure that the operator works as intended (maybe not even that case). But if this is a separate annotation, then it needs to be put there explicitly, otherwise it would be quite contra intuitive to have the functionality turned on without the annotation.

metacosm: The annotation is only needed if you want to change the default values. We already do that in lots of other cases.

The only scenario where things are not clear is if you want to turn it off completely. Then again, that's not clear either with the current implementation.

csviri: Where do we do that? What I mean is, that in spring and other frameworks  a functionality is turned on by adding an annotation, for me it sounds a little smelly that if we have an annotation for that but its not there and still turned on.

I created a constant for that to turn off, or any non negative value would do. It's quite common to do it this way.

csviri: Maybe we should ask the users tbh. @andreaTP @lburgazzoli what do you think, should be this a separate annotation or not:
https://github.com/java-operator-sdk/java-operator-sdk/issues/848

docs/documentation/features.md
csviri: hehe, too late :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
metacosm: The namespace control is done that way. Pretty much all the annotation fields have default values so we already turn things on without the annotation being present. This would be the exact same thing here: the child annotation would just allow people to change the default value.

metacosm: To be clear, I'm not saying we should have a completely separate annotation, just a composite child annotation for this.

csviri: > To be clear, I'm not saying we should have a completely separate annotation, just a composite child annotation for this.

Ahhh, ok that is a completely different story! Sure that makes sense! Sorry....

metacosm: I just understood that I hadn't been clear enough with what I was saying with your last commit :)
Sorry! 😓

docs/documentation/features.md
```diff
@@ -223,6 +221,28 @@ public class DeploymentReconciler
...
+
+In case informers are all in place and reconciler is implemented correctly, there is no need for additional triggers. 
+However, it's a [common practice](https://github.com/java-operator-sdk/java-operator-sdk/issues/848#issuecomment-1016419966) 
+to have a failsafe periodic trigger in place,
+just to make sure the resources are reconciled after certain time. This functionality is in place by default, there
+is quite high interval (currently 10 hours) while the reconciliation is triggered. See how to override this using 
+the standard annotation:
+
+```java
+@ControllerConfiguration(finalizerName = NO_FINALIZER, reconciliationMaxInterval = 2L,
```
metacosm, line 234: This needs to be updated.

metacosm (approved):

metacosm (changes requested):
LGTM apart the doc that needs to be updated. I might open a subsequent PR to mirror the annotation split in the configuration itself later.

metacosm (approved):
Cleaner, I like it! 👍🏼

Issue #94, opened 2020-05-04
Support multiple versions of a CRD


csviri:
Within one CRD there could be multiple versions defined, we need to explicitly support this.

stoetti:
@csviri Is this on the roadmap in the near future?

csviri:
@stoetti yes, this should be actually one of the priorities in the close future. 

pinging also @metacosm

metacosm:
This should actually work already, no? If not, could you explicit a scenario that doesn't work?

csviri:
@metacosm  good question, went though again on the docs:
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning

1. If we see this in a way that we are the client only of the API, and the conversion hooks are properly configured for a custom resource, the migration can be seamlessly done by upgrading the operator to support JUST the most recent version. 
While Kubernetes API should support multiple versions at the same time, the intention is that the clients (in this case the operator) is migrated eventually to the new version. So it's not the case that we have to support 2 versions of the same CR at the same time. 
To be honest I don't fully understand what happens if the conversion strategy is None (maybe worth to try it out).

2, In addition to that, supporting conversion webhooks is out of the scope of the framework IMHO, or could be part of an isolated part of the SDK (not the framework), but that is a longer discussion or an other issue.

So IMHO, if nobody else sees it otherwise (@stoetti ? ) we are actually good on this. There is nothing to be done.  I would even close the issue.

stoetti:
I gave it a try without using the conversionStrategy 'None'. In that case k8s just changes the apiVersion to the version thats marked with stored=true. This obviously leads to a deserialization-error if there is an incompatible change in the CRD.

I will work on a sample with a conversion webhook and see if there are any aspects that are worth building into a subproject. From the documentation the conversion-webhook is just a simple REST endpoint serving a specific interface.

csviri:
> I gave it a try without using the conversionStrategy 'None'. In that case k8s just changes the apiVersion to the version thats marked with stored=true. This obviously leads to a deserialization-error if there is an incompatible change in the CRD.

yes, just don't know how we would support that, probably does not make sense. 


> I will work on a sample with a conversion webhook and see if there are any aspects that are worth building into a subproject. From the documentation the conversion-webhook is just a simple REST endpoint serving a specific interface.

[...]

PR #874, opened 2022-01-26
Multiple CRD versions support

andreaTP:
This is an initial POC for fixing #94 

Opening the PR to discuss how/if we want to proceed in this direction and the pros/cons of this approach.

You can test the example by installing locally `kubernetes-client` coming from this open PR:
https://github.com/fabric8io/kubernetes-client/pull/3786

And test with this target example project:
https://github.com/andreaTP/josdk-multiple-crd-versions-example


csviri:
will be replaced by this, as we discussed: https://github.com/java-operator-sdk/java-operator-sdk/pull/879

PR #876, opened 2022-01-27
refactor: clean up configurations

metacosm:
- refactor: use DefaultResourceConfiguration for controller configuration
- feat: encapsulate Kubernetes dependent configuration

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResourceController.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResourceController.java
@@ -13,14 +13,13 @@ import io.javaoperatorsdk.operator.processing.event.source.informer.InformerEven
 
 public class KubernetesDependentResourceController<R extends HasMetadata, P extends HasMetadata>
     extends DependentResourceController<R, P> {
-  private final InformerConfiguration<R, P> configuration;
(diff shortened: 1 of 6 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResourceController.java
```diff
@@ -13,14 +13,13 @@
 
 public class KubernetesDependentResourceController<R extends HasMetadata, P extends HasMetadata>
```
csviri, line 14: TBH, don't see why we need this controller per resource model. Why cannot just the dependent resource manager handle this? Like having it  aggregate the controllers and the configuration. But this is probably something to discuss if we decide we got his way.

csviri (approved):
LGTM within this scope

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResourceController.java
metacosm: Because it needs to be passed around at the moment. This change is about putting together the informer part and the ownership configuration into a single coherent configuration. Cleans up things by showing what belongs together. Helps things with the Quarkus extension as well. 
We might indeed revisit, clean things further later but this change makes sense within the current scope.

PR #883, opened 2022-01-28
fix: decouple DependentResource creation from configuration

metacosm:
- fix: do not call default resource class resolution logic by default
- fix: do not create DependentResources directly from configuration

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
@@ -1,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+public interface DependentResourceConfiguration<R, P extends HasMetadata> {
+
+  Class<? extends DependentResource<R, P>> getDependentResourceClass();
+
+  Class<R> getResourceClass();
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
@@ -48,7 +48,7 @@ public interface ControllerConfiguration<R extends HasMetadata> extends Resource
     return ResourceEventFilters.passthrough();
   }
 
-  default List<DependentResource> getDependentResources() {
+  default List<DependentResourceConfiguration> getDependentResources() {
     return Collections.emptyList();
   }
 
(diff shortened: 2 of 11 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
```diff
@@ -0,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+public interface DependentResourceConfiguration<R, P extends HasMetadata> {
+
+  Class<? extends DependentResource<R, P>> getDependentResourceClass();
+
+  Class<R> getResourceClass();
```
csviri, line 9: This should not be `getPrimaryResourceClass()' ? or? it's not clear for me.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
```diff
@@ -0,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+public interface DependentResourceConfiguration<R, P extends HasMetadata> {
```
csviri, line 5: Why are these classes configurable? or what is the point of this configuration?

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
metacosm: No, that's actually not the primary resource class but the dependent resource one. The naming is tricky and should probably improved indeed: `getDependentResourceClass` returns the `DependentResource` implementation class (e.g. `DeploymentDependentResource`), while `getResourceClass` returns the actual associated resource class (e.g. `Deployment`).
Note that I've been trying to be consistent in the parameter types naming for these new classes: `R` represents a resource class, while `P` represents the primary resource types when it's needed. If we wanted to be more explicit we could use more explicit parameter types (e.g. use `Primary` instead of simply `P`) but that's not a common pattern in Java code…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResourceConfiguration.java
metacosm: The goal is to decouple the configuration of the dependent resource from its creation. Previously, this was done by the same class and the `AnnotationConfiguration` was doing the instantiating of the `DependentResource` implementation which led to it being too complex but also needed to know some details about how the implementation worked (in particular how to deal with Kubernetes dependent resources).
This also forced the quarkus extension to have to instantiate the dependent resource implementation class at build time because it was right there on the `ControllerConfiguration.getDependentResources()` method. Well, it's not exactly true because it's possible to delay the instantiation until the method is called but that wasn't clean enough and led to more complications.

PR #891, opened 2022-01-31
Improvements to the junit extension

andreaTP:
This PR integrates a few desired features into the current junit extension, without breaking the backward compatibility.


sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorNewE2E.java
```diff
@@ -0,0 +1,111 @@
...
+  // and run the tests as usual
+
+  @RegisterExtension
+  OperatorExtension operator =
+      OperatorExtension.builder()
+          .withConfigurationService(DefaultConfigurationService.instance())
+          .withReconciler(new MySQLSchemaReconciler(client,
+              new MySQLDbConfig("127.0.0.1", "3306", "root", "password")))
+          .withDeployment(
+              client.load(new FileInputStream("k8s/operator.yaml")).get())
```
andreaTP, line 45: cool enough, the `yaml` files provided in the `k8s` folder are already good enough to support this workflow seamlessly!

sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorNewE2E.java
```diff
@@ -0,0 +1,111 @@
...
+import io.javaoperatorsdk.operator.junit.OperatorExtension;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class MySQLSchemaOperatorNewE2E {
```
andreaTP, line 27: This test now runs locally:
```bash
mvn clean compile test -f sample-operators/mysql-schema/pom.xml -Dtest=MySQLSchemaOperatorNewE2E
```

and remotely:
```bash
eval $(minikube -p minikube docker-env)
mvn clean compile jib:dockerBuild test -f sample-operators/mysql-schema/pom.xml -Dtest=MySQLSchemaOperatorNewE2E -Djosdk.it.localOperatorDeployment=false
```

andreaTP:
Those changes are largely inspired by @jonathanvila on this PR: https://github.com/keycloak/keycloak/pull/9625/files

andreaTP:
Added a pipeline of the "mysql" sample to showcase the usage and benefits.
I do believe that those changes are reducing the complexity of the user's code and setup while they are not dramatically increasing the complexity in this codebase.

operator-framework-core/pom.xml
```diff
@@ -56,6 +56,7 @@
     <dependency>
       <groupId>io.fabric8</groupId>
       <artifactId>openshift-client</artifactId>
+      <version>${fabric8-client.version}</version>
```
andreaTP, line 59: I don't know why, Maven is complaining about those missing versions ... any hint on how to solve the issue without this change? cc. @csviri

andreaTP:
Happy to have a review of those proposed changes @csviri @metacosm

sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorNewE2E.java
metacosm: I have some ideas on maybe improving on this… but I need to find some time to work on it :/

operator-framework-core/pom.xml
```diff
@@ -98,6 +99,7 @@
     <dependency>
       <groupId>io.fabric8</groupId>
       <artifactId>kubernetes-server-mock</artifactId>
+      <version>${fabric8-client.version}</version>
```
metacosm, line 102: This one shouldn't be needed…

operator-framework-core/pom.xml
metacosm: Hmm, actually, it is so not sure why it's complaining now when it wasn't before… :(

sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorNewE2E.java
andreaTP: @metacosm this is going to work smoothly with the Quarkus SDK resources generated in the `target/kubernetes` folder.

> I have some ideas on maybe improving on this

do you wanna share?

operator-framework-core/pom.xml
andreaTP: could it be related to a local setup? I think I don't have anything fancy in place ...

operator-framework-core/pom.xml
andreaTP: fixed

operator-framework-core/pom.xml
andreaTP: Fixed, for the records, I had cached a "bad version" of the `kubernetes-client` BOM.

andreaTP:
NDR: rebased on latest `main` and Sonar check is correctly skipped @csviri

sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorNewE2E.java
andreaTP: An annotation processor would make it ... but here we are in the test, I think that performance is not critical ...

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -118,6 +139,10 @@ public Reconciler getFirstReconciler() {
         .findFirst().orElseThrow();
   }
 
+  public boolean isLocal() {
+    return this.localOperatorDeployment;
```
lburgazzoli, line 143: wonder if this can be implicit, if you have a operator deployment, then I guess it is local

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -47,24 +50,42 @@
...
+      boolean localOperatorDeployment) {
 
     this.kubernetesClient = new DefaultKubernetesClient();
     this.configurationService = configurationService;
     this.reconcilers = reconcilers;
+    this.operatorDeployment = operatorDeployment;
+    this.deploymentTimeout = deploymentTimeout;
+    this.infrastructure = infrastructure;
+    this.infrastructureTimeout = infrastructureTimeout;
     this.operator = new Operator(this.kubernetesClient, this.configurationService);
```
lburgazzoli, line 84: maybe this should be passed in the constructor and created by the builder I don't think there is a case where you need both a local operator and a deployed operator

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
```diff
@@ -275,12 +404,38 @@ public Builder withReconciler(Class<? extends Reconciler> value) {
       return this;
     }
 
+    public Builder withInfrastructure(List<HasMetadata> hm) {
```
csviri, line 407: Would not be better with  `withInfrastructure(HasMetadata ... hm)` ?

csviri:
Conceptually we are mixing here the Integration and E2E tests. In general not opposed to this. But not sure if these should not be two different extensions, in Integration tests we don't usually use other resources (e.g. infrastructure) we don't deploy of course the operator etc. Maybe having a hierarchy between these two extension? like a common base class, if we need to share code.
So in short these are little different domain, with different needs, this way we provide a mixed api for the two domains. Separating it to two extensions would make it more explicit.

But don't have strong opinion just an idea.

lburgazzoli:
> Conceptually we are mixing here the Integration and E2E tests. In general not opposed to this. But not sure if these should not be two different extensions, in Integration tests we don't usually use other resources (e.g. infrastructure) we don't deploy of course the operator etc. Maybe having a hierarchy between these two extension? like a common base class, if we need to share code. So in short these are little different domain, with different needs, this way we provide a mixed api for the two domains. Separating it to two extensions would make it more explicit.
> 
> But don't have strong opinion just an idea.

I don't have a strong opinion too but I think this is a valid point. 

Maybe (but it can be done in a follow up PR) we should end up with two different extensions: one that deals with an embedded operator and one with a deployed one.

andreaTP:
@csviri yes, indeed, we are intentionally mixing integration and E2E tests, the objective is to enable this workflow:

 - locally develop the operator and the tests running from your machine (integration) - fast iteration - missing checks on k8s (such as RBAC etc.)
 - in CI using the very same test/code and a simple switch from command line (E2E) - possibly slower - very close if not identical to the final product
 
 I'm not opposed to split the implementation using two subclasses but I would really like to make it easy to "switch" integration/E2E keeping the same(or almost) test sources.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/OperatorExtension.java
andreaTP: yup, done 👍

andreaTP:
Separated into a common abstract class to refactor common functionalities and two sub-classes `OperatorExtension` and `E2EOperatorExtension`.
This actually makes everything more clean, great suggestions thanks!

I think this is ready for another round of review 🙏  @lburgazzoli @csviri

csviri (approved):
LGTM, I think we can remove the old tests and the related workflow and merge it

andreaTP:
@csviri thanks! 👍 
I'll clean up the sample operators tests and remove the old "kind" approach in favour of the `E2EOperatorExtension` setup, anything else that should be cleaned up?

lburgazzoli (approved):

andreaTP:
- cleaned up
 - ported the E2E of the Tomcat operator to this new structure
 - unified the E2E workflows that are now using a GH action `matrix`

Should be good for a final check @csviri 🙂

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          minikube version: v1.24.0
+          kubernetes version: v1.23.0
+          github token: ${{ secrets.GITHUB_TOKEN }}
+          driver: docker
+
+      - name: Set up Java and Maven
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
```
metacosm, line 34: Should we target 17 instead?

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,50 @@
...
+          distribution: adopt-hotspot
```
metacosm, line 35: Is using a different distribution from the tests on purpose here? Might actually be a good idea since it improves somewhat VM coverage?

.github/workflows/e2e-test.yml
andreaTP: 👍  done

.github/workflows/e2e-test.yml
andreaTP: `adopt` rebranded the distribution to `temurin`, so adopt v17 doesn't exists (it's `temurin` now).

Switched to `temurin`, I "think" that, in this repository, we are not leveraging sensible parts of the JVM and we should not need to test against different distributions ... but open to suggestions!

metacosm (approved):
LGTM, will open a subsequent PR to improve things a little

PR #898, opened 2022-02-01
test(mysql-schema): fix sql query in userExists

scrocquesel:
actually, the method always return true


metacosm (approved):

csviri:
thx @scrocquesel !!

PR #899, opened 2022-02-01
fix: NPE when no namespace is provided in kube config

### LINKED ISSUE #897: NPE when initializing the operator if there's no explicit namespace set in Kubernetes config

metacosm:
- fix: fail explicitly if current namespace is requested but not available
- fix: retain annotation for reflective access

Fixes #897

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
@@ -4,6 +4,7 @@ import java.util.Collections;
 import java.util.Set;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.ReconcilerUtils;
 import io.javaoperatorsdk.operator.api.reconciler.Constants;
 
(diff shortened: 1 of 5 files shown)


csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
```diff
@@ -67,7 +68,12 @@ static boolean currentNamespaceWatched(Set<String> namespaces) {
         throw new IllegalStateException(
             "Parent ConfigurationService must be set before calling this method");
       }
-      targetNamespaces = Collections.singleton(parent.getClientConfiguration().getNamespace());
+      String namespace = parent.getClientConfiguration().getNamespace();
```
andreaTP, line 71: I'm not sure about this change, is possibly going to break user's (especially CIs) environment when a provider doesn't set the default namespace.

I would suggest that if the namespace is `null` you can output a warning message and automatically set it to `default`.

csviri: Btw not sure if this can happen in a real environment (I mean pod), probably just locally, for that purpose might be better to have an exception, so developer won't search for the problem unnecessarily.

andreaTP: A possibility, for the deployment, is to use a [downwardAPI](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api).

csviri: I mean, a default namespace will be always there.

metacosm: Yes, this should only happen when outside of the cluster…

andreaTP: > not sure if this can happen in a real environment

I'm mostly worried about CI ... but, well, an additional command doesn't hurt as well 🤷

metacosm: Let's see how it goes and maybe revisit if needed?

andreaTP (approved):

PR #900, opened 2022-02-02
fix: fail explicitly if current namespace is requested but not available

### LINKED ISSUE #897: NPE when initializing the operator if there's no explicit namespace set in Kubernetes config

metacosm:
Fixes #897

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
@@ -1,7 +1,6 @@
 package io.javaoperatorsdk.operator.processing;
 
 import java.util.List;
-import java.util.Objects;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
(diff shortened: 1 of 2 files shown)


csviri (approved):
LGTM

csviri:
The sonar should not be skipped here.

cc @andreaTP

andreaTP:
@csviri you are right, I messed up the key, it "should" be:

```
github.event.pull_request.head.repo.owner.login
```
instead of:
```
github.event.pull_request.repository_owner
```

here: https://github.com/java-operator-sdk/java-operator-sdk/blob/14403fb1908773afe439835406970c2f1bcf5a0b/.github/workflows/sonar.yml#L18

Do you mind to check?

(ref: https://github.community/t/get-the-head-repository-name-in-prs/17669/2 )

csviri:
> Do you mind to check?

Sure, thx for the hint, will fix it.

csviri (approved):

metacosm:
Well, it turns out that this "feature" might actually break the Quarkus extension… 😓

metacosm:
Would be nice to set up some sort of CI that would check the quarkus extension for each PR we create it, similarly to what the Quarkus CI does…

andreaTP:
NDR: the Quarkus CI uses the "star" ⭐  action as an hook for running external GH actions

metacosm:
> Well, it turns out that this "feature" might actually break the Quarkus extension… 😓

Actually, it's not breaking so all good. Setting up CI should still be done.

PR #906, opened 2022-02-03
redesign

metacosm:
- fix: fail explicitly if current namespace is requested but not available
- fix: retain annotation for reflective access
- feat: add desired and matches methods on DependentResource


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
```diff
@@ -12,4 +13,46 @@ default EventSource initEventSource(EventSourceContext<P> context) {
   default Class<R> resourceType() {
     return (Class<R>) Utils.getFirstTypeArgumentFromInterface(getClass());
   }
+
+  /**
+   * todo: return Optional<R></R> instead?
```
csviri, line 18: Yes, I think should be optional, we should never use `null` if it's avoidable

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
```diff
@@ -12,4 +13,46 @@ default EventSource initEventSource(EventSourceContext<P> context) {
...
+   * different.
+   * 
+   * @param actual the current state of the resource as fetched from the cluster
+   * @param primary the primary resource associated with the reconciliation request
+   * @param context the {@link Context} associated with the reconciliation request
+   * @return {@code true} if the actual state of the resource matches the desired state expressed by
+   *         the specified primary resource, {@code false} otherwise
+   */
+  default boolean match(R actual, P primary, Context context) {
+    return true;
```
csviri, line 56: shouldn't we use `equals` here?

csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResourceController.java
```diff
@@ -8,11 +8,8 @@
...
 
 public class DependentResourceController<R, P extends HasMetadata, C extends DependentResourceConfiguration<R, P>>
-    implements DependentResource<R, P>, Builder<R, P>, Updater<R, P>, Persister<R, P>,
-    Cleaner<R, P> {
+    implements DependentResource<R, P>, Persister<R, P>, Cleaner<R, P> {
 
-  private final Builder<R, P> builder;
-  private final Updater<R, P> updater;
   private final Cleaner<R, P> cleaner;
   private final Persister<R, P> persister;
```
csviri, line 14: do we need the persister?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
```diff
@@ -67,21 +67,15 @@ public void injectInto(EventSourceContext context) {
...
-            dependentResource = dependent.buildFor(resource, context);
-            createOrReplaceDependent(resource, context, dependent, dependentResource, "Creating");
-          } else if (dependent.updatable()) {
-            dependentResource = dependent.update(dependentResource, resource, context);
-            createOrReplaceDependent(resource, context, dependent, dependentResource, "Updating");
-          } else {
-            logOperationInfo(resource, dependent, dependentResource, "Ignoring");
-          }
-        });
+    dependents.stream().forEach(dependent -> {
```
csviri, line 70: I would put this logic the the reconcile method of the resource, and just call here reconcile

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
```diff
@@ -12,4 +13,46 @@ default EventSource initEventSource(EventSourceContext<P> context) {
...
+   * associated dependent should never be created by the associated reconciler or that the global
+   * state of the cluster doesn't allow for the resource to be created at this point.
+   * 
+   * @param primary the primary resource associated with the reconciliation process
+   * @param context the {@link Context} associated with the reconciliation process
+   * @return an instance of the dependent resource matching the desired state specified by the
+   *         primary resource or {@code null} if the dependent shouldn't be created at this point
+   *         (or ever)
+   */
+  default R desired(P primary, Context context) {
```
csviri, line 33: I think we will need here an extended context, with the other resources included. If for exmample a resource is depending on an other resource output?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
metacosm: No, see the javadoc 😉

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResourceController.java
metacosm: Yes for non-Kubernetes resources, see the MySQL sample

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
metacosm: Probably makes sense, yes

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
metacosm: That's what the context is for. This gives you accès to other resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
csviri:
```
for typical Kubernetes resource implementations, simply calling
   * {@code desired(primary, context).equals(actual)} is not enough because metadata will usually be
   * different.
```
This is true, but in a default implementation I would just compare specs. And in reconcile, would merge the target resource with the actual from the server. Thus just override the spec, add other fields if new to metadata (and override if some is generated). Would also make this behavior customizable (already is, some can override), to make a custom comaparator.

```
  * The default implementation always return {@code true}, which corresponds to the behavior where
   * the dependent never needs to be updated after it's been created.   
```
Why should be this the default?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
csviri: Was thinking a more explicit access, but for now this might be enough.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResourceController.java
csviri: Not sure if it make sense for kubernetes objects; maybe we should have this just implicit in the reconcile function. Or just have it for a subclass for non-kubernetes objects and get called from reconcile in that impl.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
metacosm: Done.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
metacosm: Done.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResourceController.java
metacosm: We could probably remove the Persister interface but I think that would require people defining Kubernetes dependents to either extend a class that would provide the behavior as it exists right now or do more work to implement the methods that are currently in Persister.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
metacosm: > This is true, but in a default implementation I would just compare specs. And in reconcile, would merge the target resource with the actual from the server. Thus just override the spec, add other fields if new to metadata (and override if some is generated). Would also make this behavior customizable (already is, some can override), to make a custom comaparator.
 
That's a good point

> Why should be this the default?

Why shouldn't it? 😄 
More seriously what do you think the default behavior should be? Not updating the resource after it's been created sounds like a reasonable default which covers a lot of use cases (for example, pretty much all our samples), imo.

csviri: I think it should not have a default implementation. 

The most common case is like in the WebPage, when all the resources are managed by the operator, IMHO that should be the default approach. If the resource is not managed by us (so not created and not updated) should be probably a feature flag/mode on the resources. Maybe eventually a different type of resource.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
```diff
@@ -0,0 +1,62 @@
...
+   * state of the cluster doesn't allow for the resource to be created at this point.
+   * 
+   * @param primary the primary resource associated with the reconciliation process
+   * @param context the {@link Context} associated with the reconciliation process
+   * @return an instance of the dependent resource matching the desired state specified by the
+   *         primary resource or {@code empty} if the dependent shouldn't be created at this point
+   *         (or ever)
+   */
+  default Optional<R> desired(P primary, Context context) {
+    return Optional.empty();
```
csviri, line 38: I think it should not have a default implementation. 

The most common case is like in the WebPage, when all the resources are managed by the operator, IMHO that should be the default approach. If the resource is not managed by us (so not created and not updated) should be probably a feature flag/mode on the resources. Maybe eventually a different type of resource.

csviri (approved):

csviri:
Made one more comment, I think we can merge it. Made a comment, but probably the best appraoch would be to fine tune it in a subsequent PRs.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/DependentResource.java
csviri: The default should be to generate the target default with desired, and compare the specs. 

The reason is that I think the default use case should be what we have in WebPage (what actally needs some cleanup), but basically that the reconciler manages all the resources. That is not that strong assumption IMHO, sure there are cases where not, but this would be quite sensible default.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: I'm not sure I follow. If we're not creating, nor updating the resource, why would it even be a dependent?

csviri: To get input from it, like not sure from nodes or other configurations of the cluster. But something like this happens also in the tomcat example too if remember correctly.

PR #908, opened 2022-02-04
Faster CI feedback loop

andreaTP:
- separated check format and unit tests from the matrix build with kubernetes/Java versions
- excluded all but the first and last kubernetes version for Java 11 (keeping Java 17 for coverage)

As a "Free" GH Actions account we have 20 concurrent jobs and we are already pretty much flooding them up.

I do believe that this setup is not compromising test coverage.

This proposed setup is biased toward Java 17, open to swap it with Java 11 on request.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -11,20 +11,15 @@ on:
     branches: [ main, v1 ]
   workflow_dispatch:
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


csviri (approved):
LGTM. Thx!

metacosm (approved):

PR #914, opened 2022-02-07
Dependent resources standalone mode

csviri:
Idea for standalon mode, it's a skeleton only, base for discussions.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResource.java
@@ -36,7 +36,7 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
     this.client = client;
   }
 
-  protected void postProcessDesired(R desired, P primary) {
+  protected void beforeCreateOrUpdate(R desired, P primary) {
     if (owned) {
       desired.addOwnerReference(primary);
     }
@@ -52,7 +52,7 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
   protected R create(R target, P primary, Context context) {
     log.debug("Creating target resource with type: " +
         "{}, with id: {}", target.getClass(), ResourceID.fromResource(target));
-    postProcessDesired(target, primary);
+    beforeCreateOrUpdate(target, primary);
     Class<R> targetClass = (Class<R>) target.getClass();
     return client.resources(targetClass).inNamespace(target.getMetadata().getNamespace())
         .create(target);
@@ -63,7 +63,7 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
   protected R update(R actual, R target, P primary, Context context) {
     log.debug("Updating target resource with type: {}, with id: {}", target.getClass(),
         ResourceID.fromResource(target));
-    postProcessDesired(target, primary);
+    beforeCreateOrUpdate(target, primary);
     Class<R> targetClass = (Class<R>) target.getClass();
     return client.resources(targetClass).inNamespace(target.getMetadata().getNamespace())
         .replace(target);


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,33 @@
...
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+
+public abstract class AbstractDependentResource<R, P extends HasMetadata>
+    implements DependentResource<R, P> {
+
+  @Override
+  public void reconcile(P primary, Context context) {
+    var actual = getResource(primary);
+    var desired = desired(primary, context);
+    if (actual.isEmpty()) {
```
metacosm, line 13: How do you specify that a dependent should never be created with this? It's a pretty common use case that you only want to update a dependent but not create it and I don't see how you can express this with this implementation.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,33 @@
...
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+
+public abstract class AbstractDependentResource<R, P extends HasMetadata>
+    implements DependentResource<R, P> {
+
+  @Override
+  public void reconcile(P primary, Context context) {
+    var actual = getResource(primary);
+    var desired = desired(primary, context);
```
metacosm, line 12: We should avoid creating the desired version if it's not needed because it could be costly.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,33 @@
...
+      if (!match(actual.get(), desired, context)) {
+        update(actual.get(), desired, context);
+      }
+    }
+  }
+
+  protected abstract R desired(P primary, Context context);
+
+  protected boolean match(R actual, R target, Context context) {
+    return actual.equals(target);
```
metacosm, line 25: This default implementation will result in infinite loops when working with Kubernetes resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
```diff
@@ -46,45 +43,16 @@ public KubernetesDependentResourceController(DependentResource<R, P> delegate,
...
+  public Optional<R> getResource(P primaryResource) {
+    return delegate.getResource(primaryResource);
   }
 
-  public boolean owned() {
-    return getConfiguration().isOwned();
-  }
-
-  @Override
-  protected void createOrReplaceDependent(P primary, R dependent, Context context) {
```
metacosm, line 84: This feature needs to be restored for automatic owner reference management. This is something that people have requested.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
```diff
@@ -46,45 +43,16 @@ public KubernetesDependentResourceController(DependentResource<R, P> delegate,
...
-    informer = new InformerEventSource<>(configuration, context);
-    return informer;
-  }
-
   @Override
-  public void createOrReplace(R dependentResource, Context context) {
-    client.resource(dependentResource).createOrReplace();
+  public Optional<EventSource> eventSource(EventSourceContext<P> context) {
+    var informer = new InformerEventSource<>(configuration, context);
+    ((KubernetesDependentResource) delegate).withInformerEventSource(informer);
```
metacosm, line 49: Maybe we should parameterize the super class with the dependent resource class so that the delegate would always be of the proper type and avoid having to cast? It would also prevent people from creating a `KubernetesDependentResourceController` and passing it a non Kubernetes `DependentResource`…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
```diff
@@ -46,45 +43,16 @@ public KubernetesDependentResourceController(DependentResource<R, P> delegate,
...
-  public EventSource initEventSource(EventSourceContext<P> context) {
-    this.client = context.getClient();
-    informer = new InformerEventSource<>(configuration, context);
-    return informer;
-  }
-
   @Override
-  public void createOrReplace(R dependentResource, Context context) {
-    client.resource(dependentResource).createOrReplace();
+  public Optional<EventSource> eventSource(EventSourceContext<P> context) {
```
metacosm, line 47: Why are we returning an optional? In what situation(s) would one not want to create an event source?

metacosm (changes requested):

scrocquesel:
A common use case is to provide an external identifier of the target, being a name/namespace for k8s. How will it behave if the spec change the name. Will it undo desired on the old one ans apply desired to new one ?

csviri:
> A common use case is to provide an external identifier of the target, being a name/namespace for k8s. How will it behave if the spec change the name. Will it undo desired on the old one ans apply desired to new one ?

The default `KubernetesDependentResource` is currently just applies if not changed. But of the aspects is that this whole logic can be overriden. So you can extend `KubernetesDependentResource` and override the `reconcile` method. Or just implement reconciliation from scratch.

Could you maybe create a separate issue for your case, I'm not sure if understand, and if such naming is that common. Cannot be solved with labels  more elegantly?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: That would be probably a feature flag, we can add to this default implementation, and be part of the logic. But we can also extend this or the KubernetesDepenendtResource and override the reconcile method. This is how is exactly meant, like the reconcile logic could be adjusted to special cases.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: As far I can see it is always used on both branches.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: This is changed since. Please pull.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
csviri: Will add this logic to the `KubernetesDependentResource` , but good point.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
metacosm: In this implementation, yes, that wasn't the case before :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
metacosm: A common use case should be supported by default without people having to do extra things to support it. I'd argue that the default should indeed be the case where we don't want to create a dependent, just update it or query its information. Implementing something is needed when creating resources so it's normal that it requires input. I think that the previous implementation was cleaner on that part.

metacosm:
> A common use case is to provide an external identifier of the target, being a name/namespace for k8s. How will it behave if the spec change the name. Will it undo desired on the old one ans apply desired to new one ?

I'm not sure I understand the use case here… Could you please detail it some more?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: Actually, the good part is that we can model that separatelly with other subclass maybe of the DependentResource interface. So it's more explicit. Will take a look and see how it looks.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
csviri: For example, when the controller is just periodically checking some external resource(s). I think we should have this possibility to have it without event source, although most of the cases it will have.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
metacosm: I would really prefer we avoid forcing people to learn about a class hierarchy to use dependent resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
metacosm: But you would still need an event source in that case, though… so I guess I don't understand how that's supposed to work.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
```diff
@@ -1,6 +1,11 @@
 package io.javaoperatorsdk.operator.api.reconciler;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 public final class Constants {
+  // Shared object mapper across app
+  public static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
metacosm, line 7: This is wrong. It shouldn't be shared across the app and always be retrieved from the `ConfigurationService`. The reason for this is that users might want to provide their own mapper and, in the case of the Quarkus extension, actually do.

metacosm (changes requested):

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
```diff
@@ -0,0 +1,78 @@
...
+  public StandaloneDependentTestReconciler() {
+    configMapDependent = new StandaloneKubernetesDependentResource<>() {
+      @Override
+      protected boolean match(Deployment actual, Deployment target, Context context) {
+        return Objects.equals(actual.getSpec().getReplicas(), target.getSpec().getReplicas()) &&
+            actual.getSpec().getTemplate().getSpec().getContainers().get(0).getImage()
+                .equals(target.getSpec().getTemplate().getSpec().getContainers().get(0).getImage());
+      }
+    };
+    configMapDependent.setResourceType(Deployment.class);
```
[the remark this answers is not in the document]
csviri: This can go to constructor yes here.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
```diff
@@ -98,4 +105,40 @@ public static String getDefaultReconcilerName(String reconcilerClassName) {
...
+  public static Object getSpec(HasMetadata resource) {
+    try {
+      Method getSpecMethod = resource.getClass().getMethod("getSpec");
+      return getSpecMethod.invoke(resource);
+    } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  public static void addOwnerReference(HasMetadata resource, HasMetadata owner) {
```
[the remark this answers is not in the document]
csviri: Could you pls elaborate, not sure how do you mean it

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
```diff
@@ -98,4 +105,40 @@ public static String getDefaultReconcilerName(String reconcilerClassName) {
...
+      return OBJECT_MAPPER.readTree(c1json).equals(OBJECT_MAPPER.readTree(c2json));
+    } catch (JsonProcessingException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  // will be replaced with: https://github.com/fabric8io/kubernetes-client/issues/3816
+  public static Object getSpec(HasMetadata resource) {
+    try {
+      Method getSpecMethod = resource.getClass().getMethod("getSpec");
```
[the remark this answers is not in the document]
csviri: Yes. This was the quick and dirty, created the related issue, if fabric8, but until then not sure how to solve this simple and elegantly. we can discuss the best strategy.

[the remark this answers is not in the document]
csviri: Maybe we should create a separate issue for these comparison.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
csviri: I think I already moved it from a static context, but can take a look.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
metacosm: Still not sure how that would work…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: We can take a look together, not sure how you mean it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
csviri: So basically imagine that there will be just an operator that reconciles every 5 mins, all the resources. All external, so no caches locally. This can pretty much work without an event source.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
csviri: Added a coment will revisit this.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
csviri: fixed

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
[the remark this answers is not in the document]
csviri: done.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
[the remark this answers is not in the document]
csviri: fixed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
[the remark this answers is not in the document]
csviri: created separate issues: https://github.com/java-operator-sdk/java-operator-sdk/issues/920

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: So currently there is just `KubernetesDependentResource` and an abstract above with generic logic. And one for standalone.  

Will take a look if the Kubernetes and the StandaloneOne could be unified. Probably yes.

scrocquesel:
> > A common use case is to provide an external identifier of the target, being a name/namespace for k8s. How will it behave if the spec change the name. Will it undo desired on the old one ans apply desired to new one ?
> 
> I'm not sure I understand the use case here… Could you please detail it some more?

@csviri @metacosm You can look at https://github.com/redhat-developer/service-binding-operator and find some example here https://redhat-developer.github.io/service-binding-operator/userguide/creating-service-bindings/creating-service-binding.html

```
apiVersion: binding.operators.coreos.com/v1alpha1
kind: ServiceBinding
metadata:
  name: account-service
spec:
  application:
    version: v1
    group: apps
    kind: Deployment
    name: online-banking
  services: 
  - version: v1alpha1 
    group: example.com 
    kind: AccountService 
    name: prod-account-service
```

Basically, you're telling the operator to update the Deployment named `online-banking`. But, the user may change the `application.name` value (and even `application.kind`). In such a case, I expect the operator to undo (like if the CR was deleted), and then apply to the new Deployment.

This use case happens a lot with operator that patch other resources. You must tell the target to patch.

csviri:
> > > A common use case is to provide an external identifier of the target, being a name/namespace for k8s. How will it behave if the spec change the name. Will it undo desired on the old one ans apply desired to new one ?
> > 
> > 
> > I'm not sure I understand the use case here… Could you please detail it some more?
> 
> @csviri @metacosm You can look at https://github.com/redhat-developer/service-binding-operator and find some example here https://redhat-developer.github.io/service-binding-operator/userguide/creating-service-bindings/creating-service-binding.html
> 
> ```
> apiVersion: binding.operators.coreos.com/v1alpha1
> kind: ServiceBinding
> metadata:
>   name: account-service
> spec:
>   application:
>     version: v1
>     group: apps
>     kind: Deployment
>     name: online-banking
>   services: 
>   - version: v1alpha1 
>     group: example.com 
>     kind: AccountService 
>     name: prod-account-service
> ```
> 
> Basically, you're telling the operator to update the Deployment named `online-banking`. But, the user may change the `application.name` value (and even `application.kind`). In such a case, I expect the operator to undo (like if the CR was deleted), and then apply to the new Deployment.
> 
> This use case happens a lot with operator that patch other resources. You must tell the target to patch.

In this case the old target is referenced in the status? Or we are just aware of the new spec?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/KubernetesDependentResourceController.java
metacosm: Yes but that's not the most common case and in the mean time, the most common cases is polluted by the `Optional` handling… Let's see if we can address this in a subsequent PR.

metacosm:
> 
> Basically, you're telling the operator to update the Deployment named `online-banking`. But, the user may change the `application.name` value (and even `application.kind`). In such a case, I expect the operator to undo (like if the CR was deleted), and then apply to the new Deployment.
> 
> This use case happens a lot with operator that patch other resources. You must tell the target to patch.

I would argue that this use case is not covered by the dependent resource support as it is as the indirection level makes it operator-specific to reason about what needs to be done in that case. I'm not sure how the Service Binding operator does when you modify the parameters of the initial binding… Note that having easier dependent resource support doesn't mean that you cannot write operators without it anymore! 😉

metacosm (approved):
There are some issues that need to be resolved but let's get this merged and address them in subsequent PRs. I've added more entries to #923 to reflect this fact.

csviri:
> > Basically, you're telling the operator to update the Deployment named `online-banking`. But, the user may change the `application.name` value (and even `application.kind`). In such a case, I expect the operator to undo (like if the CR was deleted), and then apply to the new Deployment.
> > This use case happens a lot with operator that patch other resources. You must tell the target to patch.
> 
> I would argue that this use case is not covered by the dependent resource support as it is as the indirection level makes it operator-specific to reason about what needs to be done in that case. I'm not sure how the Service Binding operator does when you modify the parameters of the initial binding… Note that having easier dependent resource support doesn't mean that you cannot write operators without it anymore! wink

This could be covered by a custom DependentResource implementation, where the `reconcile()` implements this custom logic. I don't see issues with that. If this would be requested from lot's of user we can also proveide an implementation out of the box.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  public void reconcile(P primary, Context context) {
+    var actual = getResource(primary);
+    var desired = desired(primary, context);
+    if (actual.isEmpty()) {
+      create(desired, primary, context);
+    } else {
+      if (!match(actual.get(), desired, context)) {
+        update(actual.get(), desired, primary, context);
+      }
```
scrocquesel, line 18: if match doesn't assert that it has been reconciled. The secondary R may exist prior to the primary one.  We should be able to put in the context what has been observed to be desired, so that the reconciler can update the status if it is empty.

In the case of mysql schema operator, if the database exists when the primary is created on the cluster, the status is never updated. But, it tricky here because the user is not symetric due to password.

if we split, schema and user in two different dependent. Thus, User::match cannot be sure actual = desired because the password may not be the desired one. So if the user exist prior to the CR, either it should fail the reconcile or the password should be changed and secret updated/recreated.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,31 @@
+package io.javaoperatorsdk.operator.api.reconciler.dependent;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+
+public abstract class AbstractDependentResource<R, P extends HasMetadata>
```
scrocquesel, line 6: If it stays as a Factory (i.e. not exist = create)  it should be renamed. And _sharing_ version could be added (i.e. not exist = update status with missing target condition). The _sharing_ version is another beast because, delete is about undoing what reconcile does not deleting. Again thinks of https://github.com/redhat-developer/service-binding-operator. When the CR is removed, the deployment is updated to suppress any env var that has been injected.
I don't if those two behaviors can be merge in one class

scrocquesel (commented):
Sorry to be late to review, I had a pretty overload week at work and not much spare time.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -0,0 +1,31 @@
...
+    implements DependentResource<R, P> {
+
+  @Override
+  public void reconcile(P primary, Context context) {
+    var actual = getResource(primary);
```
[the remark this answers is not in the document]
csviri: Actually, it's other way around, to for now the event sources are caching the resource ( this is an interesting thing we might want to re-evaluate this later, to separate these two ). But basically at the end the dependent resource should manage it itself, either having an event source (or directly calling the target API). So IMHO it should not have the context in getResource() basically forcing this pattern. (See `KubernetesDependentResource`)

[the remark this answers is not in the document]
csviri: On level above on DependentResource we don't have desired, so it's really about reading the resource and the desired here is just a way to generically implement it. Hmm will think about it, so yes, it's always actual, not sure if needs to be that explicit.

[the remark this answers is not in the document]
csviri: @metacosm  what do you think?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/StandaloneKubernetesDependentResource.java
```diff
@@ -0,0 +1,69 @@
...
+  public StandaloneKubernetesDependentResource(
+      KubernetesClient client, Class<R> resourceType, DesiredSupplier<R, P> desiredSupplier) {
+    super(client);
+    this.desiredSupplier = desiredSupplier;
+    this.resourceType = resourceType;
+  }
+
+  @Override
+  protected R desired(P primary, Context context) {
+    if (desiredSupplier != null) {
```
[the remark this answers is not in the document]
csviri: Good point, removed!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResource.java
```diff
@@ -0,0 +1,146 @@
...
+
+  public KubernetesDependentResource() {
+    this(null);
+  }
+
+  public KubernetesDependentResource(KubernetesClient client) {
+    this.client = client;
+  }
+
+  protected void postProcessDesired(R desired, P primary) {
```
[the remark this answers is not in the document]
csviri: Hmm not sure about this naming, it's really about filling some common data for the desired object created. But yes that maybe is better will change it. thx!

It's needed for now, there is a plan to (discuss first) to preserve labels and annotations and mainly finalizers (will actually do soon) because otherwise the update would remove this data from the resource. 

Eventually this might improve with the followup compare issue. Ideally we would do server apply just fabric8 client does not support it yet.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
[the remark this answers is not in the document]
metacosm: `actual` could make sense if we had `desired` at the same level, indeed. Maybe something that implies that the method is responsible for retrieving the resource more strongly? The previous implementation used `getFor` for this purpose but it's not really better. How about `fetchFrom(primary)` which would convey that the responsibility is to fetch the secondary based on the information provided by the primary?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/KubernetesDependentResource.java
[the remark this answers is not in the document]
metacosm: Yes, I'm not too happy with that part either as this feels kind of awkward since to me this should be handle by the controller rather (i.e. this shouldn't be something that implementers of dependent resources should worry about). The controller should take these automated steps based on the dependent configuration (in this case, whether the dependent is supposed to be owned or not) and deal with owner references automatically.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: > if match doesn't assert that it has been reconciled. The secondary R may exist prior to the primary one. We should be able to put in the context what has been observed to be desired, so that the reconciler can update the status if it is empty.

Not sure I'm following this. We read the R, and compare it to a target state, this is how P gets reconciled (at least this is a part of it). 

In MySQL there is no update now available for the password. But the whole mysql we will revisit shortly.

PR #919, opened 2022-02-09
fix: run e2e tests on next

csviri:


metacosm:
Weird I'm sure I had fixed this already. Maybe that got lost when `main` was merged back into `next`…

PR #921, opened 2022-02-10
Add groupId in Maven query

Sgitario:
When querying Maven central using the artifactId `operator-framework`, there were multiple matches:
- One using the groupId: `com.github.containersolutions`
- Another one using: `io.javaoperatorsdk`

While is obvious that the right one is the one using the groupId `io.javaoperatorsdk`, we can avoid confusions by querying Maven central using both the artifactId and the group Id fields.


csviri (approved):
LGTM

PR #926, opened 2022-02-11
Configuration polish

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/dependent/KubernetesDependent.java
```diff
@@ -12,15 +12,13 @@
 public @interface KubernetesDependent {
 
   boolean OWNED_DEFAULT = true;
-  boolean SKIP_UPDATE_DEFAULT = true;
```
csviri, line 15: We might want to put this back, it was then used in informer as recognized later, but still don't understand why was it added in first place.

metacosm:
Shouldn't have you merged this one before the other one?

PR #928, opened 2022-02-13
Experimental matchers

csviri:
WIP! 
Experimental Matchers to compare the actual spec with the spec of the desired object


csviri:
will open a separate PR just with resulted matcher.

PR #932, opened 2022-02-14
Dynamic config

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -115,12 +114,13 @@ public void delete(P primary, Context context) {
     resource.ifPresent(r -> client.resource(r).delete());
   }
 
+  protected abstract Class<R> resourceType();
```
metacosm: This makes things harder for the Quarkus extension because it will now need to check more possible cases.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/kubernetes/KubernetesDependentResourceInitializer.java
```diff
@@ -0,0 +1,68 @@
...
+import static io.javaoperatorsdk.operator.api.config.Utils.valueOrDefault;
+
+public class KubernetesDependentResourceInitializer<T extends DependentResource<?, ?>>
+    implements DependentResourceInitializer<T> {
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public T initialize(
+      Class<T> dependentResourceClass,
+      ControllerConfiguration<?> configurationService,
```
metacosm: This shouldn't be named `configurationService` as it is confusing.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/kubernetes/KubernetesDependentResourceInitializer.java
csviri: will fix

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageReconciler.java
```diff
@@ -118,6 +118,11 @@ protected boolean match(Deployment actual, Deployment target, Context context) {
             // todo comparator
             return true;
           }
+
+          @Override
+          protected Class<Deployment> resourceType() {
```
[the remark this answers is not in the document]
csviri: We had issues with that, at least getting it from generics, but can try it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/kubernetes/KubernetesDependentResource.java
csviri: What exactly?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
```diff
@@ -75,30 +91,55 @@ public DeleteControl cleanup(P resource, Context context) {
...
-
   private void initContextIfNeeded(P resource, Context context) {
     if (reconciler instanceof ContextInitializer) {
       final var initializer = (ContextInitializer<P>) reconciler;
       initializer.initContext(resource, context);
     }
   }
 
-  private DependentResource from(DependentResourceConfiguration config, KubernetesClient client) {
+  private DependentResourceInitializer getOrInitInitializerForClass(
```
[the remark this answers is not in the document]
metacosm: The purpose of the configuration objects was also to make things overridable by the Quarkus extension.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResourceInitializer.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  @SuppressWarnings("unchecked")
+  public T initialize(
+      Class<T> dependentResourceClass,
+      ControllerConfiguration<?> controllerConfiguration,
+      KubernetesClient client) {
+    try {
+      KubernetesDependentResource<?, ?> kubernetesDependentResource =
+          (KubernetesDependentResource<?, ?>) dependentResourceClass.getConstructor().newInstance();
+      kubernetesDependentResource.setKubernetesClient(client);
```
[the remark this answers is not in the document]
csviri: will do

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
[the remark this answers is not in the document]
csviri: What part exactly?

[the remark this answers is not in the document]
csviri: > This cannot be done like this. All this logic should be performed at build time using the Quarkus extension and this makes it impossible to do.

Which part exactly?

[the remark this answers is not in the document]
csviri: > The purpose of the configuration objects was also to make things overridable by the Quarkus extension.

Yes, realized it little later. But was quite confusing, will take a look on that and try to come up with something.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
```diff
@@ -44,8 +42,8 @@ public StandaloneDependentTestReconciler() {
 
   @Override
   public void setKubernetesClient(KubernetesClient kubernetesClient) {
+    configMapDependent = new DeploymentDependentResource(kubernetesClient);
```
metacosm, line 45: This looks/feels weird…

csviri: well, the way around is put back the setter :)

metacosm (commented):
Can you show an example of how `DependentResourceConfigService` are supposed to be used? Also, to me, this feels like making things more complex just to avoid having a configuration object for dependent resources. The approach of tying a dependent resource with a configuration makes more sense to me. This mirrors what we have for the reconciler and fits the overall design (dependent resources are basically mini-reconcilers so they might as well behave similarly).

csviri:
> Can you show an example of how `DependentResourceConfigService` are supposed to be used? Also, to me, this feels like making things more complex just to avoid having a configuration object for dependent resources.


It's used in `KubernetesDependentResourceInitializer`.

https://github.com/java-operator-sdk/java-operator-sdk/blob/dynamic-config/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResourceInitializer.java

> The approach of tying a dependent resource with a configuration makes more sense to me. This mirrors what we have for the reconciler and fits the overall design (dependent resources are basically mini-reconcilers so they might as well behave similarly.

The point is that if somebody for example create wants to create just a standalone dependent resource should not care that there is a whole configuration loading and and service layer mechanism in the background. So basically the idea is to have the dependent resource as simple as possible. 

It not completely same as in controller and/or reconciler, since these typec can be added and configured dynamically by the user, so custom annotations made and stuff. So it has much more dynamic nautire. In the former implementation there was hardcoded how to parse the KubernetesDependentResource, this is what I wanted to avoid. So basically have this dynamically extensible.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
metacosm: To me, that's a smell in the design…

csviri: Where exactly? in dependent resource side, or in the test?

metacosm:
> > Can you show an example of how `DependentResourceConfigService` are supposed to be used? Also, to me, this feels like making things more complex just to avoid having a configuration object for dependent resources.
> 
> It's used in `KubernetesDependentResourceInitializer`.
> 
> https://github.com/java-operator-sdk/java-operator-sdk/blob/dynamic-config/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResourceInitializer.java

This doesn't show how to use it from a user's perspective, though.

> 
> > The approach of tying a dependent resource with a configuration makes more sense to me. This mirrors what we have for the reconciler and fits the overall design (dependent resources are basically mini-reconcilers so they might as well behave similarly.
> 
> The point is that if somebody for example create wants to create just a standalone dependent resource should not care that there is a whole configuration loading and and service layer mechanism in the background. So basically the idea is to have the dependent resource as simple as possible. Not forcing anything on it, like a mandatory config loading implementations etc.

Event sources need to be configured: it's just a fact, not an extra layer. Right now, this implementation is hardcoding that configuration in subclasses instead of having a general mechanism to associate an event source and its configuration within a dependent resource implementation. The configuration is there, it's just less obvious and forces people to learn how each dependent resource implementation works instead of having a more general mechanism to deal with dependent sources in a similar way, regardless of the implementation.

> 
> It not completely same as in controller and/or reconciler, since these types/dependentResources can be added and configured dynamically by the user, so custom annotations made and stuff. So it has much more dynamic nautire. In the former implementation there was hardcoded how to parse the KubernetesDependentResource, this is what I wanted to avoid. So basically have this dynamically extensible.

I don't get that point. Currently, this implementation hardcodes how to deal with annotations with even more classes in a class that shouldn't even have to know how a dependent resource is configured: it should just match a configuration with the dependent resource implementation and that's it…
https://github.com/java-operator-sdk/java-operator-sdk/pull/932/files#diff-b67488930905dbf2637bb895b299fbd00b354777b49bd1467f896df272b80e74R107-R129

Annotations shouldn't be processed in the core of the SDK.

> This is very similar at the end just separating the initialization from the service layer that loads the config from depenedent resource.

Except that it's not because the concern are mixed: the `DependentResourceManager` class shouldn't have to deal with extracting configuration, the configuration needs to be processed before

csviri:
> This doesn't show how to use it from a user's perspective, though.

True, will show it on MySQL sample. Commit coming soon. 

> Event sources need to be configured: it's just a fact, not an extra layer. Right now, this implementation is hardcoding that configuration in subclasses instead of having a general mechanism to associate an event source and its configuration within a dependent resource implementation. The configuration is there, it's just less obvious and forces people to learn how each dependent resource implementation works instead of having a more general mechanism to deal with dependent sources in a similar way, regardless of the implementation.

The problem with that general mechanism, that there is no common intersection, in the other words for KubernetesDependentResource and other won't share any part of the configuration, so the user will have to learn it anyways.
In the current we are leaking these values into config, non of them is needed in general, that makes it very confusing:

```
Class<? extends DependentResource<R, P, ? extends DependentResourceConfiguration<R, P>>> getDependentResourceClass();

  Class<R> getResourceClass();
```

> I don't get that point. Currently, this implementation hardcodes how to deal with annotations with even more classes in a class that shouldn't even have to know how a dependent resource is configured: it should just match a configuration with the dependent resource implementation and that's it…
https://github.com/java-operator-sdk/java-operator-sdk/pull/932/files#diff-b67488930905dbf2637bb895b299fbd00b354777b49bd1467f896df272b80e74R107-R129

yes, but you define the initalizer. If basically we just add like this an adaptor layer for the config. Other parts are the same. But without initializer, we need the setter for the `kubernetesClient` right? we cannot call the constructor, just the config method.
So the initialization layer makes is very generic. Maybe it's worth a call to discuss this, hard to explain it without pointing multiple points in the code.


> Annotations shouldn't be processed in the core of the SDK.

This is not true if you refer to the custom annotations like `KubernetesDependent`.
It's true, for the `AnnotationControllerConfiguration` that should be moved to the core, will do. Not sure why is that separated in first place.


> Except that it's not because the concern are mixed: the DependentResourceManager class shouldn't have to deal with extracting configuration, the configuration needs to be processed before. 

It's not dealing with it, it just passes it to the initalizer as before.

csviri:
Just to clarify something. The initializer serves two purposes:
- Decuples the config services from the DependentResource (I think it's nicer this way, reasons above)
- Makes the whole mechanism extendable. Like now the KunbernetesDependent resource is hard coded, also the parsing is hardcoded, this should not be a case, why should be this special? To allow annotations like this in a extensible way we need the initializer layer (eiher as it is now or maybe moving it to the config layer).

csviri:
Another thing with the current implementation on `configuration-polish` is that it creates the whole InformerConfiguration. While as we discussed we don't want to set the `AssociatedSecondaryResourceIdentifier` and  `PrimaryResourcesRetriever`, this should be done using the defauls or implementing interfaces in DependentResources. So this is an other leaking config there.

metacosm:
> 
> > Annotations shouldn't be processed in the core of the SDK.
> 
> This is not true if you refer to the custom annotations like `KubernetesDependent`. It's true, for the `AnnotationControllerConfiguration` that should be moved to the core, will do. Not sure why is that separated in first place.

`AnnotationConfiguration` is a configuration implementation based on annotations. It's not in the core on purpose to keep it lighter for runtimes that don't need that processing. Also, in the current implementation `KubernetesDependent` is not processed at runtime if you're not using `AnnotationConfiguration`. 

> > Except that it's not because the concern are mixed: the DependentResourceManager class shouldn't have to deal with extracting configuration, the configuration needs to be processed before.
> 
> It's not dealing with it, it just passes as before just now to initalizer not to the dependent resource.

It checks whether you have defined an initializer which is part of the configuration of the dependent resource. This shouldn't be the responsibility of the `DependentResourceManager` class.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/SchemDependentResourceInitializer.java
```diff
@@ -0,0 +1,40 @@
...
+    this.configService = configService;
+  }
+
+  @Override
+  public SchemaDependentResource initialize(
+      Class<SchemaDependentResource> resourceClass,
+      ControllerConfiguration<?> controllerConfiguration,
+      KubernetesClient client) {
+    try {
+      int pollPeriod = resourceClass.getAnnotation(ResourcePoller.class).pollPeriod();
```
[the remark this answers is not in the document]
csviri: Oki, replaced with this: https://github.com/java-operator-sdk/java-operator-sdk/pull/937
will close this part for now, we can check on later to support custom annotation and/or initializers.

PR #935, opened 2022-02-15
Log the detailed exception when reconciliation

zuston:
When implementing the controller like `TomcatReconciler` and throw exception when accepting k8s event, the log will not throw the exception stacktrace. Due to this limitation, I spent lot of time to dig operator framework and found the fundamental failure.

So i hope this PR to solve this problem.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
@@ -2,6 +2,9 @@ package io.javaoperatorsdk.operator.processing;
 
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.api.model.apiextensions.v1.CustomResourceDefinition;
@@ -30,6 +33,8 @@ import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 @SuppressWarnings({"unchecked"})
 public class Controller<R extends HasMetadata> implements Reconciler<R>,
     LifecycleAware, EventSourceInitializer<R> {
+  private static final Logger log = LoggerFactory.getLogger(Controller.class);
+
   private final Reconciler<R> reconciler;
   private final ControllerConfiguration<R> configuration;
   private final KubernetesClient kubernetesClient;
@@ -98,7 +103,12 @@ public class Controller<R extends HasMetadata> implements Reconciler<R>,
 
           @Override
           public UpdateControl<R> execute() {
-            return reconciler.reconcile(resource, context);
+            try {
+              return reconciler.reconcile(resource, context);
+            } catch (Exception e) {
+              log.error("Errors on reconciliation.", e);
+              throw e;
+            }
           }
         });
   }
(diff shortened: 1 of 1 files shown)


zuston:
@csviri @metacosm Please review it.

csviri (approved):
LGTM

csviri (changes requested):
actually this would be better to catch in `ControllerExecution`. @zuston could you change it pls ? or I can issue a new PR

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -98,7 +103,12 @@ public String successTypeName(UpdateControl<R> result) {
 
           @Override
           public UpdateControl<R> execute() {
-            return reconciler.reconcile(resource, context);
+            try {
+              return reconciler.reconcile(resource, context);
+            } catch (Exception e) {
+              log.error("Errors on reconciliation.", e);
```
metacosm, line 109:
```suggestion
              log.error("Reconciliation error for resource: " + ResourceID.fromResource(resource)
                  + " / reconciler: " + controllerName(), e);
```

metacosm (approved):
LGTM

csviri:
Or rather here:

```
  public PostExecutionControl<R> handleExecution(ExecutionScope<R> executionScope) {
    try {
      return handleDispatch(executionScope);
    } catch (KubernetesClientException e) {
      log.info(
          "Kubernetes exception {} {} during event processing, {} failed",
          e.getCode(),
          e.getMessage(),
          executionScope);
      return PostExecutionControl.exceptionDuringExecution(e);
    } catch (RuntimeException e) {
      log.error("Error during event processing {} failed.", executionScope, e);
      return PostExecutionControl.exceptionDuringExecution(e);
    }
  }
```
Wondering why this does not work.

metacosm:
> actually this would be better to catch in `ControllerExecution`. @zuston could you change it pls ? or I can issue a new PR

I was wondering about that, indeed, and thought that we could do that in a subsequent PR.

csviri:
Probably just this should be an error level:
```
   } catch (KubernetesClientException e) {
      log.info(
          "Kubernetes exception {} {} during event processing, {} failed",
          e.getCode(),
          e.getMessage(),
          executionScope);
      return PostExecutionControl.exceptionDuringExecution(e);
```

zuston:
Thanks for you quick reply @csviri @metacosm .
Maybe this also should be added in `cleanup` and other methods exposed for users.

csviri:
> Thanks for you quick reply @csviri @metacosm . Maybe this also should be added in `cleanup` and other methods exposed for users.

The thing is, for me it seems that the only problem is we make different logging for client exception, as you can see we catch all the exceptions also for cleanup in the quoted code. So I would just remove that special case for the client.

csviri:
created this PR: https://github.com/java-operator-sdk/java-operator-sdk/pull/936
would rather do it this way and just close this.

zuston:
> created this PR: #936 would rather do it this way and just close this.
> 
> But thx @zuston to pointing this out!

I like this general fix. Thanks @csviri. 

Could you let me into slack group, now I use this project to develop our elastic-yarn operator.

csviri:
> Could you let me join into slack group, now I use this project to develop our elastic-yarn operator.

We don't have a slack group, there is a Discord server, use this to join: https://discord.gg/DacEhAy
Or join the generic Kuberentes slack, usually watching this channel: https://kubernetes.slack.com/archives/CAW0GV7A5

PR #947, opened 2022-02-16
fix: provide configuration via interface on reconciler

### LINKED ISSUE #940: Review how configuration is injected in DependentResource
It's currently quite awkward to inject configuration into a dependent resource. See https://github.com/java-operator-sdk/java-operator-sdk/pull/937#discussion_r807130890_

metacosm:
Fixes #940

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
@@ -11,6 +11,7 @@ import org.slf4j.LoggerFactory;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
+import io.javaoperatorsdk.operator.api.config.dependent.DependentResourceConfigurationProvider;
 import io.javaoperatorsdk.operator.api.config.dependent.DependentResourceSpec;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.ContextInitializer;
@@ -49,8 +50,7 @@ public class DependentResourceManager<P extends HasMetadata>
         dependentResources.stream()
             .map(
                 drc -> {
-                  final var dependentResource =
-                      from(drc, context.getClient());
+                  final var dependentResource = from(drc, context.getClient());
                   dependentResource
                       .eventSource(context)
                       .ifPresent(es -> sources.add((EventSource) es));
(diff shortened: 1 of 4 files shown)


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
```diff
@@ -40,6 +42,13 @@
 
   public MySQLSchemaReconciler() {}
 
+  @Override
+  public Optional<Object> configurationFor(DependentResourceSpec<?, ?> dependent) {
```
csviri, line 46: We should just receive the config here, not the whole spec.

csviri: The class is basically this class in the spec

metacosm: I hesitated here but I thought that passing the entire spec was better, in particular because I think at some point we might have a name (or some sort of qualifier) on the spec to distinguish between dependents of the same kind… Or should that be implemented using different `DependentResource` implementations? 🤔

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
```diff
@@ -90,8 +90,14 @@ private DependentResource from(DependentResourceSpec dependentResourceSpec,
       if (dependentResource instanceof KubernetesClientAware) {
         ((KubernetesClientAware) dependentResource).setKubernetesClient(client);
       }
-      dependentResourceSpec.getDependentResourceConfiguration()
-          .ifPresent(dependentResource::configureWith);
+
+      var configuration = dependentResourceSpec.getDependentResourceConfiguration();
+      if (reconciler instanceof DependentResourceConfigurationProvider) {
+        final var provider = (DependentResourceConfigurationProvider) reconciler;
+        configuration = provider.configurationFor(dependentResourceSpec);
```
csviri, line 97: we should just pass the config part here not the whole spec.

metacosm: Not sure I understand… The whole idea is to provide the configuration so if we're passing it, why are we providing it? 😄

csviri: Ahh sorry I miss read it / did not check it properly. Hmm don't think the Reconciler should be a provider.

csviri: let's take a look on this togther.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
```diff
@@ -40,6 +42,13 @@
...
+    return SchemaDependentResource.class.equals(dependent.getDependentResourceClass())
+        ? Optional.of(new ResourcePollerConfig(500, MySQLDbConfig.loadFromEnvironmentVars()))
```
csviri, line 48: This could be just the default in the dependent resource `configureFor` method. Don't think we need this at all.

metacosm:
Not currently needed. Let's revisit this later if needed.

Issue #870, opened 2022-01-26
Filtering Event of Informers After Watched Resource Update


csviri:
## Problem Statement
In case we during the reconciliation update a dependent resource, we receive a new event regarding the change of the dependent resource from the informer. In general this even can be skipped since we know that the state of the target resource is already the desired one. The way to do this would be compare `resourceVersion` of the resource of the received event with the resource version of a cached resource (not the current informer cache). 

## Proposed Solution
One relatively transparent way would be to propagate the `resourceVersion` to the informer to cache it. When we receive the next event compare the resources versions, if are the same just skip the event. But either is the same or not, instantly remove the cached `resourceVersion` from the cache. This way we don't cache too long the version, and don't have to specially handle the cleanup on resource delete event.

Note that dependent resources could do the propagation automatically, while on low level API this needs to be done explicitly by the implementation.

scrocquesel:
Regarding concurrency issue would a semaphore mechnism allows to update the resource while blocking the evensource until the `resourceVersion` is propagated.  ? The drawback is that the informer will not be able to process new events, so maybe it can retain events concerning only the resource being edited.

This should have no impact because we are still in the reconcile loop and imho, we should return the same version of the dependent resource that raise the reconciliation. That is to say, two different calls in the same reconcile loop to `context.getSecondaryResource` shall return the same resource version.

csviri:
> Regarding concurrency issue would a semaphore mechnism allows to update the resource while blocking the evensource until the resourceVersion is propagated. ? The drawback is that the informer will not be able to process new events, so maybe it can retain events concerning only the resource being edited

This is not really feasible, because there might be other events, meanwhile, and the controller might fail or other errors might happen etc, that would need special care etc. Also the event can be lost in some corner cases.

I think the proposed solution is good enough, btw we already do something similar with the custom resource, that seems to be work fine.

> That is to say, two different calls in the same reconcile loop to context.getSecondaryResource shall return the same resource version.

Ideally yes, there should be kinda "snapshot isolation", but its not that big issue if doesn't imho, also it can be handled on code level, plus usually the resource only read once. I pretty sure that controller-runtime is not doing this either.

csviri:
The linked implementation does the job, but not always, especially locally the event is sometimes arrived before the temporal cached is updated. I guess this is the best we can do ATM.

csviri:
[...]

PR #953, opened 2022-02-17
Update resource cache after update/create from Dependent Resource

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -39,4 +40,8 @@ public void stop() {
     super.stop();
     manager().stop();
   }
+
+  public void populateCacheUpdatedResource(R resource) {
+    cache.put(ResourceID.fromResource(resource), resource);
```
metacosm, line 45: Based on https://github.com/fabric8io/kubernetes-client/pull/3654#issuecomment-1026838969, I'm not sure we should be modifying the informer's cache :(
Another thing is that this won't be possible in 6.0 anymore so we need to think about an alternative, in particular in the light of: https://github.com/fabric8io/kubernetes-client/pull/3616. So we might have to think about re-adding our own caching layer sooner than later…

csviri: Yes, the plan is to do it with the separate cache if the informers will be modified for that. Not sure if we can get away with this until that time. It's not 100% correct, but the chance to have there a race condition is very close to zero

csviri: An alternative is to wait for the event that the cache received the resource with the resource version.

csviri: That would mean not the update/delete would propagate the cache but the `getResource()` will be kinda "synchronous"

csviri: Solved it in a different way, pls see updated PR. (Will add unit tests soon)

metacosm:
I'm thinking that this work should be integrated in the `InformerResourceCache` layer instead of a separate layer that needs to be managed by, and therefore only useful to, `KubernetesDependentResource`. This would make the functionality available even for people not using `KubernetesDependentResource` and solve the problematic informer cache update that's currently performed by `InformerResourceCache` in a neat way. What do you think?

csviri:
The thing is that this is not needed when you work directly with resources, since when a resource is created by the client a new object is returned that users can work with further. So it us really useful for dependent resources. Since the getResource() is called eventually by other dependent resource. 

Plan is to put it there yes, after the cache layer will be separated from informers, as mentioned by @shawkins; for an upcoming version of informers. So for now I would just merge this, and create a followup issue.

Putting this updated resource now to cache of the informers has race condition issues, that are  not nicely solvable now.

csviri:
just a remark this issue I actually plan to put into informer event source: https://github.com/java-operator-sdk/java-operator-sdk/issues/870

It's kinda similar to this. So we can revisit also there. Not sure we need this to MVP though, maybe right after?

csviri:
Hmm, actually can take a look if you insist, now this impl looks quite different then on begging. Maybe would be able to implement the mentioned issue in a simple way.

csviri:
this PR is replaced by: https://github.com/java-operator-sdk/java-operator-sdk/pull/965

PR #954, opened 2022-02-17
feat: share InformerManager when configuration is equal

### LINKED ISSUE #942: Sharing informers between Managed Dependent Resources

metacosm:
Fixes #942

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
@@ -141,4 +141,8 @@ public class InformerManager<T extends HasMetadata, C extends ResourceConfigurat
     getSource(key.getNamespace().orElse(ANY_NAMESPACE_MAP_KEY))
         .ifPresent(c -> c.put(key, resource));
   }
+
+  int numberOfSources() {
+    return sources.size();
+  }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -1,5 +1,10 @@
 package io.javaoperatorsdk.operator.processing.event.source.informer;
 
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
@@ -13,15 +18,31 @@ public abstract class ManagedInformerEventSource<R extends HasMetadata, P extend
     extends CachingEventSource<R, P>
     implements ResourceEventHandler<R> {
 
+  @SuppressWarnings("rawtypes")
+  private static final ConcurrentMap<ResourceConfigurationAsKey, InformerManager> managedInformers =
+      new ConcurrentHashMap<>(17);
+
   protected ManagedInformerEventSource(
       MixedOperation<R, KubernetesResourceList<R>, Resource<R>> client, C configuration) {
     super(configuration.getResourceClass());
+    initCache(configuration);
     manager().initSources(client, configuration, this);
+  }
+
+  @SuppressWarnings("unchecked")
+  private void initCache(C configuration) {
+    final var key = new ResourceConfigurationAsKey(configuration);
+    var existing = managedInformers.get(key);
+    if (existing == null) {
+      existing = new InformerManager<>();
+      managedInformers.put(key, existing);
+    }
+    cache = existing;
   }
 
   @Override
   protected UpdatableCache<R> initCache() {
-    return new InformerManager<>();
+    return null; // cache needs the configuration to be properly initialized
   }
 
   protected InformerManager<R, C> manager() {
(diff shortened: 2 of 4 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
```diff
@@ -141,4 +141,8 @@ public void put(ResourceID key, T resource) {
     getSource(key.getNamespace().orElse(ANY_NAMESPACE_MAP_KEY))
         .ifPresent(c -> c.put(key, resource));
   }
+
+  int numberOfSources() {
+    return sources.size();
```
csviri, line 146: Should we just return an immutable map, this is not very useful outside of the tests.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -13,15 +18,31 @@
     extends CachingEventSource<R, P>
     implements ResourceEventHandler<R> {
 
+  @SuppressWarnings("rawtypes")
+  private static final ConcurrentMap<ResourceConfigurationAsKey, InformerManager> managedInformers =
```
csviri, line 22: TBH, don't know what to think about this static context. 
What is this magic number 17?

csviri: This will be started and stopped multiple times. Don't like the facts that it is even shared between the controllers. It will be started and stopped multiple times.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
metacosm: This is indeed just for tests, which is why it's package-protected.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
metacosm: Well, you're the one who wanted to share informers between sources… :)
We now avoid starting or stopping multiple times anyway.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -51,14 +51,18 @@ private void initCache(C configuration) {
...
-    super.start();
+    if(!isRunning()) {
+      manager().start();
+      super.start();
+    }
   }
 
   @Override
   public void stop() {
-    super.stop();
```
csviri, line 39: Not sure about the thread safety but probably ok this way

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
csviri: Pls add integration test for this, like when 2 reconcilers added with 2 Informers, so there is only one informer in the background.

csviri:
BTW this is not exactly the same as sharing the Informer event source, we are just sharing the informer, so the events are still propagated twice in case two informers are created?!

metacosm:
> BTW this is not exactly the same as sharing the Informer event source, we are just sharing the informer, so the events are still propagated twice in case two informers are created?!

That's a good point but I'm actually unsure which behavior we would want.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -13,15 +18,31 @@
...
 
+  @SuppressWarnings("rawtypes")
+  private static final ConcurrentMap<ResourceConfigurationAsKey, InformerManager> managedInformers =
+      new ConcurrentHashMap<>();
+
   protected ManagedInformerEventSource(
       MixedOperation<R, KubernetesResourceList<R>, Resource<R>> client, C configuration) {
     super(configuration.getResourceClass());
+    initCache(configuration);
     manager().initSources(client, configuration, this);
```
csviri, line 29: Won't this multiple time init sources?

csviri: Ehh we need a quite good quality intergation tests here IMHO. This is quite risky to do this way.

csviri:
> > BTW this is not exactly the same as sharing the Informer event source, we are just sharing the informer, so the events are still propagated twice in case two informers are created?!
> 
> That's a good point but I'm actually unsure which behavior we would want.

Well with sharing the event source it's one event per event source ( and one informer as here). So we are sharing here the informer not the event source. Not sure that is a good thing, especially quite hard to reason about, well maybe we need to get used to it. 

Basically we are sharing it because of the cache. From the point the event source is created the dependent resource only accesses the cache. There should be just one event that triggers the reconciliation from this point. If the primaryResourceMapper would be different for the same informer, that would be a different story.

metacosm:
I'm actually not sure I understand the use case for sharing an event source.

csviri:
Was thinking more of this, actually the idea is very good, just maybe we should come up with a design how exactly do this. Also not sure again on what layer, probably needed for informers but also for event source layer.

csviri:
@metacosm  can we close this PR?
probably part of a bugger topic that we should discuss / design better, see : https://github.com/java-operator-sdk/java-operator-sdk/issues/1134 (point 3)

Issue #887, opened 2022-01-30
Dependent Resource Standalone Building Blocks


csviri:
todo

PR #956, opened 2022-02-17
fix: make it possible to have read-only dependents

### LINKED ISSUE #955: Make it possible to have dependents that are never created or updated
It is very often the case that we want our operator to react to events pertaining to resources that are outside of the control scope of the operator (for example, because they need user input) and should therefore never be created, nor updated by our reconciler.

metacosm:
Fixes #955

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
@@ -24,7 +24,9 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata, C>
     });
   }
 
-  protected abstract Optional<R> desired(P primary, Context context);
+  protected Optional<R> desired(P primary, Context context) {
+    return Optional.empty();
+  }
 
   protected abstract boolean match(R actual, R target, Context context);
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
@@ -1,5 +1,7 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent;
 
+import java.util.Optional;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 
@@ -8,18 +10,21 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata, C>
 
   @Override
   public void reconcile(P primary, Context context) {
-    var actual = getResource(primary);
-    var desired = desired(primary, context);
-    if (actual.isEmpty()) {
-      create(desired, primary, context);
-    } else {
-      if (!match(actual.get(), desired, context)) {
-        update(actual.get(), desired, primary, context);
+    var maybeActual = getResource(primary);
+    var maybeDesired = desired(primary, context);
+    maybeDesired.ifPresent(desired -> {
+      if (maybeActual.isEmpty()) {
+        create(desired, primary, context);
+      } else {
+        final var actual = maybeActual.get();
+        if (!match(actual, desired, context)) {
+          update(actual, desired, primary, context);
+        }
       }
-    }
+    });
   }
 
-  protected abstract R desired(P primary, Context context);
+  protected abstract Optional<R> desired(P primary, Context context);
 
   protected abstract boolean match(R actual, R target, Context context);
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
@@ -10,9 +10,9 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata, C>
 
   @Override
   public void reconcile(P primary, Context context) {
-    var maybeActual = getResource(primary);
     var maybeDesired = desired(primary, context);
     maybeDesired.ifPresent(desired -> {
+      var maybeActual = getResource(primary);
       if (maybeActual.isEmpty()) {
         create(desired, primary, context);
       } else {
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/config/UtilsTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/config/UtilsTest.java
@@ -1,5 +1,7 @@
 package io.javaoperatorsdk.operator.api.config;
 
+import java.util.Optional;
+
 import org.junit.jupiter.api.Test;
 
 import io.fabric8.kubernetes.api.model.apps.Deployment;
@@ -23,8 +25,8 @@ class UtilsTest {
       extends KubernetesDependentResource<Deployment, TestCustomResource> {
 
     @Override
-    protected Deployment desired(TestCustomResource primary, Context context) {
-      return null;
+    protected Optional<Deployment> desired(TestCustomResource primary, Context context) {
+      return Optional.empty();
     }
   }
 }
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/standalonedependent/StandaloneDependentTestReconciler.java
@@ -1,6 +1,7 @@
 package io.javaoperatorsdk.operator.sample.standalonedependent;
 
 import java.util.List;
+import java.util.Optional;
 
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.client.KubernetesClient;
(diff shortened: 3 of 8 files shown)


csviri:
Isn't read only dependent just an informer? 
I would not change the AbstractDependentResource, would rather extend `DependentResource` and have `ReadOnlyDependetResource`.

csviri:
> > Isn't read only dependent just an informer?
> > I would not change the AbstractDependentResource, would rather extend `DependentResource` and have `ReadOnlyDependetResource` - that would have empty reconcile method. Basically just wrap an informer to getting a resource. That could be done quite in general way also for other informers.
> 
> I knew you would say that and I don't agree. smile The class hierarchy is already deep enough and there's no need to create another class when this can be solved by just changing the return type of the `desired` method.

This should be about the right abstractions. It would not be deeper, wider, this is why we have `DependentResource` this abstract in first place. So if we want to have read only dependent resources I think we should reflect that in our mode, like this is the domain language.  `AbstractReadOnlyDependetResource` would be really usefull for other and most of it can be implemented in general. Basically just the `getResource` needs to be implemented, in the abstract. So that would be very clean.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -8,18 +10,23 @@
...
-    var desired = desired(primary, context);
-    if (actual.isEmpty()) {
-      create(desired, primary, context);
-    } else {
-      if (!match(actual.get(), desired, context)) {
-        update(actual.get(), desired, primary, context);
+    var maybeActual = getResource(primary);
+    var maybeDesired = desired(primary, context);
+    maybeDesired.ifPresent(desired -> {
+      if (maybeActual.isEmpty()) {
```
scrocquesel, line 16:
```suggestion
    var maybeDesired = desired(primary, context);
    maybeDesired.ifPresent(desired -> {
      var maybeActual = getResource(primary);
      if (maybeActual.isEmpty()) {
```

scrocquesel: small perf optim

scrocquesel:
> > > Isn't read only dependent just an informer?
> > > I would not change the AbstractDependentResource, would rather extend `DependentResource` and have `ReadOnlyDependetResource` - that would have empty reconcile method. Basically just wrap an informer to getting a resource. That could be done quite in general way also for other informers.
> > 
> > 
> > I knew you would say that and I don't agree. smile The class hierarchy is already deep enough and there's no need to create another class when this can be solved by just changing the return type of the `desired` method.
> 
> This should be about the right abstractions. It would not be deeper, wider, this is why we have `DependentResource` this abstract in first place. So if we want to have read only dependent resources I think we should reflect that in our mode, like this is the domain language. `AbstractReadOnlyDependetResource` would be really usefull for other and most of it can be implemented in general. Basically just the `getResource` needs to be implemented, in the abstract. So that would be very clean.

If it looks like a dependent but don't behave like a dependent, then it is not a dependent. I think a definition of what is a dependent resource should be settled. To me `Dependent` is too generic. Does it mean a resource the primary resource reconciler :
- manage the lifecycle
- modify but do not own the lifecycle
- needs to read to behave on other dependents

From what I understand of the actual `DependentResource` interface, it has been narrowed to be only the first case. And I'm not convince, it will be a one fit all. Maybe better to be prepared to handle other kinds ?

Even if it is abstract away with noop default, it is weird to have a reconcile and delete method and resource that don't need it. To me, it is more like a LookupResource that can be added to dependents ?

```
@ControllerConfiguration(finalizerName = NO_FINALIZER,
    dependents = {
        @Lookup(type = MyLookupResource.class),
        @Dependent(type = MyDependentResource.class),
        @Edit(type = MyEditableResource.class
    })

interface LookupResource {
    getResource(T primary);
}

interface EditResource {
    getResource(T primary);

    apply(...);
    undo(...);

    ...
}

```

csviri:
I agree with this basically. This is also a matter of definition, what is a dependent resource, but this when needs to just read is not really a dependent resource, true. At least not feels like it. It could be defined as a super interface actually to the dependent resource, maybe just having `getResource()`. Basically that it is, what we need from it.

The one reason that on `DependetResource` there is a reconcile method, is to able to cover cases which are not exactly an "ideal" dependent resource. For example the editable feels more like this. So covering that as a abstraction as `EditedDependentResource` what extends  `DependentResource` is reasonable still IMHO.

So right modeling is essential here. @metacosm wants to limit the hierarchy and that is not a bad aspect of the problem. On the other hand we should capture these notions in the domain model. And for now it does not seems very extereme with the mentioned abstractions. 

I think first we should think in object models, then have an annotation structure defined, since that is limited. There are problem which are essentially harder to do in the background. I would like to have this api with annotations, ideally this whole structure is declarative, but that should be the second step. Like for example in standalone mode if we need a status of a dependent resource as an input of an other it really can just reference the dependent resource as an attribute in class. For annotations we need additional mechanisms in the background what gives more complexity.

metacosm:
The deeper the class hierarchy the more fragile the architecture becomes and the more complex it is to configure things via annotations. So yes, a read-only Kubernetes dependent is essentially an informer but this allows to very quickly create an informer using simply an annotation instead of figuring out which class to use among 5-6 of them and have to manually wire it.

Of note, the initial implementation used to split the different behaviors by interfaces where create, update, delete were traits that you could add to your dependent.

csviri:
Maybe it's just me, but for me is much more easy to understand a clean class hierarchy then check why a functionality is hacked to an abstraction what serves a completely different purpose. Like from now on if somebody wants to create a read only resource (for an external resource let's say) will have extend implement AbstractDependentResource, override `create` and `update` methods with empty body since those are abstract. For me this is very very confusing. 

In addition to that now every KubernetesDependentResource `desire` method override will return an Optional. And this is mandatory to override in every case. What is also seems to smelly. 

Also why it is harder to configure? it's anyway a new subclass, where just a config needs to be passed.

Can provide an alternate implementation to show how I mean if that is not clear. With this approach, how I meant it originally:

```mermaid

classDiagram    
    DependentResource <|-- AbstractDependentResource
    DependentResource <|-- EditableResource     
    DependentResource <|-- ReadOnlyDependentResource
    AbstractDependentResource <|-- KuberentesDependentResource


```

scrocquesel:
> Of note, the initial implementation used to split the different behaviors by interfaces where create, update, delete were traits that you could add to your dependent.

Isn't the reconciler v2 refactoring all about using trait interface ?

Class hierarchy are a pain when we tends to put things in the hierarchy that should not. But the same fragility will happens when building fat class. Here, Interface Segregation principle is violated and to me, while it seems to be a quick win now, it certainly become a pain latter.

> Maybe it's just me, but for me is much more easy to understand a clean class hierarchy then check why a functionality is hacked to an abstraction what serves a completely different purpose. Like from now on if somebody wants to create a read only resource (for an external resource let's say) will have extend implement AbstractDependentResource, override `create` and `update` methods with empty body since those are abstract. For me this is very very confusing.
> 
Confusing enough, `delete` on the parent interface has been forgotten too. 

> I agree that **this has it's negative sides** too. So maybe it's really make sense to compare prototypes. But the goal is also to have it easily extendable for external resources. As we have with the EventSources (Polling, PerResource polling). Then it's good to have these notions there explicitly.
> 
> Nevertheless this is an interesting topic, so thx to initiate and starting on it.

I like your last diagram. Even if for me DependentResource may be an empty marker interface to ease configuration.

Looking at DependentResourceManager, it is just a delegation pattern over a collection. Maybe complexity should be there. Manager is the glue between the `prepareEventSources`/`reconcile`/`cleanup` logic and the specific behavior of the resource itself. Maybe their should be a hierachy in there to handle, Generic, Editable, ReadOnly.

Roughly,
```
DependentResourceManager<Generic> {
    // as actual.
}

EditableResourceManager<EditableResource> {
    reconcile(P primary){
      dependent.getResource().ifPresent(editable -> dependent.apply(editable));
    }
    cleanup(P primary){
      dependent.getResource().ifPresent(editable -> dependent.undo(editable));
    }
}

ReadOnlyDependentResourceManager<ReadOnlyDependentResource> {
    reconcile(P primary){
    }
    cleanup(P primary){
    }
}
```

csviri:
Created a prototype with class hierarchy:
https://github.com/java-operator-sdk/java-operator-sdk/pull/958

It ended up quite ok. Especially when for a polling (readonly) dependent resource. But when created an abstract Kuberentes Resource, turned out also quite reasonable for the sub-classes, and it's quite clear.

The editable at the end added as a feature flag to the `KubernetesDependentResource` since it's just about to turning on/off one part of it. 

Pls take a look, what do you think about this way.

csviri:
> Isn't the reconciler v2 refactoring all about using trait interface ?

Not really, but there were changes regarding that, too. It was more about how we address resources in event sources, using informers for main event source (also for caching) etc.

> Looking at DependentResourceManager, it is just a delegation pattern over a collection. Maybe complexity should be there. Manager is the glue between the prepareEventSources/reconcile/cleanup logic and the specific behavior of the resource itself. Maybe their should be a hierachy in there to handle, Generic, Editable, ReadOnly.

One of the design aspects however is to able to use the DependentResources in standalone mode:
https://github.com/java-operator-sdk/java-operator-sdk/issues/887
Moving logic into separate managers contradicts with this as far I can see.

metacosm:
One aspect that this change addresses that is not otherwise addressed (so far) is the fact that it's now possible to express that a dependent doesn't yet need to be created or doesn't need to be updated, thus allowing unnecessary server calls. One could return `empty` if the resource doesn't need to be updated, for example. With the current implementation, it's not possible to do so and a call to the server is always made even if the resource hasn't changed.
So, regardless of other considerations, class hierarchy or not, I believe we ought to return `Optional` anyway.

Issue #957, opened 2022-02-17
Reconciler should know when dependent resource already exist


scrocquesel:
When the `AbstractDependentResource` already exist prior to creating the primary resource, the reconciler should know about it to be able to handle the situation and update the primary resource status accordingly.
For eg., the reconciler may set a CONFLICT status to let the user knows the observed spec conflict with an existing resource.

I have pushed a commit to demonstrate the use case on mysql-schema sample : https://github.com/scrocquesel/java-operator-sdk/tree/schema-already-exist

Be aware that, on startup, when `match` is true, the dependent resource alone cannot tell if there is a possible conflict or if it is what it should be from a prior reconciliation, actual and desired just match. Only the reconciler can tell apart if it saves the previous observed state.

For eg.

```java
return context.get(BUILT_SCHEMA, Schema.class).map(s -> {
      updateStatusPojo(schema, context.getMandatory(MYSQL_SECRET_NAME, String.class),
          context.getMandatory(MYSQL_SECRET_USERNAME, String.class));
      log.info("Schema {} created - updating CR status", schema.getMetadata().getName());
      return UpdateControl.updateStatus(schema);
    }).orElseGet(() -> context.get(EXISTING_SCHEMA, Schema.class).map(s -> {
      if (schema.getStatus().getStatus() != "CREATED")
        schema.getStatus().setStatus("CONFLICT");
      log.info("Schema {} already exists - updating CR status", schema.getMetadata().getName());
      return UpdateControl.updateStatus(schema);
    }).orElse(UpdateControl.noUpdate()));
```

scrocquesel:
Sorry, I forgot to mention that currently there is no way to set `(EXISTING_SCHEMA, Schema.class)` in the context.

csviri:
@scrocquesel is this issue still relevant? not sure if I follow, so you need to know if the resource existed on beginning of reconciliation. This kinda contradicts with the Kubernetes / Controller philosophy of level based event handling. That you should not work based on the history but just the current state.  What is the goal you want to achieve with this?

BTW we were discussion some more inf returned from `dependentResource.reconcile()`, to denote what exactly happened during the reconciliation. Like if the resource is created, updated or deleted during the reconciliation. Just want to understand you use case more.

csviri:
see also: https://github.com/java-operator-sdk/java-operator-sdk/pull/1030
(this is for standalone mode)

scrocquesel:
[...]

PR #958, opened 2022-02-18
Dependent resource hierarchy

csviri:
Featuring also ReadOnly, PollingReadOnly resources. Edit feature for Kubernetes

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/GenericDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/GenericDependentResource.java
@@ -3,11 +3,11 @@ package io.javaoperatorsdk.operator.api.reconciler.dependent;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 
-public abstract class AbstractDependentResource<R, P extends HasMetadata, C>
-    implements DependentResource<R, P> {
+public interface GenericDependentResource<R, P extends HasMetadata>
+    extends DependentResource<R, P> {
 
-  @Override
-  public void reconcile(P primary, Context context) {
+
+  default void reconcile(P primary, Context context) {
     var actual = getResource(primary);
     var desired = desired(primary, context);
     if (actual.isEmpty()) {
@@ -19,13 +19,13 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata, C>
     }
   }
 
-  protected abstract R desired(P primary, Context context);
+  R desired(P primary, Context context);
 
-  protected abstract boolean match(R actual, R target, Context context);
+  boolean match(R actual, R target, Context context);
 
-  protected abstract R create(R target, P primary, Context context);
+  void create(R target, P primary, Context context);
 
   // the actual needed to copy/preserve new labels or annotations
-  protected abstract R update(R actual, R target, P primary, Context context);
+  void update(R actual, R target, P primary, Context context);
 
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReadOnlyDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReadOnlyDependentResource.java
@@ -1,0 +1,15 @@
+package io.javaoperatorsdk.operator.api.reconciler.dependent;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.reconciler.Context;
+
+public interface ReadOnlyDependentResource<R, P extends HasMetadata>
+    extends DependentResource<R, P> {
+
+  @Override
+  default void reconcile(HasMetadata primary, Context context) {}
+
+  @Override
+  default void delete(HasMetadata primary, Context context) {}
+
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/PollingDependentResourceConfig.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/PollingDependentResourceConfig.java
@@ -1,0 +1,23 @@
+package io.javaoperatorsdk.operator.processing.dependent.external;
+
+public class PollingDependentResourceConfig {
+
+  public static final long DEFAULT_POLLING_PERIOD = 700;
+
+  private long pollingPeriod = DEFAULT_POLLING_PERIOD;
+
+  public PollingDependentResourceConfig() {}
+
+  public PollingDependentResourceConfig(long pollingPeriod) {
+    this.pollingPeriod = pollingPeriod;
+  }
+
+  public PollingDependentResourceConfig setPollingPeriod(long pollingPeriod) {
+    this.pollingPeriod = pollingPeriod;
+    return this;
+  }
+
+  public long getPollingPeriod() {
+    return pollingPeriod;
+  }
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/PollingReadOnlyDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/PollingReadOnlyDependentResource.java
@@ -1,0 +1,45 @@
+package io.javaoperatorsdk.operator.processing.dependent.external;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.config.Utils;
+import io.javaoperatorsdk.operator.api.reconciler.EventSourceContext;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResourceConfigurator;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.EventSourceProvider;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.ReadOnlyDependentResource;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.EventSource;
+import io.javaoperatorsdk.operator.processing.event.source.polling.PollingEventSource;
+
+import static io.javaoperatorsdk.operator.processing.dependent.external.PollingDependentResourceConfig.DEFAULT_POLLING_PERIOD;
+
+public abstract class PollingReadOnlyDependentResource<R, P extends HasMetadata, C extends PollingDependentResourceConfig>
+    implements ReadOnlyDependentResource<R, P>, EventSourceProvider<P>,
+    Supplier<Map<ResourceID, R>>, DependentResourceConfigurator<C> {
+
+  private PollingEventSource<R, P> pollingEventSource;
+  private long pollingPeriod = DEFAULT_POLLING_PERIOD;
+
+  @Override
+  public EventSource eventSource(EventSourceContext<P> context) {
+    pollingEventSource = new PollingEventSource<>(this, pollingPeriod, resourceType());
+    return null;
+  }
+
+  @Override
+  public Optional<R> getResource(P primaryResource) {
+    return pollingEventSource.getAssociated(primaryResource);
+  }
+
+  protected Class<R> resourceType() {
+    return (Class<R>) Utils.getFirstTypeArgumentFromExtendedClass(getClass());
+  }
+
+  @Override
+  public void configureWith(C config) {
+    this.pollingPeriod = config.getPollingPeriod();
+  }
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/InformerConfig.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/InformerConfig.java
@@ -1,0 +1,46 @@
+package io.javaoperatorsdk.operator.processing.dependent.kubernetes;
+
+import io.javaoperatorsdk.operator.api.config.ConfigurationService;
+
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.EMPTY_STRING;
+
+public class InformerConfig {
+
+  private String[] namespaces = new String[0];
+  private String labelSelector = EMPTY_STRING;
+  private ConfigurationService configurationService;
+
+  public InformerConfig() {}
+
+  public InformerConfig(String[] namespaces, String labelSelector,
+      ConfigurationService configurationService) {
+    this.namespaces = namespaces;
+    this.labelSelector = labelSelector;
+    this.configurationService = configurationService;
+  }
+
+  public void setNamespaces(String[] namespaces) {
+    this.namespaces = namespaces;
+  }
+
+  public void setLabelSelector(String labelSelector) {
+    this.labelSelector = labelSelector;
+  }
+
+  public void setConfigurationService(
+      ConfigurationService configurationService) {
+    this.configurationService = configurationService;
+  }
+
+  public String[] namespaces() {
+    return namespaces;
+  }
+
+  public String labelSelector() {
+    return labelSelector;
+  }
+
+  public ConfigurationService getConfigurationService() {
+    return configurationService;
+  }
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
@@ -12,8 +12,11 @@ import static io.javaoperatorsdk.operator.api.reconciler.Constants.EMPTY_STRING;
 public @interface KubernetesDependent {
 
   boolean ADD_OWNER_REFERENCE_DEFAULT = true;
+  boolean EDIT_ONLY_DEFAULT = false;
 
   boolean addOwnerReference() default ADD_OWNER_REFERENCE_DEFAULT;
+
+  boolean editOnly() default EDIT_ONLY_DEFAULT;
 
   /**
    * Specified which namespaces this Controller monitors for custom resources events. If no
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
@@ -7,13 +7,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.config.ConfigurationService;
 import io.javaoperatorsdk.operator.api.config.informer.InformerConfiguration;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.EventSourceContext;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.GenericDependentResource;
-import io.javaoperatorsdk.operator.api.reconciler.dependent.KubernetesClientAware;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 import io.javaoperatorsdk.operator.processing.event.source.AssociatedSecondaryResourceIdentifier;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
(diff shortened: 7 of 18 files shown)


csviri:
> This doesn't appear to address the read-only use-case for Kubernetes dependent resources.

There is the KubernetesReadOnlyDependentResource: 
https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-3d696bac55047d47272489b0c39f3cc786d2f5ae358cb230c0fe933d495eb2c9
Just with this approach it becomes a very simple layer.

> Also, if we're going to have configuration for this, I'd rather do it via interfaces to express traits than making it flags in the configuration. Using traits, we would only add the methods when needed (and to implement the support for create/update you would need these methods anyway) rather than do it with sub-classes (having trait interfaces would bypass the need for sub-classes since the base classes wouldn't need to provide abstract/default implementations).

The downside of this is that, it's not very explicit. So what is the behavior if a trait object is missing. Also not sure how do you mean it. The `DependentResource` would be still as the same as it is, and there will be a subclass that has these traits?

> Moreover, what happens if someone uses KubernetesReadOnlyResource and overrides reconcile? It's possible and, while unlikely, this could cause difficult to trace errors… The SDK should strive to help users avoid coding mistakes.

The point is that for different behaviors like the generic case or the read only behavior we already provide abstract implementations, if somebody overrides it should know what is doing. I guess this is generally problem with anything else, like overriding a trait object and braking the behavior.

> Another issue is that this implementation duplicates code in different branches of the hierarchy (notably getResource) making it more difficult to maintain. 

I don't understand this, where do you mean it? Could you put here links to the code pls.

metacosm:
> > This doesn't appear to address the read-only use-case for Kubernetes dependent resources.
> 
> There is the KubernetesReadOnlyDependentResource: https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-3d696bac55047d47272489b0c39f3cc786d2f5ae358cb230c0fe933d495eb2c9 Just with this approach it becomes a very simple thin layer.

It is not read-only though: the resource will still update (unless you override `match` to always return `true`, which more ugly than return `empty`, imo).

> 
> > Also, if we're going to have configuration for this, I'd rather do it via interfaces to express traits than making it flags in the configuration. Using traits, we would only add the methods when needed (and to implement the support for create/update you would need these methods anyway) rather than do it with sub-classes (having trait interfaces would bypass the need for sub-classes since the base classes wouldn't need to provide abstract/default implementations).
> 
> The downside of this is that, it's not very explicit. So what is the behavior if a trait object is missing. Also not sure how do you mean it. The `DependentResource` would be still as the same as it is, and there will be a subclass that has these traits? Also if we just have traits, the common behavior is shared, what could be limiting, see the case when the spec reference object name is changing, and other use cases we don't think of now. Having a very generic `reconcile()` method is basically a future proof. I'm fine to experiment using traits objects as a special implementation of the current `DependentResource` abstraction btw.

There will only be one default implementation that covers most cases. If people have a case that isn't covered, they could still create their own `DependentResource` implementation completely. The point of the dependent resources implementation as it is, is not to cover all possible use cases (because it's just not possible), the point is to provide an easy to use feature that covers 80% of the use cases. 

Having lots of implementations for special cases doesn't help because it makes things more complex to understand, to maintain and to use. And it still wouldn't cover all the use cases.

> 
> > Moreover, what happens if someone uses KubernetesReadOnlyResource and overrides reconcile? It's possible and, while unlikely, this could cause difficult to trace errors… The SDK should strive to help users avoid coding mistakes.
> 
> The point is that for different behaviors like the generic case or the read only behavior we already provide abstract implementations, if somebody overrides it should know what is doing. I guess this is generally problem with anything else, like overriding a trait object and braking the behavior.

There should only be one default implementation that covers 80% of the cases in a nice way. Not 10 different implementations that cover 90% in an awkward and error-prone way. :)

> 
> > Another issue is that this implementation duplicates code in different branches of the hierarchy (notably getResource) making it more difficult to maintain.
> 
> I don't understand this, where do you mean it? Could you put here links to the code pls.

https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-3d696bac55047d47272489b0c39f3cc786d2f5ae358cb230c0fe933d495eb2c9R13

https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-aa20588ab4b1ff4f171a897d4042d1b055c02737b067d09aaeb9cfd770adf3a0R136

csviri:
> It is not read-only though: the update method will still be called unless you override match to always return true, which more ugly than returning empty, imo. Granted, in most cases, it won't do anything but if you mistakenly happen to override the method, then your resource is not read-only anymore. So the framework is simply relying on naming to enforce something instead of enforcing it in the code.

Pls check it again, I think you misread it, it does not extend the KubernetesDependentResource, and has no access to any of the methods you mention.

csviri:
> There should only be one default implementation that covers 80% of the cases in a nice way. Not 10 different implementations that cover 90% in an awkward and error-prone way. :)

What use case we don't cover with this PR?

csviri:
> https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-3d696bac55047d47272489b0c39f3cc786d2f5ae358cb230c0fe933d495eb2c9R13
> 
> https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-aa20588ab4b1ff4f171a897d4042d1b055c02737b067d09aaeb9cfd770adf3a0R136

that had completely differnet scop, and we dsicussed that it's not the way to go as remember.

metacosm:
> > It is not read-only though: the update method will still be called unless you override match to always return true, which more ugly than returning empty, imo. Granted, in most cases, it won't do anything but if you mistakenly happen to override the method, then your resource is not read-only anymore. So the framework is simply relying on naming to enforce something instead of enforcing it in the code.
> 
> Pls check it again, I think you misread it, it does not extend the KubernetesDependentResource, and has no access to any of the methods you mention.

So I guess I can't override `reconcile` or `delete` in read-only resources? What makes them read-only apart from their name?

csviri:
> So I guess I can't override reconcile or delete in read-only resources? What makes them read-only apart from their name?

you don't have `delete` there if you take a look. If somebody calls reconcile it won't do anything, you can just call `getResource()`

csviri:
`The prototype was the implementation that was written 3 months ago… :)`

That was quite a different situation we did not have the `reconcile()` or the current `DependentResource` in there at all.
As I said, that is fine to have just, you did not answered the question, what case we don't cover now, with this simple class hierarchy.

metacosm:
> > https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-3d696bac55047d47272489b0c39f3cc786d2f5ae358cb230c0fe933d495eb2c9R13
> > https://github.com/java-operator-sdk/java-operator-sdk/pull/958/files#diff-aa20588ab4b1ff4f171a897d4042d1b055c02737b067d09aaeb9cfd770adf3a0R136
> 
> fixed this. Moved to base class. But needs different case for generic see, other PR.

Since the class is now empty, why do we even need it? It doesn't enforce anything, doesn't provide any new feature, it simply increases the cognitive load on the user that needs to figure out why they would use this implementation as opposed to another one.

> > So I guess I can't override reconcile or delete in read-only resources? What makes them read-only apart from their name?
> 
> you don't have `delete` there if you take a look. If somebody calls reconcile it won't do anything, you can just call `getResource()`

`ReadOnlyDependentResource` is a `DependentResource` so it provides `reconcile` and `delete`, which can be overridden. The only thing it does is that it provides a default empty implementation. Again, nothing enforces the read-only nature of implementations of this interface. If your implementation overrides `reconcile` (which, I admit, would be pretty weird), your implementation is not read-only anymore. The SDK doesn't enforce that read-only nature in the code, only in convention.

csviri:
> Since the class is now empty, why do we even need it? It doesn't enforce anything, doesn't provide any new feature, it simply increases the cognitive load on the user that needs to figure out why they would use this implementation as opposed to another one.

It implements the read only interface.  And that is not desired for the generic one. This is not that exotic in java.

> ReadOnlyDependentResource is a DependentResource so it provides reconcile and delete, which can be overridden. The only thing it does is that it provides a default empty implementation. Again, nothing enforces the read-only nature of implementations of this interface. If your implementation overrides reconcile (which, I admit, would be pretty weird), your implementation is not read-only anymore. The SDK doesn't enforce that read-only nature in the code, only in convention.
More precisely, how is relying on empty implementation of methods any better than enforcing a behavior via a return type? For the latter, you can actually enforce it in code.

You can override reocncile in any case, this is not an argument at all. If somebody want's to break it can. It's not about that, it's defining with an interface a common behavior abstraction.

The optional way woudl be pretty ugly on the other hand, since one would have to override delete and update and create. While those won't be used. Think about how something like `PollingReadOnlyDependentResource` would be implemented, would be super ugly to force it there.

metacosm:
> > Since the class is now empty, why do we even need it? It doesn't enforce anything, doesn't provide any new feature, it simply increases the cognitive load on the user that needs to figure out why they would use this implementation as opposed to another one.
> 
> It implements the read only interface. And that is not desired for the generic one.

Then just implement the interface. No need for an empty class that doesn't add anything useful.

> 
> > ReadOnlyDependentResource is a DependentResource so it provides reconcile and delete, which can be overridden. The only thing it does is that it provides a default empty implementation. Again, nothing enforces the read-only nature of implementations of this interface. If your implementation overrides reconcile (which, I admit, would be pretty weird), your implementation is not read-only anymore. The SDK doesn't enforce that read-only nature in the code, only in convention.
> > More precisely, how is relying on empty implementation of methods any better than enforcing a behavior via a return type? For the latter, you can actually enforce it in code.
> 
> You can override reocncile in any case, this is not an argument at all. If somebody want's to break it can. It's not about that, it's defining with an interface a common behavior abstraction.

But that behavior doesn't exist. It's only materialised by a specific implementation and a naming convention. There's no "read-only" concept that you can enforce in the code of the SDK with this implementation.

> 
> The optional way woudl be pretty ugly on the other hand, since one would have to override delete and update and create. While those won't be used. Think about how something like `PollingReadOnlyDependentResource` would be implemented, would be super ugly to force it there.

There's no `PollingDependentResource` at the moment either, so you already need to implement all these methods and I would argue that there shouldn't be one such implementation anyway. Returning `empty` instead of something would not make things more difficult to implement, imo. 

The value of the SDK should be to make the common use cases very easy to implement, define a pattern that can be extended to all implementations, provide support for that pattern in the code, instead of trying to exhaustively support every use case we can think of. Having a class hierarchy doesn't address the generality problem, it just keeps adding special cases that make the code base more confused, more difficult to maintain and harder to learn. It also forces extensions of the SDK to know about theses implementations instead of relying on a simpler, generic pattern thus making them more fragile with respect to changes in the implementation.

csviri:
> Then just implement the interface. No need for an empty class that doesn't add anything useful.

The problem with this, that the `KubernetesDependentResource` already would implement Generic and ReadOnly interface as the result. What is a bigger smell IMHO.

> But that behavior doesn't exist. It's only materialised by a specific implementation and a naming convention. There's no "read-only" concept that you can enforce in the code of the SDK with this implementation.

> There's no PollingDependentResource at the moment either, so you already need to implement all these methods and I would argue that there shouldn't be one such implementation anyway. Returning empty instead of something would not make things more difficult to implement, imo.

> The value of the SDK should be to make the common use cases very easy to implement, define a pattern that can be extended to all implementations, provide support for that pattern in the code, instead of trying to exhaustively support every use case we can think of. 

With this PR basically we do this, with a transparent way. So again don't understand why is having an Optional in desired makes this more clean, it makes the whole api very confusing. We just separating the concerns that should be separated, and as @scrocquesel mentioned not use fat classes, that does everything. In this way also is nicely extensible.

> Having a class hierarchy doesn't address the generality problem, it just keeps adding special cases that make the code base more confused, more difficult to maintain and harder to learn. It also forces extensions of the SDK to know about theses implementations instead of relying on a simpler, generic pattern thus making them more fragile with respect to changes in the implementation.

This is I think crucial, we add special cases with the desired optional, for an implementation that is already targets the generic use cases. That is hard to understand. Much harder then having a dedicated abstraction for a sub-problem.

metacosm:
> > Then just implement the interface. No need for an empty class that doesn't add anything useful.
> 
> The problem with this, that the `KubernetesDependentResource` already would implement Generic and ReadOnly interface as the result. What is a bigger smell IMHO.

Extending an empty class is a bigger smell than implementing a marker interface because it forces you into a class hierarchy, which is a bad pattern for frameworks/SDK. If you think that's a smell, then that's a problem with the design. The bigger problem is whether you extend the implementation or implement the interface, the SDK still cannot enforce the read-only functionality, which is the bigger smell of all, imo. :)

> **IMHO We add special cases with the "optional desired", for an implementation that is already targets the generic use cases. That is hard to understand. Much harder then having a dedicated abstraction for a sub-problem.**

It's very simple, conceptually, though: if, for any reason, your implementation doesn't know what to return at a specific time, then you return `empty`. As simple as that. This doesn't overload the concepts, it's very natural for an implementation to not have a desired state at some point because it might not be ready at this time, or because it never has a desired state at all. Returning `Optional` covers more cases than just the read-only case, without having to create 5 new classes just to add a behavior that is not even enforced in the SDK.

metacosm:
> > the SDK still cannot enforce the read-only functionality, which is the bigger smell of all, imo.
> 
> Hmm, but how is possible to do that? I mean even with traits, or with the desired, it's basically you can always do that. So not sure how you mean it exactly.

If `desired` returns `empty` then you can enforce that you're in the not created/not updated case for this particular reconciliation (note that what is returned doesn't need to be constant, it can vary over time and depend on external conditions). For the read-only case, then `desired` would always return `empty`  and the default implementation would enforce this read-only situation because `update` or `create` would never be called. Granted, this would only work in the "managed" case. If people are implementing `reconcile` directly, then, obviously, all bets are off. 

> 
> > It's very simple, conceptually, though: if, for any reason, your implementation doesn't know what to return at a specific time, then you return empty. As simple as that. This doesn't overload the concepts, it's very natural for an implementation to not have a desired state at some point because it might not be ready at this time, or because it never has a desired state at all. Returning Optional covers more cases than just the read-only case, without having to create 5 new classes just to add a behavior that is not even enforced in the SDK.
> 
> I understand what you mean, and I agree that it does not bring too much complexity in that regard. The problem is it breaks the abstraction. So the GenericDependentResource meant to describe the cases when the reconciliation looks exactly like this:
> 
> ```java
>  var actual = getResource(primary);
>     var desired = desired(primary, context);
>     if (actual.isEmpty()) {
>       create(desired, primary, context);
>     } else {
>       if (!match(actual.get(), desired, context)) {
>         update(actual.get(), desired, primary, context);
>       }
>     }
> ```
> 
> What is the most common case, when an controller owns the resources, and reconciles it. Whit optinal desired, we cover that and some additional more exotic case, what could desire a completely different abstraction, since we don't just kubernetes but also other external resources that will implement that. In this case this is the read only. So I don't understand this part: Why would we combine exactly these two use cases? So the only editable is not there, how that should there added?

I'm not sure how returning `Optional` breaks the abstraction. It only makes things more precise by letting the dependent communicate to the SDK that it doesn't have a `desired` state at this point in time. This works equally as well for non-Kubernetes resources.

The editable-only case would be supported as follows: `desired` would return `empty` until the dependent has been created externally. With the following implementation, `create` would not be called until `desired` until the resource would be ready to be updated:

```java
     var maybeDesired = desired(primary, context);
     maybeDesired.ifPresent(desired -> {
       if (maybeActual.isEmpty()) {
         create(desired, primary, context);
       } else {
         if (!match(actual, desired, context)) {
           update(actual, desired, primary, context);
         }
       }
     });
```

This is, indeed, a little awkward because this would mean that `desired` would need to check whether the resource exists or not.

However, in the current implementation, the abstraction is ill-defined, because why should editable-only be configured via annotation while read-only is implemented by a subclass / marker interface. So there's an asymmetry here that shouldn't be there. If one feature is configured by annotation, then the other should be as well. If the feature is implemented by a subclass then both should be as well, and in this case, this means adding more classes to support the combinations (i.e. you would need an `EditableOnly` version of the classes and a version for `CreatableOnly` for the case where you only want to create the dependent and never need to update it). I still think that a trait-based implementation would be better. Going to see how this goes.
 
> So what I try to say, is the good abstractions of the problems/cases should have the highest priority. And here the generic and the read-only cases deserves it's own abstraction under dependent resource. At least for me this is much more important than having done the current scope (that will probably be extended) with **some adjustments to abstractions that has well defined scopes**.

Except that the abstraction is not good or at least doesn't work for all cases. Why should read-only be more important than editable-only or creatable-only? The current implementation doesn't address this problem. I do agree, though, that returning `Optional` for `desired` probably isn't the best abstraction as it's not completely explicit. That said, you could create a class hierarchy that provides the proper behavior based on this, if you really wanted to (which I don't want to) 😄:

read-only: just return `empty` for `desired` all the time
create-only: return `true` for `match` all the time
editable-only: return `empty` for `desired` until the resource exists (implies that `desired` more or less needs to call `getResource`)

> 
> > This doesn't overload the concepts, it's very natural for an implementation to not have a desired state
> 
> It's not that natural, basically then it's not a dependent resource, it's an informer. Not sure if we should actually even solve this. I mean we have to just because we want to have it nicely done with annotations, without that actually we don't even need this. We could just reference the informer from the dependent resource what is using it as an input.

This just isn't true. It often happens when you reconcile dependents that there just isn't any valid state to create/update at some point in time. Granted, though, that enters the domain of managing dependencies between dependents.

cortex93:
Trying to keep pace.

Readonly(informer only) resource is something a reconciler needs to know the state. So in this sense, it's a dependent but it can't be reconciled. It's not depending on its state, it is intrisics. And to enforce that, its public contract should not expose reconcile/delete. Having a base class that expose reconcile/delete from the ground up tell that every subclass can be reconciled which is not true. And as such, whatever that used them must reconcile them and delete.

I understand the need for being standalone but can some logic be put in the manager ? This way, only traits can be on the dependent ? Trying to put the managing code at the same place as the behavior itself made things not SRP and force to have a base class with all the complexity.

I agree that clean separation of concern will add more code and classes. There is a tradeoff between less code and clean object modeling.

```java

public interface DependentResource<R, P extends HasMetadata> {
  Optional<R> getResource(P primaryResource);
}

public interface Desiring { R desired(...); }
public interface Create implements Desiring {  void create(...); }
public interface Update implements Desiring {  void update(...); bool match(...); }
public interface Delete{  void delete(...); }

public abstract class KubernetesDependentResource<R, P extends HasMetadata> implements
   DependentResource, EventSourceProvider, Create, Update, Delete {
}

public DependentResourceManager {

@Override
  public UpdateControl<P> reconcile(P resource, Context context) {
    initContextIfNeeded(resource, context);
    dependents.forEach(dependent -> {
      if (dependent instanceof Create/Update) {
        if (maybeActual.isEmpty()) {
          if (dependent instanceof Create) {
            creator.create(desired, primary, context);
          }
        } else if (dependent instanceof Update ) {
          if (!matcher.match(actual, desired, context)) {
            updater.update(actual, desired, primary, context);
          }
        }
      }
      });
    return UpdateControl.noUpdate();
  }

  @Override
  public DeleteControl cleanup(P resource, Context context) {
    initContextIfNeeded(resource, context);
    dependents.stream()
        .filter(dependent -> dependent instanceof Delete)
        .forEach(Delete::delete);
    return Reconciler.super.cleanup(resource, context);
  }
```

In this case, there is no need for default/noop behavior in the class hierarchy, if needed, another manager can handle different use case of mixins. I can even enrich the logic and reuse the existing class. For exemple, adding else case (https://github.com/java-operator-sdk/java-operator-sdk/issues/957).

But then, what is standalone is not the dependent itself but it's manager.

scrocquesel:
Trying to keep pace.

Readonly(informer only) resource is something a reconciler needs to know the state. So in this sense, it's a dependent but it can't be reconciled. It's not depending on its state, it is intrisics. And to enforce that, its public contract should not expose reconcile/delete. Having a base class that expose reconcile/delete from the ground up tell that every subclass can be reconciled which is not true. And as such, whatever that used them must reconcile them and delete.

I understand the need for being standalone but can some logic be put in the manager ? This way, only traits can be on the dependent ? Trying to put the managing code at the same place as the behavior itself made things not SRP and force to have a base class with all the complexity.

I agree that clean separation of concern will add more code and classes. There is a tradeoff between less code and clean object modeling.

```java

public interface DependentResource<R, P extends HasMetadata> {
  Optional<R> getResource(P primaryResource);
}

public interface Desiring { R desired(...); }
public interface Create implements Desiring {  void create(...); }
public interface Update implements Desiring {  void update(...); bool match(...); }
public interface Delete{  void delete(...); }

public abstract class KubernetesDependentResource<R, P extends HasMetadata> implements
   DependentResource, EventSourceProvider, Create, Update, Delete {
}

public DependentResourceManager {

@Override
  public UpdateControl<P> reconcile(P resource, Context context) {
    initContextIfNeeded(resource, context);
    dependents.forEach(dependent -> {
      if (dependent instanceof Create/Update) {
        if (maybeActual.isEmpty()) {
          if (dependent instanceof Create) {
            creator.create(desired, primary, context);
          }
        } else if (dependent instanceof Update ) {
          if (!matcher.match(actual, desired, context)) {
            updater.update(actual, desired, primary, context);
          }
        }
      }
      });
    return UpdateControl.noUpdate();
  }

  @Override
  public DeleteControl cleanup(P resource, Context context) {
    initContextIfNeeded(resource, context);
    dependents.stream()
        .filter(dependent -> dependent instanceof Delete)
        .forEach(Delete::delete);
    return Reconciler.super.cleanup(resource, context);
  }
```

In this case, there is no need for default/noop behavior in the class hierarchy, if needed, another manager can handle different use case of mixins. I can even enrich the logic and reuse the existing class. For exemple, adding else case (https://github.com/java-operator-sdk/java-operator-sdk/issues/957).

But then, what is standalone is not the dependent itself but it's manager.

PR #961, opened 2022-02-21
fix: sonar security issue

csviri:


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
```diff
@@ -54,6 +54,13 @@ public void initContext(MySQLSchema primary, Context context) {
     context.put(MYSQL_SECRET_USERNAME, userName);
   }
 
+  private String randomPassword(int length) {
+    SecureRandom random = new SecureRandom(); // Compliant for security-sensitive use cases
```
metacosm, line 58: We shouldn't need to do this, though… We need a way to exclude such false positive from Sonar.

metacosm (approved):

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
csviri: Did not find a way, not sure if possible for the security issues.

PR #963, opened 2022-02-21
feat: dependents use traits to specify which features they support

metacosm:
This is meant to be a cleaner alternative to both #956 and #958.


scrocquesel:
We do have sort of the same idea (https://github.com/java-operator-sdk/java-operator-sdk/pull/958#issuecomment-1047286996)

And I see that you were force to slip some logic in the manager. But if you do for delete why not for the reconcile ? The dependent is not really standalone anymore.
BTW, the StandaloneDependentTestReconciler assume `delete` should not be called while it should because `KubernetesDependentResource` implements `Deleter`. If has used the appropriate DependentManager, it shouldn't have been an issue.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
```diff
@@ -3,29 +3,49 @@
...
 
+  private final Creator<R, P> creator;
+  private final Updater<R, P> updater;
+  private final Matcher<R> matcher;
+
+  @SuppressWarnings("unchecked")
+  public AbstractDependentResource() {
+    creator = this instanceof Creator ? (Creator<R, P>) this : Creator.NOOP;
+    updater = this instanceof Updater ? (Updater<R, P>) this : Updater.NOOP;
+    matcher = this instanceof Matcher ? (Matcher<R>) this : Matcher.DEFAULT;
```
csviri, line 17: I'm not sure about the matcher. It's existence is never checked, and this is a very generic problem. I would rather have it as a method with defaul implementation that can be overriden. Also in general does not fit in this CRUD picture of interfaces.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
```diff
@@ -72,7 +73,10 @@ public DependentResourceManager(Controller<P> controller) {
   @Override
   public DeleteControl cleanup(P resource, Context context) {
     initContextIfNeeded(resource, context);
-    dependents.forEach(dependent -> dependent.delete(resource, context));
+    dependents.stream()
+        .filter(dependent -> dependent instanceof Deleter)
```
csviri, line 77: As mentioned above, would rather put back the delete into `DependentResource`.

csviri (changes requested):
In general like this implementation. Made some comments that might be worth to revisit and discuss but in general it is good IMHO.

Just one thing I don't really get, does this solve the Read Only case for KubernetesDependetResource? since it is already implements all the trait interfaces. Eventually those will be feature flags?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
```diff
@@ -8,7 +8,5 @@
 public interface DependentResource<R, P extends HasMetadata> {
   void reconcile(P primary, Context context);
 
-  void delete(P primary, Context context);
```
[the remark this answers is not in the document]
sclorng: reconcile and delete are two sides of the same coin. Basically, delete is just reconcile with primary being null. So maybe 

`void reconcile(Optional<P> primary, Context context)` can embrace both. And then, in the default implementation, call delete if primary is not present and it implements `Deleter`.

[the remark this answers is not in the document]
csviri: This would not work, basically the delete is only needed if there is non kubernetes objects managed, so no owner reference set. But then we don't know what to delete if there is no primary as input.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
csviri: Yes, that could be an other way to have it nicer.
But again matching is quite a general problem, that can be solved out of the box, for most cases (at least for K8S).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
sclorng: I agree, there is some inconsistencies. Either Manager deal with all traits abstraction, or it is deal within abstractDependentResource  base class implementation. The former allow for more extensibility and reuse, the latter allow for simpler standalone use case.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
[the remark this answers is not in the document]
sclorng: > This would not work, basically the delete is only needed if there is non kubernetes objects managed or no owner reference set. But then we don't know what to delete if there is no primary as input.

That's true. Still, delete and reconcile are tight together in some way.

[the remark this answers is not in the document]
csviri: Currently the delete is called if there is finalizer in place, and the custom resource is marked for deletion.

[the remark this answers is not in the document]
csviri: In terms of workflows there might be an actual relationship (eventually planned to manage workflows automatically), the delete is a reverse workflow maybe just calling delete instead of reconcile? 
I mean there is such a relationship for sure, but cannot think of other.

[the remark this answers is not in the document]
sclorng: > In terms of workflows there might be an actual relationship (eventually planned to manage workflows automatically), the delete is a reverse workflow maybe just calling delete instead of reconcile? I mean there is such a relationship for sure, but cannot think of other.

Yes, delete should be understood as remove desired on actual. For creatable, it means to delete the actual. For editonly, it means to undo the patch on actual
Regarding finalizers, in the future, maybe the default can be guess automatically from dependent and set if one dependent is actually implementing traits Deleter.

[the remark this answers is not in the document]
csviri: What is undo patch? Not sure that is possible in general. How it would know the previous state?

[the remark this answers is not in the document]
sclorng: some operators just need to patch a dependent resource. If you don't handle spec change, the dependent knows the content of the patch from the spec. Otherwise, yes, applied spec should be persist somewhere to be able to undo it.

Currently, in my operator, I have something like below and I'm pretty sure it can fit in the dependent model

```java
    @Override
    public UpdateControl<Primary> reconcile(Primary primary,
            Context context) {

        var secondaryResource = context.getSecondaryResource(Dependent.class);
        return secondaryResource.map((Dependent dependent) -> {
            // ensure the secondary resource has still the desired patch applied
            // and update the status
            return ensureIsPatched(dependent, primary);
        }).orElseGet(() -> {
            // give a chance to update the primary resource status
            // for eg. adding a False Ready Condition with reason "missing secondary"
            return missingSecondary(primary);
        });
    }

    @Override
    public DeleteControl cleanup(Primary primary, Context context) {

        var target = context.getSecondaryResource(Dependent.class);
        target.ifPresent((Dependent dependent) -> {
            ensureIsNotPatched(dependent, primary);
        });

        return DeleteControl.defaultDelete();
    }
```

ensureIsPatched actually do something like

```java
k8sclient.resources(actual).edit( (a) -> applyPatch(a));
primary.getStatus().setObservedSpec(primary.getSpec());
return UpdateControl.updateStatus(primary);
```

and ensureIsNotPatched

```java
Optional.ofNullable(primary.getStatus().getObservedSpec()).map(spec -> {
            this.client.resources(actual).edit(o -> undoPatch(o, spec));    
        });
```

The main issue is that while it's a dependent, the reconcile logic is a bit different and having this logic implemented in the dependent base clase itself do not allow much of reuse. I need to take time to look at it with traits aspect.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/AbstractDependentResource.java
metacosm: Except that people might want to avoid using the rather inefficient generic implementation. Some of the examples actually override the method to use an optimised version. I've put the `match` method on `Updater` to make the link between both explicit. I've kept the `Matcher` interface to be able to provide standalone implementations that can be reused, in particular in `GenericKubernetesResourceMatcher`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/DependentResourceManager.java
metacosm: Done. `Deleter` is now managed in a way that consistent with the rest of the traits.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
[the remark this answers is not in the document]
metacosm: Interesting use case! 😄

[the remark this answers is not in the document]
csviri: @sclorng  I see, but this is quite an exotic case, I mean probably not makes sense to generalize the framework to support this case at this point. But certainly would be doable just overriding the appropriate methods.

csviri:
Do I understand correctly that in a read only resource the initializer would be overriden with NOOP operations? 
So it would be something Like this in constructor?

```java

ReadOnlyDeployment() {
   init(Creator.NOOP, Updater.NOOP, Deleter.NOOP);
}

```

One more thing that I might consider, is replacing trait methods with feature flags (so boolean values for editor, updator, deleter), and for NOOP that would mean the method is empty - so in `AbstractDependentResource` would have already an empty implementation. Wouldn't be more straigforward?

In more detail:

```java

public abstract class AbstractDependentResource<R, P extends HasMetadata>
    implements DependentResource<R, P> {

  protected boolean creator;
  protected boolean updater;
  protected boolean deleter;

  public AbstractDependentResource(boolean creator, boolean updater, boolean deleter) {
      this.creator = creator;
      this.updater = updater;
      this.updater = updater;
  }

  @SuppressWarnings({"unchecked", "rawtypes"})
  protected void init(Creator defaultCreator, Updater defaultUpdater, Deleter defaultDeleter) {
    creator = this instanceof Creator ? (Creator<R, P>) this : defaultCreator;
    updater = this instanceof Updater ? (Updater<R, P>) this : defaultUpdater;
    deleter = this instanceof Deleter ? (Deleter<P>) this : defaultDeleter;
  }


  @Override
  public void reconcile(P primary, Context context) {

    if (creator || updater) {
      if (maybeActual.isEmpty()) {
        if (creator) {
          create(desired, primary, context);
        }
      } else {
        if (updater  && match(actual, desired, context)) {
          update(actual, desired, primary, context);
        }
      }
    }
  }

   boolean match(R actualResource, R desiredResource, Context context) {
      Objects.equals(actualResource, desiredResource);
   }

  public void update(R actual, R target, P primary, Context context) {
  }
  public void create(R target, P primary, Context context) {
  }

````

The positive side would be that it's really just flags, for me it feels that there is quite a lot ceremony to allow these methods for CUP (create,update,delete), while these are anyways implemented in KubernetesDependentResource. Just in little compex way with private base classes in the background.

And in the read only version, or just edit or just create version (not sure we want something like just create actually, but lets say some might need it), these flags needs to be configured anyways just in constructor just in form of methods - what is a little strange.

```java

public class OnlyEditKubernetesClass extends KubernetesDependentResource{
 
public JustEditClass() {
  init(NOOP, new UpdateDependentOperation(), NOOP)
}

// ...
}

```  



What do you think?

metacosm:
The idea is that `DependentResource` implementations need to implement the traits to support the feature. That should be the only thing that is required if you somehow extends `AbstractDependentResource`. By default, if your implementation doesn't implement any traits then it will be read-only. Each trait you implement adds a feature: if you want to be edit-only, you only need to implement `Updater`, if you want to be create-only, you only need to implement `Creator`, if you want to be delete-only (not sure that's useful but this makes things coherent), you only need to implement `Deleter`.

The `init` method is only there to provide sub-classes of `AbstractDependentResource` with a way to specify their own default behavior if one of their sub-classes wants to support one of the trait without actually providing an implementation because they are satisfied with the behavior provided by their super-class. This is indeed a little tricky but it mostly makes sense for `KubernetesDependentResource` where we want to provide default, generic behavior that could be used with minimal effort without actually implementing the traits directly. This allows for a sub-class of `KubernetesDependentResource` to implement `Creator` for example without having to actually implement the method since it will inherit the behavior from `KubernetesDepedentResource` automatically (because `KubernetesDependentResource` implements all the traits without actually declaring it does, thus allowing sub-classes to inherit default implementations for free).

metacosm:
The problem with flags is that you could have an implementation that sets the flags without actually providing the implementation which makes things incoherent. 

The SDK should try to enforce the semantics: if a user says that their dependent resource needs to be created, then it should enforce that there is indeed logic to create the dependent, though, granted, that logic could be empty. 

The act of requiring to implement an interface makes things more explicit than a boolean in a constructor (which could also be passed wrong, i.e. in the wrong order, it's always tricky to have several parameters of the same type on methods).

Another aspect of flags is that these should probably be made available to the annotation layer making it more complex as well. The first dependent resource implementations actually had these flags but I now think that trait interfaces are a more elegant solution because it extracts the information from the code (i.e. what you actually implement) instead of relying on an explicit configuration step, which creates the problem of having to deal with the configuration (and you've been removing the configuration layers that the non-standalone version had) and with the fact that there could be a mismatch between what the configuration says and what is actually implemented. Using traits solves the problem elegantly I think by making the code the configuration 😄.

csviri:
I understand the goal here, and the structure. I really like it, just wanted to discuss if it could be simplified by the flags.

Maybe we could add a sample Read Only and Update Only implementation, how it would look like, just we can see the code together, but all I want say that at the end this is how an edit only kubernetes would look like at the end:

```java
public class OnlyUpdateKubernetesResource extends KubernetesDependentResource<Deployment, Primary> {
 
public JustEditClass() {
  init(NOOP, new UpdateDependentOperation(), NOOP)
}

 @Override
  public Deployment desired(Primary primary, Context context) {
   //...
  }
}
```

Where this part:

`init(NOOP, new UpdateDependentOperation(), NOOP)` is basically same as `init(false, true, false)`; Just more unnatural/complex to understand. So in that regard it's already feature flags used.

That is true that there can be a mismatch for flags and overriden methods. We could provide adapters, which overrides the only the wanna be empty methods and sets the flags. In that regard the current implementation is really good.

I think we should create issues to provide abstract dependent resource for external ones, since there will be the same issue as with kubernetes ones for the getResource(). Will do that.

But I'm otherwise fine to merge this, it's really good.

csviri:
Also we should add unit tests :)

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -75,45 +86,80 @@ public void configureWith(ConfigurationService configurationService,
...
-        .create(target);
+  private class UpdateDependentOperation extends DependentOperation implements Updater<R, P> {
+    private final Matcher<R> matcher;
+    private final ResourceUpdatePreProcessor<R> processor;
+
+    @SuppressWarnings("unchecked")
+    public UpdateDependentOperation() {
+      super("Updating");
+      matcher = GenericKubernetesResourceMatcher.matcherFor(resourceType());
+      processor = KubernetesDependentResource.this instanceof ResourceUpdatePreProcessor
```
csviri, line 133: Maybe we should have a setter for this still on KDR but that can be an another PR when needed

metacosm: There actually was one but I removed it with the idea that now everything is defined in a similar fashion using traits to override the default behavior.

csviri: The thing is that for standalone DRs, it wouldn't be necessary to make subclasses, but it is how it is. So we will se how it evolves, maybe if someone for standalone ones would not want to do subclass we can provide an "adaptor".

sample-operators/tomcat-operator/src/main/java/io/javaoperatorsdk/operator/sample/DeploymentDependentResource.java
```diff
@@ -5,11 +5,14 @@
...
 import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.Creator;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.Updater;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependent;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResource;
 
 @KubernetesDependent(labelSelector = "app.kubernetes.io/managed-by=tomcat-operator")
-public class DeploymentDependentResource extends KubernetesDependentResource<Deployment, Tomcat> {
+public class DeploymentDependentResource extends KubernetesDependentResource<Deployment, Tomcat>
+    implements Creator<Deployment, Tomcat>, Updater<Deployment, Tomcat> {
```
csviri, line 15: Cannot this be done by the init?

metacosm: No, the whole point is to declare what your dependent is doing explicitly. `init` is only there to set the defaults for subclasses of `AbstractDependentResource`.

csviri: Hmm, won't in most cases actually a full CRUD resource? Should we have an adaptor for that?

csviri (approved):

PR #966, opened 2022-02-22
Utility around curl in Cluster

andreaTP:
Moving the code for running curl in cluster to it's own utility class as it's a really useful thing to have.


metacosm (approved):
LGTM

csviri (approved):
LGTM

PR #979, opened 2022-02-28
feat: dependent resource context + my sql e2e test improvements

csviri:


csviri:
Added access to managed dependent resources, not sure if that is there the map in the context makes sense anymore.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -281,4 +281,8 @@ public void stop() {
       eventSourceManager.stop();
     }
   }
+
+  public DependentResourceManager<R> getDependents() {
```
metacosm, line 285: Do we really need to expose `DependentResourceManager` here? Can we just expose the dependents directly?

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
```diff
@@ -26,42 +27,23 @@
...
   @Override
   public UpdateControl<MySQLSchema> reconcile(MySQLSchema schema, Context context) {
     // we only need to update the status if we just built the schema, i.e. when it's present in the
     // context
-    return context.get(BUILT_SCHEMA, Schema.class).map(s -> {
-      updateStatusPojo(schema, context.getMandatory(MYSQL_SECRET_NAME, String.class),
-          context.getMandatory(MYSQL_SECRET_USERNAME, String.class));
+    Secret secret = context.getSecondaryResource(Secret.class).orElseThrow();
+    SchemaDependentResource schemaDependentResource = context.managedDependentResourceContext()
+        .getDependentResource(SchemaDependentResource.class);
```
metacosm, line 43: I think it'd be better to get the dependent via its class in a way similar to what we do for secondary resources, i.e. `getDependentResource(Schema.class)`, though I guess this way this side-steps the issue of what to do when there are multiple dependents with the same type? Maybe provide both versions?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
csviri: So this is just a getter for that in the controller. For general purpose. Controller is not exposed to the user so don't think this is particularly bad.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
csviri: > I think it'd be better to get the dependent via its class in a way similar to what we do for secondary resources, i.e. getDependentResource(Schema.class)

That is how it's done, just hidden behind dedicated context. Or?

>  though I guess this way this side-steps the issue of what to do when there are multiple dependents with the same type? Maybe provide both versions?

I intentionally did not cover this, since on denepnds_on we plan to have names of the resources, so I would add that as part of it?

csviri:
> Added access to managed dependent resources, not sure if that is there the map in the context makes sense anymore. @metacosm

What do you thin about this ? @metacosm

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: I'd rather not expose this class if possible since I consider it as an internal class.

csviri:
> > > Added access to managed dependent resources, not sure if that is there the map in the context makes sense anymore. @metacosm
> > 
> > 
> > What do you thin about this ? @metacosm
> 
> We can remove it for now and re-add it if needed.

oki, will create a separate PR, others are fixed.

PR #1010, opened 2022-03-09
feat: retrieve ConfigurationService from ConfigurationServiceProvider

metacosm:
Depends on #1013


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -228,20 +225,7 @@ public void start() throws OperatorException {
   }
 
   private ConfigurationService configurationService() {
```
csviri, line 227: Do we need this method? Would just delete it.

csviri:
Does this help with quarkus extension too? or the point is just to have a global config serivce?

csviri (approved):
LGTM , just 1 minor comment

metacosm:
> Does this help with quarkus extension too? or the point is just to have a global config serivce?

I'm not sure yet… It started with the idea that it would help but I need to try it out…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: Done.

PR #1013, opened 2022-03-09
feat: add non-apt-tied behavior annotation and configuration to core

### LINKED ISSUE #1012: Consider moving `DefaultConfigurationService` and `AnnotationControllerConfiguration` into core module
The initial thinking was that there were not needed in the core but not having them is actually causing more problems than having them there.

metacosm:
Fixes #1012


csviri:
LGTM,

(maybe one thing to consider to name differently the two `AnnotationControllerConfiguration` classes.)

csviri (approved):

metacosm:
> (maybe one thing to consider to name differently the two `AnnotationControllerConfiguration` classes.)

They're in different packages and modules and we do something similar already with `ControllerConfiguration` but yes, maybe we should rename the new one. Any idea?

metacosm:
Will do these changes in a follow-up PR if needed.

PR #1020, opened 2022-03-10
feat: dependent resource inherits namespaces from controller config

csviri:
The primary objective was to inherit the default namespace to the dependent informers if not configured explicitly (standalone). 
For the we need the `ControllerConfiguration` not `ConfigurationService` ( but that is accessible from controller config). So unified also the contexts to provide the controller configuration instead of configuration service. (This part is optional)


metacosm (approved):
I'm OK with the change but that is a breaking change that will impact users (they will have to change their reconciler implementations).

PR #1021, opened 2022-03-10
fix: renaming mapper primary <-> secondary

csviri:
@metacosm what do you think about this renaming. Not sure how much actual backwards compatibiilty issues it will make since usually implemented as lambda function. But we can document these minor changes.
The current names are very obscure for me, have to think about every time I check.

Also would add `glossary` page to the web page docs, to cleary define:
- Secondary Resource (what is a resource actually controller creates)
- Dependent Resource (is the dependent resource feature)

If you agree. 
Later we can extend glossary with additional notions.


metacosm (approved):
That's fine with me. The old names were indeed confusing. That said, the new names are not that much better but I don't have any better proposal. 😬

PR #1030, opened 2022-03-14
Reconcile api proposal

csviri:
This PR is about proposing an more info about reconciliation (currently for standalone mode). And also a discussion starter for the topic :)

see also: https://github.com/java-operator-sdk/java-operator-sdk/issues/957


metacosm (approved):
LGTM

PR #1057, opened 2022-03-21
fix: set default CRD check to false

csviri:


metacosm:
The problem is that the behavior was actually wrong in the core, which ended up behaving like the default was `false` but, as far as I'm aware at least, this behaved properly in both the Spring Boot starter and the Quarkus extension… so in this case, this would be a breaking change.

csviri:
Will created an another PR where we just log that checking CRD, and how to turn it off. That is always useful.

csviri:
> > It would not be breaking I think, I mean we can still keep the default true on those extensions. It won't be breaking in the sense that this is less restrictive. Everything will work that worked until now. But in pure java is actually a breaking change, users already complained on our discord channel.
> 
> Yes but they're actually complaining that a bug was fixed because that was the intended behavior wink

Just users did not know that, nowhere documented, and just everybody used the default. So it can be quite annoying. But basically it also should not be default `true` IMHO. We should aim for production with defaults.

csviri:
> You could make the same point about the default behavior of watching all namespaces, which is not realistic for many operators…

I don't understand how it is related. Watch all namespaces is absolutely a valid use case, lots of operators are deployed that way. But don't think there is a use case when we should check CRD on prod during startup. That might be on a dev environment.

metacosm (approved):

Issue #1038, opened 2022-03-15
DependentResource CDI awareness


scrocquesel:
I think JOSDK will need some kind of support for proper integration with CDI in Quarkus.

Actually, I find out : 
- `KubernetesDependentResource::resourceType` calls `Utils::getFirstTypeArgumentFromExtendedClass` which is not aware of proxy class. I have done an ugly fix but there is sure something better to do https://github.com/quarkiverse/quarkus-operator-sdk/pull/267#discussion_r823525514
- `DependentResourceManager::createAndConfigureFrom` use reflection to create instance of DependentResourceSpec. As such, it is not possible to have bean injection in Quarkus for DependentResource.

metacosm:
Yes, the resource type is a problem that I haven't gotten around to work on. 

For the `DependentResourceSpec` issue, the first implementation had a configuration object that got removed for the standalone version. If I'm being completely honest, I would have rather not bother with the standalone use case for now.

scrocquesel:
> If I'm being completely honest, I would have rather not bother with the standalone use case for now.
Yep, but without a CDI aware DependentResourceManager, you can't use annotation.
Either, there is factory strategy configurable on @dependent (like mapstruct does), or the DependentResourceManager should be refactored to extract the setup responsability to an higher class and keep the reconcile/cleanup logic to an abstract class. Thus Quarkus can provide an instance with dependent provided at build time ?

metacosm:
Yes, what I'm considering is delegating the actual `DependentResource` creation to a factory that would be configured via `ConfigurationService` with a default implementation being the current one while the Quarkus implementation would create beans. 
For the resource type, I'm thinking of making it a method on `DependentResource` and make it explicit. The default implementation doesn't work in all instances and, hopefully, we should be able to have a no reflection implementation in Quarkus using synthetic beans for the `DependentResources`.

metacosm:
What should the default CDI scope be for the dependents? I'm leaning on `ApplicationScoped` at the moment but could be convinced otherwise…

metacosm:
Another question is how do you distinguish between dependents of the same type (e.g. your operator needs to have two dependents of type `Deployment` which might or might not be implemented by the same class)…

scrocquesel:
> What should the default CDI scope be for the dependents? I'm leaning on `ApplicationScoped` at the moment but could be convinced otherwise…

[...]

PR #1058, opened 2022-03-21
fix: retrieve DependentResource based on name instead of class

metacosm:
This causes issues if the class is proxied for any reason (e.g. because
it's a CDI bean).

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
@@ -3,12 +3,11 @@ package io.javaoperatorsdk.operator.api.config;
 import java.time.Duration;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Optional;
+import java.util.Map;
 import java.util.Set;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.config.dependent.DependentResourceSpec;
-import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 import io.javaoperatorsdk.operator.processing.event.source.controller.ResourceEventFilter;
 
 @SuppressWarnings({"rawtypes", "unchecked", "unused"})
@@ -22,7 +21,7 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
   private ResourceEventFilter<R> customResourcePredicate;
   private final ControllerConfiguration<R> original;
   private Duration reconciliationMaxInterval;
-  private final List<DependentResourceSpec<?, ?>> dependentResourceSpecs;
+  private final Map<String, DependentResourceSpec<?, ?>> dependentResourceSpecs;
 
   private ControllerConfigurationOverrider(ControllerConfiguration<R> original) {
     finalizer = original.getFinalizerName();
@@ -89,37 +88,15 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
     return this;
   }
 
-  public void replaceDependentResourceConfig(
-      Class<? extends DependentResource<?, R>> dependentResourceClass,
-      Object dependentResourceConfig) {
-
-    var currentConfig =
-        findConfigForDependentResourceClass(dependentResourceClass);
-    if (currentConfig.isEmpty()) {
-      throw new IllegalStateException("Cannot find DependentResource config for class: "
-          + dependentResourceClass);
+  public void replaceNamedDependentResourceConfig(String name, Object dependentResourceConfig) {
+    final var currentConfig = dependentResourceSpecs.get(name);
+    if (currentConfig == null) {
+      throw new IllegalArgumentException("Cannot find a DependentResource named: " + name);
     }
-    dependentResourceSpecs.remove(currentConfig.get());
-    dependentResourceSpecs
-        .add(new DependentResourceSpec(dependentResourceClass, dependentResourceConfig));
-  }
-
-  public void addNewDependentResourceConfig(DependentResourceSpec dependentResourceSpec) {
-    var currentConfig =
-        findConfigForDependentResourceClass(dependentResourceSpec.getDependentResourceClass());
-    if (currentConfig.isPresent()) {
-      throw new IllegalStateException(
-          "Config already present for class: "
-              + dependentResourceSpec.getDependentResourceClass());
-    }
-    dependentResourceSpecs.add(dependentResourceSpec);
-  }
-
-  private Optional<DependentResourceSpec<?, ?>> findConfigForDependentResourceClass(
-      Class<? extends DependentResource> dependentResourceClass) {
-    return dependentResourceSpecs.stream()
-        .filter(dc -> dc.getDependentResourceClass().equals(dependentResourceClass))
-        .findFirst();
+    dependentResourceSpecs.remove(name);
+    dependentResourceSpecs.put(name,
+        new DependentResourceSpec(currentConfig.getDependentResourceClass(),
+            dependentResourceConfig, name));
   }
 
   public ControllerConfiguration<R> build() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
@@ -90,7 +90,7 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
     return this;
   }
 
-  public ControllerConfigurationOverrider<R> replaceNamedDependentResourceConfig(String name,
+  public ControllerConfigurationOverrider<R> replacingNamedDependentResourceConfig(String name,
       Object dependentResourceConfig) {
     final var currentConfig = dependentResourceSpecs.get(name);
     if (currentConfig == null) {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
@@ -1,6 +1,7 @@
 package io.javaoperatorsdk.operator.api.config;
 
 import java.time.Duration;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -31,7 +32,8 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
     labelSelector = original.getLabelSelector();
     customResourcePredicate = original.getEventFilter();
     reconciliationMaxInterval = original.reconciliationMaxInterval().orElse(null);
-    dependentResourceSpecs = original.getDependentResources();
+    // make the original specs modifiable
+    dependentResourceSpecs = new HashMap<>(original.getDependentResources());
     this.original = original;
   }
 
@@ -88,7 +90,8 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
     return this;
   }
 
-  public void replaceNamedDependentResourceConfig(String name, Object dependentResourceConfig) {
+  public ControllerConfigurationOverrider<R> replaceNamedDependentResourceConfig(String name,
+      Object dependentResourceConfig) {
     final var currentConfig = dependentResourceSpecs.get(name);
     if (currentConfig == null) {
       throw new IllegalArgumentException("Cannot find a DependentResource named: " + name);
@@ -97,6 +100,7 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
     dependentResourceSpecs.put(name,
         new DependentResourceSpec(currentConfig.getDependentResourceClass(),
             dependentResourceConfig, name));
+    return this;
   }
 
   public ControllerConfiguration<R> build() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
@@ -1,6 +1,11 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent;
 
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.EMPTY_STRING;
+
 public @interface Dependent {
 
+  @SuppressWarnings("rawtypes")
   Class<? extends DependentResource> type();
+
+  String name() default EMPTY_STRING;
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
@@ -12,7 +12,8 @@ public interface DependentResource<R, P extends HasMetadata> {
 
   Class<R> resourceType();
 
-  default String name() {
-    return getClass().getCanonicalName();
+  @SuppressWarnings("rawtypes")
+  static String defaultNameFor(Class<? extends DependentResource> dependentResourceClass) {
+    return dependentResourceClass.getCanonicalName();
   }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
@@ -9,4 +9,10 @@ public interface DependentResource<R, P extends HasMetadata> {
   ReconcileResult<R> reconcile(P primary, Context<P> context);
 
   Optional<R> getResource(P primaryResource);
+
+  Class<R> resourceType();
+
+  default String name() {
+    return getClass().getCanonicalName();
+  }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
@@ -5,13 +5,47 @@ import java.util.Optional;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 
+/**
+ * An interface to implement and provide dependent resource support.
+ *
+ * @param <R> the dependent resource type
+ * @param <P> the associated primary resource type
+ */
 public interface DependentResource<R, P extends HasMetadata> {
+
+  /**
+   * Reconciles the dependent resource given the desired primary state
+   *
+   * @param primary the primary resource for which we want to reconcile the dependent state
+   * @param context {@link Context} providing useful contextual information
+   * @return a {@link ReconcileResult} providing information about the reconciliation result
+   */
   ReconcileResult<R> reconcile(P primary, Context<P> context);
 
+  /**
+   * Retrieves the dependent resource associated with the specified primary one
+   *
+   * @param primaryResource the primary resource for which we want to retrieve the secondary
+   *        resource
+   * @return an {@link Optional} containing the secondary resource or {@link Optional#empty()} if it
+   *         doesn't exist
+   */
   Optional<R> getResource(P primaryResource);
 
+  /**
+   * Retrieves the resource type associated with this DependentResource
+   *
+   * @return the resource type associated with this DependentResource
+   */
   Class<R> resourceType();
 
+  /**
+   * Computes a default name for the specified DependentResource class
+   *
+   * @param dependentResourceClass the DependentResource class for which we want to compute a
+   *        default name
+   * @return the default name for the specified DependentResource class
+   */
   @SuppressWarnings("rawtypes")
   static String defaultNameFor(Class<? extends DependentResource> dependentResourceClass) {
     return dependentResourceClass.getCanonicalName();
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
@@ -2,26 +2,44 @@ package io.javaoperatorsdk.operator.api.reconciler.dependent;
 
 import java.util.Optional;
 
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
 public class ReconcileResult<R> {
 
   private final R resource;
   private final Operation operation;
+  private final Exception error;
 
   public static <T> ReconcileResult<T> resourceCreated(T resource) {
-    return new ReconcileResult<>(resource, Operation.CREATED);
+    return new ReconcileResult<>(resource, Operation.CREATED, null);
   }
 
   public static <T> ReconcileResult<T> resourceUpdated(T resource) {
-    return new ReconcileResult<>(resource, Operation.UPDATED);
+    return new ReconcileResult<>(resource, Operation.UPDATED, null);
   }
 
   public static <T> ReconcileResult<T> noOperation(T resource) {
-    return new ReconcileResult<>(resource, Operation.NONE);
+    return new ReconcileResult<>(resource, Operation.NONE, null);
   }
 
-  private ReconcileResult(R resource, Operation operation) {
+  public static <T> ReconcileResult<T> error(T resource, Exception error) {
+    return new ReconcileResult<>(resource, Operation.ERROR, error);
+  }
+
+  @Override
+  public String toString() {
+    return getResource()
+        .map(r -> r instanceof HasMetadata ? ResourceID.fromResource((HasMetadata) r) : r)
+        .orElse("no resource")
+        + " -> " + operation
+        + getError().map(e -> " (" + e.getMessage() + ")").orElse("");
+  }
+
+  private ReconcileResult(R resource, Operation operation, Exception error) {
     this.resource = resource;
     this.operation = operation;
+    this.error = error;
   }
 
   public Optional<R> getResource() {
@@ -32,7 +50,11 @@ public class ReconcileResult<R> {
     return operation;
   }
 
+  public Optional<Exception> getError() {
+    return Optional.ofNullable(error);
+  }
+
   public enum Operation {
-    CREATED, UPDATED, NONE
+    CREATED, UPDATED, NONE, ERROR
   }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
@@ -9,22 +9,17 @@ public class ReconcileResult<R> {
 
   private final R resource;
   private final Operation operation;
-  private final Exception error;
 
   public static <T> ReconcileResult<T> resourceCreated(T resource) {
-    return new ReconcileResult<>(resource, Operation.CREATED, null);
+    return new ReconcileResult<>(resource, Operation.CREATED);
   }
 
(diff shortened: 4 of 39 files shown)


scrocquesel:
Could the name be provided by the `DependentResourceSpec` and set during `createAndConfigureFrom` ?
- This would allow to have two dependents of the same type with different name. Related to your question https://github.com/java-operator-sdk/java-operator-sdk/issues/1038#issuecomment-1068488775. Even if without configuration, I doubt it would be that usefull.
- This would allow to have the real static class name from annotation in logs instead of I guess the proxy one return by `getClass.getCanonicalName` ?

scrocquesel:
> Could the name be provided by the `DependentResourceSpec` and set during `createAndConfigureFrom` ?
> 
> * This would allow to have two dependents of the same type with different name. Related to your question [DependentResource CDI awareness #1038 (comment)](https://github.com/java-operator-sdk/java-operator-sdk/issues/1038#issuecomment-1068488775).
> * This would allow to have the real static class name from annotation in logs instead of I guess the proxy one return by `getClass.getCanonicalName` ?

And name could be latter override by configuration on the annotation if needed

csviri:
Looks good so far, my question is also that, the name should be mandatory in the annotation? Or how do you see that?

sclorng:
> Looks good so far, my question is also that, the name should be mandatory in the annotation? Or how do you see that?

If not provided, then it should take the type canonical name and raise an error if there is a dup

csviri:
> > Looks good so far, my question is also that, the name should be mandatory in the annotation? Or how do you see that?
> 
> If not provided, then it should take the type canonical name and raise an error if there is a dup

Yep, probably a sensible default makes sense in this case.

metacosm:
> Could the name be provided by the `DependentResourceSpec` and set during `createAndConfigureFrom` ?

Yes, that's the plan indeed.

> 
> * This would allow to have two dependents of the same type with different name. Related to your question [DependentResource CDI awareness #1038 (comment)](https://github.com/java-operator-sdk/java-operator-sdk/issues/1038#issuecomment-1068488775).

> * This would allow to have the real static class name from annotation in logs instead of I guess the proxy one return by `getClass.getCanonicalName` ?

I think so, yes.

metacosm:
> > Looks good so far, my question is also that, the name should be mandatory in the annotation? Or how do you see that?
> 
> If not provided, then it should take the type canonical name and raise an error if there is a dup

Yes, that is the plan but this PR is not finished yet. Great to see some early feedback that validates what I was thinking, though! 😄

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -71,15 +73,22 @@ public Controller(Reconciler<P> reconciler,
...
+    final var size = specs.size();
+    if (size == 0) {
+      dependents = Collections.emptyMap();
+    } else {
+      final var dependentsHolder = new HashMap<String, DependentResource>(size);
+      specs.forEach((name, drs) -> {
+        final var dependent = createAndConfigureFrom(drs, kubernetesClient);
+        // check if dependent implements Deleter to record that fact
+        if (!hasDeleterHolder[0] && dependent instanceof Deleter) {
+          hasDeleterHolder[0] = true;
```
scrocquesel, line 86: I would populate a `List<Deleter>` instance in parrallel, this will avoid to stream/filter/cast in Cleanup.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/ManagedDependentResourceContext.java
```diff
@@ -70,43 +68,59 @@ public Optional put(Object key, Object value) {
...
+  private DependentResource getDependentResource(String name) {
+    var dependentResource = dependentResources.get(name);
+    if (dependentResource == null) {
+      throw new OperatorException("No dependent resource found with name: " + name);
+    }
+    return dependentResource;
+  }
+
+  @SuppressWarnings("unchecked")
+  public <T extends DependentResource> T getAdaptedDependentResource(String name,
```
scrocquesel, line 116: I guess it is from cdi proxy. Why should we make a difference now that the key is the name ?

metacosm: No, this actually has nothing to do with CDI proxies, just a convenience method to retrieve the dependent resource using its actual class instead of the more generic `DependentResource` interface.

metacosm:
Still having issues with this, so switching back to draft mode as I uncovered issues that need to be fixed with the event source registration. Basically, the dependent resource name should be propagated to the associated event source if it exists.  The MySQL sample is also behaving weirdly…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/ManagedDependentResourceContext.java
scrocquesel: I don't get the diff with `public <T> DependentResource<T, ?> getDependentResource(String name, Class<T> resourceClass)`.
`getAdaptedDependentResource` can be used instead of `getDependentResource`.

metacosm: Yes, not sure which one will be kept :) `getDependentResource` is more generic so could be useful in more "automated" scenarios where knowing the actual type of the dependent resource is not practical. Basically, if you are using managed dependent resources, it kind of defeats the purpose to retrieve one implementation using its type. I see that as a failure in the API design.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/ManagedDependentResourceContext.java
```diff
@@ -70,43 +71,63 @@ public Optional put(Object key, Object value) {
...
-    if (resourceList.isEmpty()) {
-      throw new OperatorException(
-          "No dependent resource found for class: " + resourceClass.getName());
+  public void setReconcileResult(String name, ReconcileResult reconcileResult) {
+    getDependentResource(name).result = reconcileResult;
+  }
+
+  private static class DependentResourceInfo {
+    private final String name;
+    private final DependentResource dependent;
```
scrocquesel, line 116: not used. Is a public getter missing or should it be removed?
I would remove it because being able to retrieve DependentResource is not needed anymore thanks to named event source. This is fine to me as the user cannot do bad things (like calling reconcile) and we got a real managed mode.
If removed, then, I don't see why we should keep this wrapper around reconcile result. It will create one more instance per dependent/per reconcile to keep track of the reconcileResult while the hashmap could store it directly.

metacosm: Yes, I was wondering about that but I agree that we should remove it for now. It'll be easier to add it later than remove it once it's already out there…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -89,37 +90,17 @@ private ControllerConfigurationOverrider(ControllerConfiguration<R> original) {
     return this;
   }
 
-  public void replaceDependentResourceConfig(
-      Class<? extends DependentResource<?, R>> dependentResourceClass,
+  public ControllerConfigurationOverrider<R> replacingNamedDependentResourceConfig(String name,
       Object dependentResourceConfig) {
```
csviri, line 94: We don't want at least a marker interface for the dependent resource configurations?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
```diff
@@ -1,6 +1,11 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent;
 
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.EMPTY_STRING;
+
 public @interface Dependent {
 
+  @SuppressWarnings("rawtypes")
   Class<? extends DependentResource> type();
+
+  String name() default EMPTY_STRING;
```
csviri, line 10: Shouldn't have a constant `DEFAULT_NAME` ? would be more desriptive then `EMPTY_STRING`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
```diff
@@ -5,8 +5,49 @@
...
+  /**
+   * Computes a default name for the specified DependentResource class
+   *
+   * @param dependentResourceClass the DependentResource class for which we want to compute a
+   *        default name
+   * @return the default name for the specified DependentResource class
+   */
+  @SuppressWarnings("rawtypes")
+  static String defaultNameFor(Class<? extends DependentResource> dependentResourceClass) {
+    return dependentResourceClass.getCanonicalName();
```
csviri, line 51: Could not we move this code to a new constructor to `DependetResourceSpec`. Isn't this related only to managed dependent resources?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
```diff
@@ -2,26 +2,44 @@
...
 import java.util.Optional;
 
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
 public class ReconcileResult<R> {
 
   private final R resource;
   private final Operation operation;
+  private final Exception error;
```
csviri, line 12: I don't think we should have this errors, we should just trow it and initiate a retry.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -71,15 +72,22 @@ public Controller(Reconciler<P> reconciler,
...
-        .peek(d -> {
-          // check if any dependent implements Deleter to record that fact
-          if (!hasDeleterHolder[0] && d instanceof Deleter) {
-            hasDeleterHolder[0] = true;
-          }
-        })
-        .collect(Collectors.toList());
+    final var specs = configuration.getDependentResources();
+    final var size = specs.size();
+    if (size == 0) {
```
csviri, line 77: Please extract this part into a separate method, I mean the code block to initiate the dependents.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -71,15 +72,22 @@ public Controller(Reconciler<P> reconciler,
     eventSourceManager = new EventSourceManager<>(this);
 
     final var hasDeleterHolder = new boolean[] {false};
```
csviri, line 74: Why is this not a simple boolean?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
```diff
@@ -12,21 +12,15 @@
  */
 public interface EventSource extends LifecycleAware {
 
-  /**
-   * An optional name for your EventSource. This is only required if you need to register multiple
-   * EventSources for the same resource type (e.g. {@code Deployment}).
-   *
-   * @return the name associated with this EventSource
-   */
-  default String name() {
```
csviri, line 21: Why do we have separate named event source now?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -191,8 +200,46 @@ EventHandler getEventHandler() {
     return controller;
   }
 
-  private static class EventSources<R extends HasMetadata> implements Iterable<EventSource> {
-    private final ConcurrentNavigableMap<String, List<EventSource>> sources =
+  static class NamedEventSource implements EventSource {
```
csviri, line 203: Could we pls extract this class? The EventSourceManager is already to big

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
metacosm: We're using `EMPTY_STRING` elsewhere already to denote the absence of user-provided value. This is nothing new.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: This isn't managed-specific: all dependent resources need to be registered with a name so if someone doesn't want to think of one, then they need to be able to compute the default one that will be used. So I think it makes sense to have that method here.

csviri: > all dependent resources need to be registered with a name

Standalones should not. or?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
metacosm: That's an option yes but the idea here is so that a dependent that's failing doesn't prevent the other ones from doing their work. The main reconciler could then decide what to do with the dependent. Without this change, if a managed dependent is failing your main reconciler might never get called. It should be up to the reconciler to decide what to do with the failing dependent.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
csviri: Maybe then we should have `NO_VALUE`. Empty string has extremeely generic semantics.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: Because you cannot use a non-final var in a closure.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
metacosm: Because names need to be configurable from outside the code. If it's part of the `EventSource` interface then the situation could happen that the annotation says one thing but the implementation specifies a different name, which would be error-prone.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
metacosm: Why?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
csviri: But that would mean that the reconciler should always check for error for each dependent.
This does not fit with the current error handling approach either, so currently nobody does a `try-cach` on every fabric8 client action. Also some might just simply don't do the other reconciliation if the one before failed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
metacosm: Sure but it's been like this since the beginning 😅

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: Standalone also need to be registered with a name. If you don't provide one, it will be generated for you, which is why you need to know what the default name is.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
csviri: it seems now it the time with v3 :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
csviri: where is the standaline registered for example in this example:

https://github.com/java-operator-sdk/java-operator-sdk/blob/7f82f54efe90caa6d3460039eb13eec00229fb1d/sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageStandaloneDependentsReconciler.java#L33-L33

?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
metacosm: We could have it at the controller level: check if any of the dependents is in error and throw a composite error if some are.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: Hmm, that's a good point… but I think that even standalone `DependentResource` implementations should be registered so that everything could be managed the same way. Probably outside of the scope for now, though. 

The thing that is annoying is that there is a more or less one-to-one relationship with the associated `EventSource` and in the non-managed way, this isn't well handled. Typically, the `EventSource` should have the same name than the associated `DependentResource` for less confusion but this isn't the case for standalone situations…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
metacosm: Should be done in a separate PR, though, imo.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: Moving it to a different methods would prevent making the `dependents` field `final` and to be honest, I'm not sure what purpose introducing another method would serve.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
metacosm: Also, it could be argued that it's a good practice for reconcilers to check the status of each dependent, which is the whole point of `ReconcileResult`.

csviri: But this contradicts with whole current concept of `ErrorStatusHanlder`, that we have a one cross cutting place to handle errors. It simply does not fit in the picture. (Not speaking that the samples are not migrated to this in the PR). 
So also goes agains the concept that dependent resources seamlesly fit into the current concepts (well in this case the current error handling approach.

(And how to force this for standalone dependent resources? )

So I would not introduce such conceptual change, with so many open questions before release. We can discuss this further in depth later, how to handle this.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
csviri: https://github.com/java-operator-sdk/java-operator-sdk/issues/1090

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
csviri: Well don't have strong opinion on this, but for now just for forward compatibility?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
csviri: can we use `AtomicBoolean` ?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
csviri: Readability, I would also reintroduc DependentResourceManager, since there is now more considerable logic.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -179,26 +187,43 @@ public String successTypeName(UpdateControl<P> result) {
...
-            dependents.forEach(dependent -> dependent.reconcile(resource, context));
+            final var result = dependents.entrySet().stream()
+                .map(entry -> {
+                  final var dependent = entry.getValue();
+                  final var name = entry.getKey();
+                  ReconcileResult reconcileResult;
+                  try {
+                    reconcileResult = dependent.reconcile(resource, context);
+                  } catch (Exception e) {
+                    reconcileResult = ReconcileResult.error(resource, e);
```
csviri, line 198: This logic should be inside of the DependetResource (not in the controller), so then it's usable in standalone mode consistenly. But again I'm strongly agains  to indoructe this is this release.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
csviri: The dependent resource - event source. Does not need to be one to one relationship. In standalone mode we can share event source now between more dependent resources. Like if someone wants to create more config maps, does not need more event source just one.
These optimizations could be done just in standalone mode. And that is right. Since all the minor optimizations could be done just there.

csviri: But by default should have the same name I guess.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/ReconcileResult.java
csviri: > Also, it could be argued that it's a good practice for reconcilers to check the status of each dependent, which is the whole point of ReconcileResult.

It was not the point until now :) since there was not even an API, you could check if it was update created or none, but error handling is completely different story.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
csviri: But also fine to have it like this, if you think is right.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
```diff
@@ -5,8 +5,49 @@
...
+   *         doesn't exist
+   */
   Optional<R> getResource(P primaryResource);
+
+  /**
+   * Retrieves the resource type associated with this DependentResource
+   *
+   * @return the resource type associated with this DependentResource
+   */
+  Class<R> resourceType();
```
csviri, line 40: Why we need this at this level?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -179,26 +187,43 @@ public String successTypeName(UpdateControl<P> result) {
...
-        .filter(dependentResource -> dependentResource instanceof EventSourceProvider)
-        .map(EventSourceProvider.class::cast)
-        .map(provider -> provider.initEventSource(context))
-        .forEach(sources::add);
+  public void initAndRegisterEventSources(EventSourceContext<P> context) {
+    dependents.entrySet().stream()
+        .filter(drEntry -> drEntry.getValue() instanceof EventSourceProvider)
+        .forEach(drEntry -> {
+          final var provider = (EventSourceProvider) drEntry.getValue();
+          final var source = provider.initEventSource(context);
```
csviri, line 217: Could we use here the name of the dependent resource to register the event source?

csviri: That would solve the same naming problem.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: To make it usable consistently for all dependent resources, it needs to be in the `Controller` because we don't control the `DependentResource` implementations. We shouldn't think that people will always use one of the classes in the hierarchy, they might implement `DependentResource` directly.

csviri: There is not way to force correct implementation of an interace in java, but this is already inconsistent this way, since we would want to have in managed and standalone the same behavior so it can be easily migrated (but also just easy to understand) between the two appraoches.

csviri: But again this is conceptually does not fit in the current model.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
csviri: oki, makes sense!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: This is exactly what's done on the next line or did I misunderstood your comment?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: To make things explicit so that we don't need to rely on a default implementation that will not work in all situations.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
csviri: Sorry, did not see that, my fault. Great!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
csviri: I mean are we sure that this will be needed in all cases? 
I mean I can imaging a dependent resource implementation that does not need resource type.

I thought it is something that is required for quarkus.

I guess we can then delete `ResourceTypeAware`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
metacosm: That should be the case.

metacosm: And actually, if you share event sources between dependents then they cannot all have the same name, obviously.

metacosm:
Can we merge this?

csviri (approved):
sure, LGTM!

PR #1059, opened 2022-03-21
Metrics

scrocquesel:
- `initContextIfNeeded` should be part of total time execution.
- some optimization to avoid resolving metrics instance for each call. It should be a singleton, so I guess it is safe to keep a reference on it. the instance can be fetch from `configuration.getConfigurationService()` but a lot of unit tests are stubbing `ConfigurationServiceProvider.instance()` instead.


csviri (approved):
LGTM

metacosm (approved):
LGTM

PR #1061, opened 2022-03-21
feat(junit): allow to forward port of local pods

### LINKED ISSUE #1060: JUnit Operator extension can port forward local pods
Following discussion on #1047, port forwarding should be moved to JUnit extension.

scrocquesel:
Closes #1060

### DIFF
--- a/sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorE2E.java
+++ b/sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorE2E.java
@@ -16,7 +16,6 @@ import io.fabric8.kubernetes.api.model.NamespaceBuilder;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
-import io.fabric8.kubernetes.client.LocalPortForward;
 import io.javaoperatorsdk.operator.junit.AbstractOperatorExtension;
 import io.javaoperatorsdk.operator.junit.E2EOperatorExtension;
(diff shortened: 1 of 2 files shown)


sample-operators/mysql-schema/src/test/java/io/javaoperatorsdk/operator/sample/MySQLSchemaOperatorE2E.java
```diff
@@ -70,9 +70,10 @@ boolean isLocal() {
                     c.replaceDependentResourceConfig(
                         SchemaDependentResource.class,
                         new ResourcePollerConfig(
-                            700, new MySQLDbConfig("127.0.0.1", "3306", "root", "password")));
+                            700, new MySQLDbConfig("127.0.0.1", "3307", "root", "password")));
```
csviri, line 73: Why did you change this to 3307?

scrocquesel: Sometime 3306 is already taken, but I don't know if it's because i'm working on WSL with vscode and there is an automatic port forwarding. What I would really want is to have a random port generated by the operator extension and have it injected in the conf (like it is often done with wiremock)

csviri: That is the default mysql port, so usually should not be taken. Otherwise this PR looks good to me.

csviri (approved):
LGTM

csviri:
I think this would be good to have, escpecially since it creates the port forwards before the operator starts. What can already help with some error in logs.

metacosm (approved):
LGTM, can confirm that port forwarding now works in local mode (while without this, the MySQL operator test would fail because of it).

metacosm:
Merging this PR because I need this to debug things locally. We can address the port issue at some later point.

PR #1064, opened 2022-03-22
docs: dependent resources and other improvements

csviri:


docs/documentation/dependent-resources.md
```diff
@@ -111,8 +111,43 @@ JOSDK will take the appropriate steps to wire everything together and call your
...
+    labelSelector = SELECTOR,
+    dependents = {
+        @Dependent(type = ConfigMapDependentResource.class),
+        @Dependent(type = DeploymentDependentResource.class),
+        @Dependent(type = ServiceDependentResource.class)
+    })
+public class WebPageManagedDependentsReconciler
+    implements Reconciler<WebPage>, ErrorStatusHandler<WebPage> {
+
+    // emitted code
```
metacosm, line 128:
```suggestion
    // omitted code
```

docs/documentation/dependent-resources.md
```diff
@@ -38,187 +36,203 @@ match -- No --> Update --> Done
...
-      .setLabelSelector(DEPENDENT_RESOURCE_LABEL_SELECTOR));
-
-    this.serviceDR = new ServiceDependentResource();
-    serviceDR.setKubernetesClient(client);
-    serviceDR.configureWith(new KubernetesDependentResourceConfig()
-      .setLabelSelector(DEPENDENT_RESOURCE_LABEL_SELECTOR));
-  }
-  
-  // omitted code
+        implements Reconciler<WebPage>, ErrorStatusHandler<WebPage>, EventSourceInitializer<WebPage> {
```
metacosm, line 197: Isn't there a way to include code directly from the repo instead of having to copy it here?

csviri: Good, point will check!

csviri: https://emgithub.com/ 
can do the trick, the problem is that it is not very well formatted (while it is on that page directly), it is not in the resulting jekyll page.
Also I would use permalinks, since any reformat / change can mess this up if we want to use it. Since we reference the code snippet based on line numbers.

csviri: image

csviri: So it messes up the formatting, I was not able to fix it for now. What do you think should we use this? @metacosm

csviri: Also we cannot ommit code from middle.

metacosm: Maybe just pointing to the code is enough?

csviri: That I already added, see "See the full source code sample here" parts.

csviri: Ahh you mean only that, I think that is very hard to read.

metacosm: I'd prefer if we could embed the code but if we can't do it nicely, then copying the code in the doc is more problematic because then it needs to be maintained and risk being outdated, causing user confusion.

csviri: I would also prefer that, it's just not formatted now :(
And was not able to fix it for now.

I would merge this as it is, to be consistent with other parts of the docs, and we can migrate / take a look on the embedding & formatting isses seratelly, pretty sure that can be fixed somehow.

csviri: And will migrate all the docs to that in a separate PR.

metacosm (approved):
LGTM

PR #1076, opened 2022-03-23
fix: config needs to be reset for each dependent

### LINKED ISSUE #1075: Dependents can currently share configurations
The logic assigning a configuration to a dependent resource spec is wrong because it doesn't properly reset the assigned object for each dependent.

metacosm:
Fixes #1075


csviri (approved):
LGTM

PR #1080, opened 2022-03-24
fix architecture-and-internals document

JaeguKim:
As I read `architecture and internals` document, I think I found some typo.


csviri (approved):
LGTM!

thank you!

PR #1081, opened 2022-03-24
fix: mysql sample delete part

csviri:


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
```diff
@@ -32,9 +33,9 @@ public MySQLSchemaReconciler() {}
     Secret secret = context.getSecondaryResource(Secret.class).orElseThrow();
     SchemaDependentResource schemaDependentResource = context.managedDependentResourceContext()
         .getDependentResource(SchemaDependentResource.class);
-    return schemaDependentResource.fetchResource(schema).map(s -> {
+    return schemaDependentResource.getResource(schema).map(s -> {
```
metacosm, line 36: Why the change here?

metacosm (approved):
LGTM

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java
csviri: It's unrelated to the problem. I think we should rather get the actual state from cache, and handle the details when to call the actual service to the dependent resource implementation.

PR #1084, opened 2022-03-24
feat: by default standalone dependent resource inherits namespace

csviri:


metacosm (approved):
LGTM, weird I thought that had been implemented already.

PR #1094, opened 2022-03-27
feat: add a DependentResource implementation that's also an EventSource

metacosm:
The benefit of this change is multi-fold:
- simplifies AbstractDependentResource which doesn't have to handle
  cases it should not be concerned about
- allows to share more code with AbstractSimpleDependentResource
- enforces that the same name is used for the DependentResource and
  EventSource even in the standalone mode when using one of the provided
  implementations

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
@@ -14,7 +14,9 @@ public interface ResourceOwner<R, P extends HasMetadata> {
   Class<R> resourceType();
 
   /**
-   * Retrieves the resource associated with the specified primary one
+   * Retrieves the resource associated with the specified primary one, returning the actual state of
+   * the resource. Typically, this state might come from a local cache, updated after
+   * reconciliation.
    *
    * @param primary the primary resource for which we want to retrieve the secondary resource
    * @return an {@link Optional} containing the secondary resource or {@link Optional#empty()} if it
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
@@ -1,0 +1,24 @@
+package io.javaoperatorsdk.operator.processing;
+
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+public interface ResourceOwner<R, P extends HasMetadata> {
+
+  /**
+   * Retrieves the resource type associated with this ResourceOwner
+   *
+   * @return the resource type associated with this ResourceOwner
+   */
+  Class<R> resourceType();
+
+  /**
+   * Retrieves the resource associated with the specified primary one
+   *
+   * @param primary the primary resource for which we want to retrieve the secondary resource
+   * @return an {@link Optional} containing the secondary resource or {@link Optional#empty()} if it
+   *         doesn't exist
+   */
+  Optional<R> getAssociatedResource(P primary);
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -71,18 +71,34 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
   protected R handleCreate(R desired, P primary, Context<P> context) {
     ResourceID resourceID = ResourceID.fromResource(primary);
     R created = creator.create(desired, primary, context);
-    cacheAfterCreate(resourceID, created);
+    processPostCreate(resourceID, created);
     return created;
   }
 
-  protected abstract void cacheAfterCreate(ResourceID resourceID, R created);
+  /**
+   * Allows sub-classes to perform additional processing (e.g. caching) on the created resource if
+   * needed.
+   *
+   * @param primaryResourceId the {@link ResourceID} of the primary resource associated with the
+   *        newly created resource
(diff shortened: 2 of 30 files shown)


csviri:
Generally speaking, I'm not sure that `AbstractDependetResource` should be aware of caches. 
Consider a case when external resources are managed: but triggers comes from an event queue, like SQS, where the message is not resource oriented. This is typically what happens (and desired also from operators view) for example with AWS resources, see for example subscription on RDS events message:

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-cloudwatch-events.sample.html

In this case the event would trigger the reconciliation, but the dependent resource would not have an event source, especially not a caching event source.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/AbstractCachingDependentResource.java
```diff
@@ -3,29 +3,20 @@
...
+    extends
+    AbstractEventSourceHolderDependentResource<R, P, ExternalResourceCachingEventSource<R, P>> {
   private final Class<R> resourceType;
 
   protected AbstractCachingDependentResource(Class<R> resourceType) {
     this.resourceType = resourceType;
   }
 
   public Optional<R> fetchResource(P primaryResource) {
-    return eventSource.getAssociated(primaryResource);
```
csviri, line 23: I know that you did not changed this, but probably the featch resource should be abstract. The idea is that it does not read from the cache.

csviri (changes requested):
See the comments above.
It would be great if we could have this, just the `AbstractEventSource`, would not be related to caching at all.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/AbstractCachingDependentResource.java
metacosm: Can we do this in a separate PR?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
```diff
@@ -0,0 +1,26 @@
+package io.javaoperatorsdk.operator.processing;
+
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+
+public interface ResourceOwner<R, P extends HasMetadata> {
```
csviri, line 7: Why was this extracted? I don't think it should be, what is an owner? 
basically this is reader, but that should always be, a DependentResource should always read the object or tell if the object does not exists

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -66,91 +59,47 @@ public AbstractDependentResource() {
...
-  }
+  /**
+   * Allows sub-classes to perform additional processing (e.g. caching) on the created resource if
+   * needed.
+   *
+   * @param primaryResourceId the {@link ResourceID} of the primary resource associated with the
+   *        newly created resource
+   * @param created the newly created resource
+   */
+  protected abstract void processPostCreate(ResourceID primaryResourceId, R created);
```
csviri, line 86: would probably rename to `postProcessCreate`.

csviri (changes requested):

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageStandaloneDependentsReconciler.java
```diff
@@ -37,56 +37,50 @@
   private static final Logger log =
       LoggerFactory.getLogger(WebPageStandaloneDependentsReconciler.class);
 
-  private KubernetesDependentResource<ConfigMap, WebPage> configMapDR;
```
csviri, line 40: I think this is an unnecessary simplification I actually plan to extend this sample with some `if` or other workflow enahncements, that is not possible with managed dependent resources.

Basically that is the point that we can hold a simple reference to the dependent resource, that can be easily accessed. Putting it into a map and accessing it via String, is just unnecessary complicaiton. 

See also other PR pl with defaul naming.

metacosm: Sure but the current example only shows code duplication which isn't really something we want people to replicate.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceOwner.java
csviri: Sorry, I misread it, for some reason I though its implemented by `AbstractDependentResource` not by dependent resource. ( I probably should close it for today :) )

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageStandaloneDependentsReconciler.java
metacosm: And calling reconcile on each dependent and create the dependents each exactly the same way… Basically, as it stands, this sample isn't really interesting (apart from showing that people should probably use managed dependents for this kind of use case 😄) so I'd just leave it as it is here and then we can make more changes to make it more interesting in other PRs.

csviri: Yes, right, will come up with something to improve it.

Btw otherwise look good to me. Happy to merge! Pls do. ( Just don't know how to approve from mobile app :) )

csviri (commented):
LGTM (found it :) )

csviri (approved):
LGTM

PR #1098, opened 2022-03-28
Utlity naming for Event Sources

csviri:


csviri:
The failed tests seems to be a glitch because first I mistakenly pushed it to the main branch.

jesperpedersen:
Wouldn't this be

```
Map<String, Collection<EventSource>>
```

instead ? Or are you assuming that you only have one `EventSource` per area ?

csviri:
> Wouldn't this be
> 
> ```
> Map<String, Collection<EventSource>>
> ```
> 
> instead ? Or are you assuming that you only have one `EventSource` per area ?

It's basically a name of an EventSource. To address it from different places.

metacosm (approved):

PR #1100, opened 2022-03-29
fix: rename getAssiciatedResource to getSecondaryResource

csviri:
`getResource` is either called on event source or dependent resource. In both context it should be obivius from `getResource` that it is about the resource dependent resource / or event source is managing / caching.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
@@ -32,7 +32,7 @@ public class DefaultContext<P extends HasMetadata> implements Context<P> {
   public <T> Optional<T> getSecondaryResource(Class<T> expectedType, String eventSourceName) {
     return controller.getEventSourceManager()
         .getResourceEventSourceFor(expectedType, eventSourceName)
-        .flatMap(es -> es.getResource(primaryResource));
+        .flatMap(es -> es.getSecondaryResource(primaryResource));
   }
 
   @Override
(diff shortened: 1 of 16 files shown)


metacosm:
I think that `getAssociatedResource` or maybe even `getAssociatedWith` is more explicit because if you look at code like https://github.com/java-operator-sdk/java-operator-sdk/pull/1100/files#diff-4a699996771e27e06c424cfbd90ae19efb23d49ff489240c9e5a6375180a3ec9R31 it's easy to get confused about what you're trying to retrieve. Ideally, there wouldn't be any possible confusion and the method name should make that explicit so that you don't need to look at the object on which the method is called to figure out the intent of the method.

csviri:
This will be the primary API that everybody will use if any, should be short, and it is pretty obvious I think mainly for` `DependentResource.getResource()` that is about the dependent resource.  Actually for me the associated was confusing. Mainly we don't use associated anywhere else in domain language, we use `secondary`. Even the mapping functions were renamed because of this.

metacosm:
Associated is quite explicit, in my opinion: it's the resource associated with the one we specify as a parameter. 

We could rename it something different but again, ideally, in a good API, one should not have to look at the type of an object to resolve any ambiguity on what a method does. `getResource` doesn't achieve that goal, in my opinion, because the resource that gets retrieved is not the one that is pointed to by the specified parameter. It doesn't need to be `associated` but some qualifier to make things more explicit would be better than just `getResource`.

csviri:
> Associated is quite explicit, in my opinion: it's the resource associated with the one we specify as a parameter.
> 
> We could rename it something different but again, ideally, in a good API, one should not have to look at the type of an object to resolve any ambiguity on what a method does. `getResource` doesn't achieve that goal, in my opinion, because the resource that gets retrieved is not the one that is pointed to by the specified parameter. It doesn't need to be `associated` but some qualifier to make things more explicit would be better than just `getResource`.

I would then name is as `getSecondaryResource` , this will match the current glossary, so we will have always:
- primary resource 
- secondary resource
- dependent resource feature 

So it will be then straightforward that we get secondary resource (ideally) either from event source or (if the feature is used) dependent resource.
What do you think?

metacosm (approved):
Thank you!

PR #1103, opened 2022-03-29
fix(mysql-schema): use proper secondary resource type

scrocquesel:
I guess branch next has been merged into main.

mysql-schema sample was retrieving a secondary resource by the primary type. I think it should have fail but it doesn't not.

That is `context.getSecondaryResource(Primary.class)` actually returns something.  Shoud I raise an issue  ?


csviri:
This is very strnage let me investigate it today first, thx for pointing it out!

csviri (approved):
LGTM

csviri:
This works because it finds the MySQLSchema based on type not by name. I think it should fail if the name set but it's not realistic name.

csviri:
see: https://github.com/java-operator-sdk/java-operator-sdk/pull/1104

metacosm (approved):

PR #1104, opened 2022-03-30
fix: event sources no ordered start & missing event source throws exception

csviri:
- Event sources not need to start in particular order this is handled by the processor - it starts after event sources started
- If asked for a secondary resource from the context for a event source using the name, throws exception if event source not present with that name

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
@@ -21,6 +21,7 @@ public interface EventSource extends LifecycleAware {
   void setEventHandler(EventHandler handler);
 
   static String defaultNameFor(EventSource source) {
-    return source.getClass().getCanonicalName();
+    // we can have multiple event sources for the same class
+    return source.getClass().getName() + "#" + source.hashCode();
   }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
@@ -20,8 +20,5 @@ public interface EventSource extends LifecycleAware {
    */
   void setEventHandler(EventHandler handler);
 
-  static String defaultNameFor(EventSource source) {
-    // we can have multiple event sources for the same class
-    return source.getClass().getName() + "#" + source.hashCode();
-  }
+
 }
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventSourceManagerTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventSourceManagerTest.java
@@ -80,8 +80,10 @@ class EventSourceManagerTest {
     // manager is initialized with a controller configured to handle HasMetadata
     EventSourceManager manager = initManager();
     EventSource source = manager.getResourceEventSourceFor(HasMetadata.class);
-
     assertThat(source).isInstanceOf(ControllerResourceEventSource.class);
+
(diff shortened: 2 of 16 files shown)


metacosm:
> * Event sources not need to start in particular order this is handled by the processor - it starts after event sources started

Are you sure? This used to be a problem…

> * If asked for a secondary resource from the context for a event source using the name, throws exception if event source not present with that name

👍🏼

csviri:
> Are you sure? This used to be a problem…

Yep, the fix in event processor happened after, so now the event processor start after each event source started. So no event processed before all of the started. (The issue before was that the ControllerEventSource did not start before than others, so the cache was not populated soon enough)

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventSourceManagerTest.java
```diff
@@ -76,22 +74,24 @@ public void startCascadesToEventSources() {
...
     // manager is initialized with a controller configured to handle HasMetadata
     EventSourceManager manager = initManager();
-    Optional<EventSource> source = manager.getResourceEventSourceFor(HasMetadata.class);
-    assertTrue(source.isPresent());
-    assertTrue(source.get() instanceof ControllerResourceEventSource);
+    EventSource source = manager.getResourceEventSourceFor(HasMetadata.class);
+    assertThat(source).isInstanceOf(ControllerResourceEventSource.class);
+
+    assertThatExceptionOfType(IllegalArgumentException.class)
+        .isThrownBy(() -> manager.getResourceEventSourceFor(HasMetadata.class, "not_known_name"));
```
metacosm, line 86:
```suggestion
        .isThrownBy(() -> manager.getResourceEventSourceFor(HasMetadata.class, "unknown_name"));
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
```diff
@@ -21,6 +21,7 @@
   void setEventHandler(EventHandler handler);
 
   static String defaultNameFor(EventSource source) {
-    return source.getClass().getCanonicalName();
+    // we can have multiple event sources for the same class
+    return source.getClass().getName() + "#" + source.hashCode();
```
metacosm, line 25: 🤔 I'm not sure how this would work… Why do we need a name if the generated names are unique already? Also, this only works if the user already has access to the `EventSource` instance, which kind of defeats the purpose of being able to retrieve it by its name… so something doesn't really work here.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
csviri: The previous implementation simply does not work when there are multiple instances with the same type. Those would get the same name (there was already a bug with that) so basically not addressable. The default name is only for the case when the user does not want to explicitly address the event source. 
So either:
- there is just 1 event source by type, so the resource is accessed via the context and type
- the cached resource addressed via a dependent resource

On every other cases it's just needs an explicit name, so the generated name is only basically a workaround to user does not need to think about a name is it's not used anywhere. It's correct in this sense.

csviri: Note that InformerEventSoure for a config map and a deployment for example still has the same class.

csviri: What could ne done is just move this to EventSourceInitializer method and make it more explicit that is generated. Would make ti a separate PR.

csviri: ok pushed a small refactor regarding to this

metacosm (approved):

Issue #1105, opened 2022-03-30
AbstractDependentResource should be able to pass his turn


scrocquesel:
Actually, when the secret is not yet created, the mysql schema sample is throwing some errors in SchemaDependentResource due to `orElseThrow`.

https://github.com/java-operator-sdk/java-operator-sdk/blob/2eac098db872cedcefecb6036ec98c77b58cf60f/sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java#L56

The same apply with the reconciler 

https://github.com/java-operator-sdk/java-operator-sdk/blob/2eac098db872cedcefecb6036ec98c77b58cf60f/sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/MySQLSchemaReconciler.java#L39

I'm trying to avoid these exceptions because there are not really exceptions, but just the fact that some requirement are not yet satisfied. So I tried,

```
@Override
    public Schema desired(MySQLSchema primary, Context<MySQLSchema> context) {
        return context.getSecondaryResource(Secret.class)
                .map(secret -> new Schema(primary.getMetadata().getName(), primary.getSpec().getEncoding()))
                .orElse(null);
    }
```

But `AbstractDependentResource.logForOperation` produces an NPE and even if it doesn't, I think that `create` should not be call.

It is not clear what to do in such a case. I don't feel like exception are the way to go because it raise a call to `updateErrorStatus` when it should definitely not. Should desired return an `Optional<R>` ?

metacosm:
I wanted to have an `ERROR` case in `ReconcileResult` but we decided that this was too much of a behavioral change at this point. That said, I agree with you, right now, we're using exceptions for behavior / control and it's a bad design, imo.

csviri:
How can it happen that the secret is not created (I mean other than an error state?). The secret is a managed dependent resource that is created before the schema. Managed dependent resources are created in order.

It's true that lately the behavior changed, that we reconcile all the resources event if the previous ones failed. This makes issues when the first operation fails, but only if the resources are not independent.

csviri:
> I wanted to have an `ERROR` case in `ReconcileResult` but we decided that this was too much of a behavioral change at this point. That said, I agree with you, right now, we're using exceptions for behavior / control and it's a bad design, imo.

The `ERROR` does not fit in the big picture now, could you create a separate issue for that to discuss the details?

[...]

PR #1112, opened 2022-03-31
fix: fix NPE when desired state is null

scrocquesel:
This could fix https://github.com/java-operator-sdk/java-operator-sdk/issues/1105 without too much modification while waiting for a better design around dependent dependencies.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -33,9 +33,11 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
       if (maybeActual.isEmpty()) {
         if (creatable) {
           var desired = desired(primary, context);
-          logForOperation("Creating", primary, desired);
-          var createdResource = handleCreate(desired, primary, context);
-          return ReconcileResult.resourceCreated(createdResource);
+          if (desired != null) {
+            logForOperation("Creating", primary, desired);
+            var createdResource = handleCreate(desired, primary, context);
+            return ReconcileResult.resourceCreated(createdResource);
+          }
         }
       } else {
         final var actual = maybeActual.get();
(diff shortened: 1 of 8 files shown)


scrocquesel:
I add a way to wait for database to be ready so that the unit test can ensure the primary never got errored properly.

If you revert the first commit (5dda86785fbf5f143d5bcadabf0bf8f05b8a3d83), you should see the test fail.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -33,9 +33,11 @@ public AbstractDependentResource() {
       if (maybeActual.isEmpty()) {
         if (creatable) {
           var desired = desired(primary, context);
-          logForOperation("Creating", primary, desired);
-          var createdResource = handleCreate(desired, primary, context);
-          return ReconcileResult.resourceCreated(createdResource);
+          if (desired != null) {
```
csviri, line 36: TBH, my opinion is we should not try to solve this case, we should simply throw exception in the desired as it is now. We are hacking the situation what should not happen

scrocquesel: > We are hacking the situation what should not happen

I agree that in the case of this sample, this is pretty unlikely, but this is definitely a situation that can happen for others inter-dependent use case.
I understand this is not yet supported by dependent for now and if required, it could be handle with a standalone implementation that would catch the exception for eg or throw a a specialized exception to be handle in the errorHandler so it can Update the primary status to say that the first dependent is ok but not the second one. Would be a bit tricky but can de done.

Should it be described in the doc that throwing an exception is a valid "return" for desired and create ?

metacosm: Relying on exceptions for flow control is a bad pattern but that's what we need to do until we implement better flow control of dependents.

scrocquesel: > Relying on exceptions for flow control is a bad pattern but that's what we need to do until we implement better flow control of dependents.

agreed. closing this

csviri: @metacosm @scrocquesel regarding the exception I created a separate issue to discuss, I also added my concerns regarding to enum extension. But more then happy to discuss it in depth under this issue: https://github.com/java-operator-sdk/java-operator-sdk/issues/1126

PR #1117, opened 2022-04-01
feat: prepareEventSources now specifies ordering of EventSources

metacosm:


csviri:
hmm, but this is not the problem, its about the dependents in the `@Dependent`  annotation. Those needs to be in order at the end in the controller, or?

metacosm:
> hmm, but this is not the problem, its about the dependents in the `@Dependent` annotation. Those needs to be in order at the end in the controller, or?

Yes, indeed, that also needs to be fixed, indeed, but it also needs to be made explicit in `EventSourceInitializer`.

csviri:
> > hmm, but this is not the problem, its about the dependents in the `@Dependent` annotation. Those needs to be in order at the end in the controller, or?
> 
> Yes, indeed, that also needs to be fixed, indeed, but it also needs to be made explicit in `EventSourceInitializer`.

Why? What is the problem it solves?

csviri:
The order of event sources should not matter, it is basically started all together, and the processor is started later. There is an explicit code than enables any order of event source starting.

PR #1119, opened 2022-04-01
fix: order dependents

metacosm:


csviri:
Can we add an intergation test? to check if the resource are reconciled in order?

PR #1120, opened 2022-04-01
fix: ordered managed dependents

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -22,7 +22,7 @@
   private ResourceEventFilter<R> customResourcePredicate;
   private final ControllerConfiguration<R> original;
   private Duration reconciliationMaxInterval;
-  private final Map<String, DependentResourceSpec<?, ?>> dependentResourceSpecs;
+  private final LinkedHashMap<String, DependentResourceSpec<?, ?>> namedDependentResourceSpecs;
```
csviri, line 25: Why use a map why not a list? as in ControllerConfiguration or DefaultControllerConfiguration: `List<DependentResourceSpec<?, ?>>`
DependentResourceSpec already contains the name. It's basically data duplication.

Would be also consistent with those 2 classes.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -79,14 +79,14 @@ public Controller(Reconciler<P> reconciler,
...
+      specs.forEach(drs -> {
         final var dependent = createAndConfigureFrom(drs, kubernetesClient);
         // check if dependent implements Deleter to record that fact
         if (!hasDeleterHolder[0] && dependent instanceof Deleter) {
           hasDeleterHolder[0] = true;
         }
-        dependentsHolder.put(name, dependent);
+        dependentsHolder.put(drs.getName(), dependent);
       });
       dependents = Collections.unmodifiableMap(dependentsHolder);
```
csviri, line 91: If we don't want to use list but an ordered Map, thus LinkedHashMap, we should have it also on the defined variable on line 57, so it's explicit on the definition.   (For some reason I cannot comment on that line)

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverriderTest.java
```diff
@@ -25,13 +25,17 @@ void replaceNamedDependentResourceConfigShouldWork() {
     var dependents = configuration.getDependentResources();
```
csviri, line 25: This test needs to be breaked down to multiple separate test cases. It's very hard to read this way.

csviri (commented):
Made some comments :)

csviri:
Ahh and it's my PR, will update :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
metacosm: Makes it easier to check if the spec is already present, find it again, naturally ensures that there is no duplication.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
metacosm: The problem if we do that is that we cannot use `Collections.emptyMap` or `Collections.unmodifiableMap` anymore… Then again, since the dependents are not exposed, it's probably OK and indeed conveys the intent more clearly.

PR #1124, opened 2022-04-01
fix: namespaces overriding should propagate to dependents if needed

### LINKED ISSUE #1122: `InformerConfiguration` doesn't properly inherit parent configuration if configuration is overridden
The `InformerConfiguration` needs to be notified of the configuration override somehow.

metacosm:
Fixes #1122 
Relies on #1120.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
@@ -10,6 +10,8 @@ import static io.javaoperatorsdk.operator.api.reconciler.Constants.NO_VALUE_SET;
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface KubernetesDependent {
+  String SAME_AS_PARENT = "JOSDK_SAME_AS_PARENT";
+  String[] DEFAULT_NAMESPACES = {SAME_AS_PARENT};
 
   /**
    * Specified which namespaces this Controller monitors for custom resources events. If no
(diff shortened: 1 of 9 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
```diff
@@ -10,6 +10,10 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface KubernetesDependent {
+  String SAME_AS_PARENT = "JOSDK_SAME_AS_PARENT";
+  String WATCH_ALL_NAMESPACES = "JOSDK_ALL_NAMESPACES";
```
csviri, line 14: This should not be in the `@ControllerConfiguration` ? 
That might be unrelated or not strictly related, but probably we should have just to watch all or selected namespaces. Not sure about the current namespace, but that should be also explicit in case.

csviri (approved):
Just made one comments, otherwise LGTM

csviri:
The mysql schema test is faling for some reason though.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
```diff
@@ -10,6 +10,10 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface KubernetesDependent {
+  String SAME_AS_PARENT = "JOSDK_SAME_AS_PARENT";
```
csviri, line 13: How is this now used or checked anywhere?

csviri: Ahh I see, nevermind.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependent.java
metacosm: No, this needs to be on this annotation for the dependent to be able to force some configurations which would otherwise be impossible to express.

PR #1133, opened 2022-04-05
fix: Ordered Managed Dependent IT and filtering

csviri:
The problem was that there were two informers without label selector for the 2 config maps, so event filtering was not working properly. There are 2 solutions in this case, share the event source between the dependents or use label selectors.


metacosm:
> The problem was that there were two informers without label selector for the 2 config maps, so event filtering was not working properly. There are 2 solutions in this case, share the event source between the dependents or use label selectors.

Would it be worth explaining why this wouldn't work without the label selectors in the docs?

metacosm (approved):

PR #1137, opened 2022-04-05
fix: shut down operator when failing to start

metacosm:
- fix: start timer as daemon to avoid prolonging the app's life
- fix: exit operator on exception when starting


csviri (approved):
LGTM

andreaTP:
Thanks 🎉 !

PR #1147, opened 2022-04-07
feat: patching with update control

csviri:
Ability to patch resource and status using `UpdateControl`.


metacosm (approved):
LGTM, though it's quite dangerous to do this kind of change now.

csviri:
It's not ideal to do it now before the release for sure. On the other hand it's quite simple change. Will merge and in case of any issues in RC we can still revert.

PR #1153, opened 2022-04-11
Workflow engine implementation

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Workflow.java
```diff
@@ -8,18 +8,31 @@
 import java.util.concurrent.Executors;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.config.ConfigurationServiceProvider;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 
+/**
+ * Dependents definition: so if B depends on A, the B is dependent of A.
```
metacosm, line 15: I think this is confusing since it's the reverse of what we use for `DependentResource`: the primary resource depends on the dependents…

csviri: Yep, I think the naming needs to be improved here in general. So this is a directed graph at the end, if reverse direction we could use simply `parents`. Not sure if `dependents` is actually bad, but yes can be confused with dependent resource from other context, maybe just use `child`?

metacosm (approved):
I think this is a good start we can merge.
Some comments: 
- I think the names being used are too confusing… I especially got confused by the dependents field in Workflow initially (and actually got rid of it in part because of this ^_^)
- We should have the code fully reactive and only wait for all resources to be processed at the top level
- Should we use a DAG library? We could probably find one that would handle our use case without needing to maintain our own (especially if we factor in other things like cycle detection and reduction of the graph)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileExecutor.java
```diff
@@ -0,0 +1,293 @@
...
+    this.context = context;
+    this.workflow = workflow;
+  }
+
+  public synchronized WorkflowExecutionResult reconcile() {
+    for (DependentResourceNode dependentResourceNode : workflow
+        .getTopLevelDependentResources()) {
+      handleReconcile(dependentResourceNode);
+    }
+    while (true) {
```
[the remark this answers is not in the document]
csviri: Would be. First I tried with reactive, but this has quite dynamic nature, while with reactive (`CompletableFuture` etc) is meant to for static workflows. Pretty sure it would be possible somehow just probably not skilled enough in this.

csviri:
> Should we use a DAG library? We could probably find one that would handle our use case without needing to maintain our own (especially if we factor in other things like cycle detection and reduction of the graph)

Well we can take a look, but there are parts when the dag is actually traversed back and forth, for example on reconcile delete is made in reverse order on a sub-tree. 
For cycle detection we can use a library, and maybe just convert the dag (actually set of dags, but that could be tranformed to single dag adding one parent to all dags) to the library's model just for cycle detection. (if that is hard to implement from scratch)

PR #1155, opened 2022-04-12
fix: make webpage sample work OOTB

rsynek:
The webpage sample does not work OOTB with the `k8s/webpage.yaml`, as the default reconciler watches for webpage resources without labels.


csviri (approved):
LGTM

metacosm (approved):

PR #1169, opened 2022-04-21
All Resource Event Sources can handle multiple seconday resources for a primary resources

csviri:


csviri:
@metacosm will review some additional parts, and javadoc. But conceptually can be reviewed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/IDMapper.java
```diff
@@ -0,0 +1,18 @@
+package io.javaoperatorsdk.operator.processing.event.source;
+
+import java.util.function.Function;
+
+public interface IDMapper<R> extends Function<R, String> {
```
metacosm, line 5: I'd rather have an explicit interface with an actual method which name we control rather than a Function extension. Easier to extend at a later time and more self-documenting than a random `apply` method in implementors, imo.
I'm also unsure about the purpose of this, actually. :(

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/IDMapper.java
csviri: yep, will make it more explicit, not extending the function. Tried to explain here, why it is needed:

https://github.com/java-operator-sdk/java-operator-sdk/blob/44987c93543814e7c3b8d5f1246adb7a8ab47f5e/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ExternalResourceCachingEventSource.java#L15-L32

csviri: pushed. 

Can try to explain more in depth here in the comment why this needed. But basically we need to have the information if it's a new resource is a new version or just the same as an other. With key check if it referring to same object with equals we check if it changed or not.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ExternalResourceCachingEventSource.java
```diff
@@ -0,0 +1,165 @@
...
+import io.javaoperatorsdk.operator.processing.event.Event;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+/**
+ * Handles caching and related operation of external event sources. It can handle multiple secondary
+ * resources for a single primary resources.
+ * <p>
+ * There are two related concepts to understand:
+ * <ul>
+ * <li>IDMapper - maps or in other words extracts a id from the resources</li>
```
metacosm, line 21:
```suggestion
 * <li>CacheKeyMapper - maps/extracts a key used to reference the associated resource in the cache</li>
```

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaPollingResourceFetcher.java
```diff
@@ -18,7 +18,7 @@ public SchemaPollingResourceFetcher(MySQLDbConfig mySQLDbConfig) {
   }
 
   @Override
-  public Optional<Schema> fetchResource(MySQLSchema resource) {
-    return schemaService.getSchema(resource.getMetadata().getName());
+  public Set<Schema> fetchResources(MySQLSchema primaryResource) {
+    return null;
```
metacosm, line 22: Is this normal? If yes, can you add a comment explaining why, please?

csviri: its not used delete.

csviri: good catch! :)

PR #1180, opened 2022-04-25
feat: throw exception if desired is null

csviri:


metacosm:
I'd don't think that this is enough. The issue occurred when `handleCreate` returned `null`. If you look at the code for the MySQL sample in the Quarkus extension you'll see that it's actually `create` that returns `null`.

csviri:
> I'd don't think that this is enough. The issue occurred when `handleCreate` returned `null`. If you look at the code for the MySQL sample in the Quarkus extension you'll see that it's actually `create` that returns `null`.

good point will add check there too

csviri:
updated.

metacosm (approved):
LGTM

PR #1184, opened 2022-04-27
fix: simplify informer config generics

csviri:


csviri:
@metacosm will merge this since other issue depends on it. We can review after an do patches in case you have some comments.

PR #1187, opened 2022-04-29
dynamic namespace config

csviri:


csviri:
@metacosm tomorrow will add tests, but this is the idea, does not look too scary for the moment.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/RegisteredController.java
```diff
@@ -0,0 +1,22 @@
+package io.javaoperatorsdk.operator.api;
+
+import java.util.Set;
+
+import io.javaoperatorsdk.operator.processing.Controller;
+
+public class RegisteredController {
```
metacosm, line 7: Why not introduce these methods directly on `Controller`?

metacosm: I guess that would expose methods we don't want to expose to users.

metacosm, line 7: Perhaps this class should be in the same package as `Operator`?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/RegisteredController.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private Controller<?> controller;
+
+  public RegisteredController(Controller<?> controller) {
```
metacosm, line 11: Putting this class in the same package as `Operator` would allow us to make this constructor package-only.

csviri: Moved to the same package. Also changed it to an interface, that is implemented by the controller. What do you think?

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -57,15 +70,61 @@ public InformerConfigurationBuilder<R> withSecondaryToPrimaryMapper(
     }
 
     public InformerConfigurationBuilder<R> withNamespaces(String... namespaces) {
-      this.namespaces = namespaces != null ? Set.of(namespaces) : Collections.emptySet();
-      return this;
+      return withNamespaces(
+          namespaces != null ? Set.of(namespaces) : ResourceConfiguration.DEFAULT_NAMESPACES);
```
csviri, line 74: This `DEFAULT_NAMESPACES` is a little confusing. Should we rename this to `WATCH_ALL_NAMESPACES_ON_CLUSTER` or such?

Since it is already defined as:
`Set<String> DEFAULT_NAMESPACES = Collections.singleton(Constants.WATCH_ALL_NAMESPACES);`

see:
https://github.com/java-operator-sdk/java-operator-sdk/blob/b9ee05e6e3c803be0b1e95633f465a632bd1dd52/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java#L13-L14

@metacosm

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/NamespaceChangeable.java
```diff
@@ -0,0 +1,26 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import java.util.Set;
+
+public interface NamespaceChangeable {
```
csviri, line 5: I was actually thinking to do such abstraction, but first I thought that anyways everbody will use `InformerEventSource`. So will add if it is asked for. But actually good that you added, have a feeling of completeness now :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
```diff
@@ -65,6 +72,31 @@ void initSources(MixedOperation<T, KubernetesResourceList<T>, Resource<T>> clien
     }
   }
 
+  public void changeNamespaces(Set<String> namespaces) {
+    if (ResourceConfiguration.allNamespacesWatched(namespaces)) {
+      throw new OperatorException("This feature is only supported for ");
```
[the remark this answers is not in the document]
csviri: Not sure how this got there, I remember removing it. Deleted now.

csviri:
Regarding this part:
https://github.com/java-operator-sdk/java-operator-sdk/blob/c60e078d0e0ca6a6550c81201d6d26effd4d225f/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java#L328-L334

I think it's fair to say that target namespaces for change should not contain these placeholders with special meaning just really a list of concrete namespace names.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -57,15 +70,61 @@ public InformerConfigurationBuilder<R> withSecondaryToPrimaryMapper(
...
+     * {@link io.javaoperatorsdk.operator.processing.Controller}, meaning that the informer will be
+     * restarted to watch the new namespaces if the parent controller's namespace configuration
+     * changes.
+     *
+     * @param context {@link EventSourceContext} from which the parent
+     *        {@link io.javaoperatorsdk.operator.processing.Controller}'s configuration is retrieved
+     * @param <P> the primary resource type associated with the parent controller
+     * @return the builder instance so that calls can be chained fluently
+     */
+    public <P extends HasMetadata> InformerConfigurationBuilder<R> withNamespacesInheritedFromController(
```
csviri, line 109: Yep, this is really god and useful IMHO.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -22,7 +22,7 @@
     protected DefaultInformerConfiguration(String labelSelector,
         Class<R> resourceClass,
         SecondaryToPrimaryMapper<R> secondaryToPrimaryMapper,
-        Set<String> namespaces, Boolean followControllerNamespaceChanges) {
+        Set<String> namespaces, boolean followControllerNamespaceChanges) {
```
metacosm, line 25: Ah yes, I saw that in my first review and then couldn't find it again! 👍🏼

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -145,8 +145,9 @@ public void broadcastOnResourceEvent(ResourceAction action, R resource, R oldRes
 
   public void changeNamespaces(Set<String> namespaces) {
     eventProcessor.stop();
-    eventSources.allEventSources()
-        .filter(es -> es instanceof NamespaceChangeable)
+    eventSources
+        .allEventSources()
+        .filter(NamespaceChangeable.class::isInstance)
```
metacosm, line 150: 👍🏼

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageReconciler.java
```diff
@@ -52,18 +52,22 @@ public WebPageReconciler(KubernetesClient kubernetesClient) {
   public Map<String, EventSource> prepareEventSources(EventSourceContext<WebPage> context) {
     var configMapEventSource =
         new InformerEventSource<>(InformerConfiguration.from(ConfigMap.class)
+            .withNamespacesInheritedFromController(context)
             .withLabelSelector(LOW_LEVEL_LABEL_KEY)
             .build(), context);
     var deploymentEventSource =
         new InformerEventSource<>(InformerConfiguration.from(Deployment.class)
+            .withNamespacesInheritedFromController(context)
```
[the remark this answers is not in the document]
csviri: Yes, I think that would be a reasonable default imo. The thing is that when I removed the that `from` before, was because it was simply not used. But will add such method.

csviri:
Wonder if we should do something like this with labels, but that is not that straightforward semantically. So would rather create a issue for that to discuss.

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageReconciler.java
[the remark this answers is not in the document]
csviri: pushed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
metacosm: I used `DEFAULT_NAMESPACES` because I thought it'd be a good idea to decouple the notion of what is used for the default value from the actual value so that it's cleared at the usage site that we're using the default value because that's what should be used at that spot instead of an arbitrary value. 

I do agree that it's a little confusing but I think it's better to keep this semantics to make it explicit that the default value is used at these spots. This way, the code using the value should also not change if the default value were to change.

csviri: Yes that makes sense, I just I think found some places where it was not used, just the value directly, will rather then create PR for that if I find it again (don't remember where it was)

metacosm (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
csviri: https://github.com/java-operator-sdk/java-operator-sdk/blob/80f631e72d7b0dbd17802435490d8944b72ef742/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AnnotationControllerConfiguration.java#L76-L76

I think here it would be good to use it too

csviri: will take a look

metacosm: Yes, that's possible, unfortunately, ran into a couple of such occurrences myself as well and tried to fix them when I could.

PR #1188, opened 2022-05-02
aggregate exception improvement

csviri:
Overriding `toString`. Adding the first exception as cause - maybe would be better to override `printStackTrace` to handle the causes? Respectively `getOurStackTrace` in `Throwable` to have it nicer?

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
@@ -1,6 +1,7 @@
 package io.javaoperatorsdk.operator;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
```diff
@@ -1,23 +1,30 @@
...
 import java.util.List;
 
 public class AggregatedOperatorException extends OperatorException {
+
   private final List<Exception> causes;
 
   public AggregatedOperatorException(String message, Exception... exceptions) {
-    super(message);
-    this.causes = exceptions != null ? Arrays.asList(exceptions) : Collections.emptyList();
+    super(message, exceptions[0]);
```
metacosm, line 11: This might NPE if no exception is passed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
```diff
@@ -1,23 +1,30 @@
...
+    this.causes = Arrays.asList(exceptions);
   }
 
   public AggregatedOperatorException(String message, List<Exception> exceptions) {
-    super(message);
-    this.causes = exceptions != null ? exceptions : Collections.emptyList();
+    super(message, exceptions.get(0));
```
metacosm, line 16: Same, we need to check that the given parameter is not null…

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
csviri: pushed with null check

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
```diff
@@ -2,18 +2,19 @@
...
 import java.util.List;
+import java.util.Objects;
 
 public class AggregatedOperatorException extends OperatorException {
 
   private final List<Exception> causes;
 
   public AggregatedOperatorException(String message, Exception... exceptions) {
-    super(message, exceptions[0]);
+    super(message, Objects.requireNonNull(exceptions)[0]);
```
metacosm, line 12: That doesn't help, actually, since this still will throw an NPE if `exceptions` isn't present. With varargs, it's a definite possibility, though, granted, creating an `AggregatedOperatorException` and not passing any exceptions is probably expected to fail…

metacosm (approved):

PR #1199, opened 2022-05-06
feat: make RegisteredController a little more useful

metacosm:
Allows to retrieve information from registered controllers from
Operator, in particular if there are any registered controllers, which
the Quarkus extension uses to determine whether or not to start the
Operator.


csviri (approved):
LGTM

PR #1203, opened 2022-05-09
feat: integration test for cross reference dependent resource

csviri:


csviri:
@metacosm first I wanted this just to reproduce the issue, but probably won't hurt to merge it to main.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/dependentresourcecrossref/DependentResourceCrossRefResource.java
```diff
@@ -0,0 +1,17 @@
...
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.Kind;
+import io.fabric8.kubernetes.model.annotation.ShortNames;
+import io.fabric8.kubernetes.model.annotation.Version;
+
+@Group("sample.javaoperatorsdk")
+@Version("v1")
+@Kind("MaxIntervalTestCustomResource")
+@ShortNames("mit")
```
metacosm, line 13:
```suggestion
```

This is wrong.

metacosm (commented):
I'm ambivalent on whether to add this test or not, actually. We could retool it for performance testing or maybe only run it in some cases like once a day or something so that we don't always run rather intensive tests.

csviri:
> I'm ambivalent on whether to add this test or not, actually. We could retool it for performance testing or maybe only run it in some cases like once a day or something so that we don't always run rather intensive tests.

Maybe we could have it less intensive. Like not 500 but just 50.

( TBH not sure if that causes lot's of overhead. Actually it runs very fast already and k8s api should handle such load without any issues. But I probably would merge it with reducted number).

metacosm (approved):
LGTM

PR #1206, opened 2022-05-10
refactor: clean namespaces constants

metacosm:
- refactor!: move SAME_AS_CONTROLLER constant to Constants class, add set
- refactor!: rename WATCH_CURRENT_NAMESPACE_SET to match other constants

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
@@ -5,12 +5,12 @@ import java.util.Set;
 
 public final class Constants {
 
-  public static final Set<String> DEFAULT_NAMESPACES =
+  public static final Set<String> DEFAULT_NAMESPACES_SET =
       Collections.singleton(Constants.WATCH_ALL_NAMESPACES);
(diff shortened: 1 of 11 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Constants.java
```diff
@@ -7,13 +7,17 @@ public final class Constants {
 
   public static final Set<String> DEFAULT_NAMESPACES =
       Collections.singleton(Constants.WATCH_ALL_NAMESPACES);
-  public static final Set<String> WATCH_CURRENT_NAMESPACE_SET =
+  public static final Set<String> WATCH_CURRENT_NAMESPACES =
```
csviri, line 10: Not sure about this, in this simple plural form it says we can watch more current namespace. While the case is that this is only a helper to wrap the WATCH_CURRENT_NAMESPACE. That is why I just added the SET suffix.

metacosm: Then let's rename all the Set constants with the `_SET` suffix

csviri: agree!

csviri (approved):
LGTM

PR #1222, opened 2022-05-18
fix: same as controller config for KubernetesDependentResource standalone

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -49,12 +51,17 @@ public KubernetesDependentResource(Class<R> resourceType) {
 
   @Override
   public void configureWith(KubernetesDependentResourceConfig config) {
-    configureWith(config.labelSelector(), config.namespaces(), !config.wereNamespacesConfigured());
+    this.kubernetesDependentResourceConfig = config;
```
metacosm, line 54: Shouldn't we call `configureWith(…)` here?

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
csviri: we dont, have the context here, so we don't know the controller namespaces. So it cannot be done at this time.

metacosm: So when is it done, then? The expectation with the `configureWith` methods is that once they're called, the dependent resource is set up. In this case, this isn't true anymore and someone calling this method would need to call one of the other versions at some later time…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -136,9 +143,17 @@ protected NonNamespaceOperation<R, KubernetesResourceList<R>, Resource<R>> prepa
 
   @Override
   protected InformerEventSource<R, P> createEventSource(EventSourceContext<P> context) {
-    configureWith(null, context.getControllerConfiguration().getNamespaces(), true);
-    log.warn("Using default configuration for " + resourceType().getSimpleName()
-        + " KubernetesDependentResource, call configureWith to provide configuration");
+    if (kubernetesDependentResourceConfig != null) {
```
csviri, line 146: @metacosm it's done here, when the event source is created.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
csviri: see comment below

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
metacosm: 🤔 I saw that but wasn't convinced there wasn't a path where this wouldn't get called… But I guess `createEventSource` is always called so we should be good! Thanks!

PR #1227, opened 2022-05-19
chore: update to proper version to release

metacosm:


csviri (approved):
LGTM 
(not sure if this needed  btw but lets do it to be safe)

PR #1230, opened 2022-05-20
feat: limited build to k8s >= 1.20.x

csviri:


.github/workflows/pr.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        kubernetes: [ 'v1.17.17','v1.18.20','v1.19.16','v1.20.15','v1.21.12', 'v1.22.9', 'v1.23.6', 'v1.24.0' ]
+        kubernetes: ['v1.20.15','v1.21.12', 'v1.22.9', 'v1.23.6', 'v1.24.0' ]
         exclude:
           - java: 11
             kubernetes: 'v1.18.20'
```
metacosm, line 39: You also need to remove these excludes that are not needed anymore…

csviri: ahh, yes, thx. done.

csviri: btw, its not clear to me why exactly those are excluded, what is the rational behind those values? I mean why are those exactly excluded. 
Should be there just the oldest and the newest not excluded for java 11? @metacosm

metacosm: I guess the idea was to test only the latest versions with java 11 to limit the CI intensity.

metacosm (approved):

PR #1231, opened 2022-05-23
fix: label selector on sample

csviri:


csviri:
> I have a question about this PR and labels: why does WebPageDependentReconciler has a DEPENDENT_RESOURCE_LABEL_SELECTOR which isn't used anywhere? Shouldn't be the same as WebPageManagedDependentsReconciler.SELECTOR?

Hi this was because of demo purposes, 
see comment here: https://github.com/java-operator-sdk/java-operator-sdk/blob/5afdecd3a00b2a5d1c0e6170f4c942bda3126dc8/sample-operators/webpage/k8s/webpage.yaml#L4-L4

TBH it would be better to remove it, you are right, so there is no confusion around it.
thx!

sample-operators/webpage/src/main/java/io/javaoperatorsdk/operator/sample/WebPageStandaloneDependentsReconciler.java
```diff
@@ -85,7 +87,7 @@ private void createDependentResources(KubernetesClient client) {
     Arrays.asList(configMapDR, deploymentDR, serviceDR, ingressDR).forEach(dr -> {
       dr.setKubernetesClient(client);
       dr.configureWith(new KubernetesDependentResourceConfig()
-          .setLabelSelector(DEPENDENT_RESOURCE_LABEL_SELECTOR));
+          .setLabelSelector(SELECTOR + "=true"));
```
metacosm, line 90: Why was the tests working if the selector was wrong?

csviri: since those were not mutal exclusive, the other selector is about not having a label what was true.

metacosm (approved):

PR #1268, opened 2022-06-03
Multiple dependent resource integration test

rsynek:
An IT that shows multiple dependent resources of the same type.

Based on a discussion on Discord with @csviri .


csviri (approved):
LGTM!

thx!

metacosm (approved):
We need to make this use case work for managed dependent resources as well… I've started looking into it and will open as issue for it since we can't currently do it correctly. A potential way to fix the problem would be to use the `name` field of the `Dependent` annotation but I still need to hammer out the details.

PR #1276, opened 2022-06-07
feat: cover more cases when trying to determine generic types

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
```diff
@@ -89,13 +90,56 @@ static boolean getBooleanFromSystemPropsOrDefault(String propertyName, boolean d
...
+          } catch (Exception ex) {
+            // try on the parent
+            return getFirstTypeArgumentFromSuperClassOrInterface(superclass,
+                expectedImplementedInterface);
+          }
+        }
+      }
+      return getFirstTypeArgumentFromInterface(clazz, expectedImplementedInterface);
+    } catch (Exception e) {
+      throw new RuntimeException(
```
csviri, line 140: Maybe throw OperatorException here?

csviri (approved):
LGTM,

made just one comment about an exception type to throw might be different

PR #1280, opened 2022-06-09
Reproduce mock server patch generation increase issue

csviri:
This PR is just to show an issue with MockServer, that generation increases when status is patched.


csviri:
cc @manusa

PR #1284, opened 2022-06-15
Parallel Additional Even Source Start

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -62,20 +59,22 @@ private void postProcessDefaultEventSources() {
...
+    eventSources.additionalNamedEventSources().parallel().forEach(this::startEventSource);
     eventProcessor.start();
   }
 
+  @SuppressWarnings("rawtypes")
+
+
+  @Override
+  public synchronized void stop() {
+    stopEventSource(eventSources.namedControllerResourceEventSource());
```
metacosm, line 72: Shouldn't we stop the controller event source last?

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
csviri: Was thinking about this, but at this does not matter.

Thinking rather that the processor should be stopped first. Before event sources.

PR #1285, opened 2022-06-15
feat: retry and retry configuration decoupling

csviri:


metacosm (approved):
LGTM but this is a breaking change so we have to figure out when / how to introduce it…

csviri:
To be honest, I think we should define what API's we cover under semver, (this might be one). But because of this would not introduce a next major version for sure. So would stick with 3.1 tbh.

csviri:
Will try to make it backwards compatible and deprecate the old previous API

csviri:
Will try to make it backwards compatible, and deprecate the old API. Then remove it in v4.

metacosm:
I'm fine with sticking with 3.1 if it's too much trouble… Or this could wait for some later time… 🤷🏼 😀

PR #1290, opened 2022-06-16
feat: rate limiting

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java
```diff
@@ -69,6 +69,9 @@
   ReconciliationMaxInterval reconciliationMaxInterval() default @ReconciliationMaxInterval(
       interval = 10);
 
+
+  RateLimit rateLimit() default @RateLimit;
```
metacosm, line 73: This will need to be fixed as this doesn't make any sense if users can provide their own `RateLimiter` implementations…

csviri: The idea is that some can configure basic rate limit, what is usuabble in most of the cases. If the default implementation does not fit, configuration overrride still can be used.

Note that this is empty by default, so there is no default rate limiter.

metacosm (changes requested):
See comments

metacosm (approved):

PR #1292, opened 2022-06-17
chore: add @Ignore to internal/intermediate DependentResources

metacosm:
The goal is to allow downstream processes (e.g. the Quarkus extension)
which might need to introspect only user-defined DependentResource
implementations, similarly to what was done for Reconciler
implementations already.


csviri (approved):
LGTM

PR #1296, opened 2022-06-22
feat: run integration tests parallel

csviri:


csviri:
the tests are failing quite randomly, (also there are some solid fails too :) ) will investigate later probably.

PR #1297, opened 2022-06-22
fix: guava version update because of vulnerability

csviri:
see: https://ossindex.sonatype.org/vulnerability/sonatype-2020-0926?component-type=maven&component-name=com.google.guava%2Fguava&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0

This issues does not have any impact on us, just would nice to cover it anyways.


csviri:
see: https://github.com/java-operator-sdk/java-operator-sdk/pull/1328

PR #1300, opened 2022-06-23
Primary to Secondary Mapper

csviri:
see related issue


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -74,28 +75,33 @@
...
     super(client.resources(configuration.getResourceClass()), configuration);
     this.configuration = configuration;
-    primaryToSecondaryIndex =
-        new PrimaryToSecondaryIndex<>(configuration.getSecondaryToPrimaryMapper());
+    primaryToSecondaryMapper = configuration.getPrimaryToSecondaryMapper();
+    if (primaryToSecondaryMapper == null) {
+      primaryToSecondaryIndex =
+          new PrimaryToSecondaryIndex<>(configuration.getSecondaryToPrimaryMapper());
+    } else {
+      primaryToSecondaryIndex = null;
```
metacosm, line 93: Maybe it'd be cleaner to have a no-op index instead of setting it to `null` and having to guard for `null` everywhere else?

csviri: Not sure about that, it is very explicit this way, that the index is not used in that case.

csviri: But also fine with that. Will try how that looks.

csviri: Hmm, here it needs to be explicit:
https://github.com/java-operator-sdk/java-operator-sdk/blob/b7720f974ad2c450c9df6ac5297847af150b33d1/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java#L193-L193

morhidi:
Attila, thank you very much for the prompt fix, I have tested this PR locally. I can confirm this addresses the issue we hit in [FLINK-28008](https://issues.apache.org/jira/browse/FLINK-28008). Do you already have an ETA for the 3.0.3 patch release?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
metacosm: One doesn't prevent the other: we could have a NOOP index and still be able to make it explicit when needed (i.e. `useSecondaryToPrimaryIndex` could check if the set index is the NOOP one instead of `null` where we need to know for sure whether to use it or not). Not a showstopper in any case.

csviri:
> Attila, thank you very much for the prompt fix, I have tested this PR locally. I can confirm this addresses the issue we hit in [FLINK-28008](https://issues.apache.org/jira/browse/FLINK-28008). Do you already have an ETA for the 3.0.3 patch release?

@morhidi after this issue is merged we can release v3.0.3, hopefully today or Monday.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: Changed.

metacosm (approved):

PR #1309, opened 2022-07-01
more event filters

metacosm:
- refactor: share filter handling (wip)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractEventSourceHolderDependentResource.java
```diff
@@ -75,15 +64,9 @@ private RecentOperationCacheFiller<R> recentOperationCacheFiller() {
...
-    this.onAddFilter = onAddFilter;
-  }
-
-  public void setOnUpdateFilter(BiPredicate<R, R> onUpdateFilter) {
-    this.onUpdateFilter = onUpdateFilter;
-  }
-
-  public void setOnDeleteFilter(BiPredicate<R, Boolean> onDeleteFilter) {
-    this.onDeleteFilter = onDeleteFilter;
+  public void initFilters(Predicate<R> onAddFilter, BiPredicate<R, R> onUpdateFilter,
```
csviri, line 67: This is not a very nice API, with all the null values required.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -131,12 +132,12 @@ public void onDelete(R resource, boolean b) {
...
     primaryToSecondaryIndex.onDelete(resource);
     super.onDelete(resource, b);
-    if (acceptedByDeleteFilters(resource, b)) {
+    if (eventAcceptedByDeleteFilters(resource, b)) {
       propagateEvent(resource);
     }
   }
 
-  private synchronized void onAddOrUpdate(Operation operation, R newObject, R oldObject,
+  private synchronized void onAddOrUpdate(ResourceAction operation, R newObject, R oldObject,
```
csviri, line 140: The separate Operation was intentional to exclude delete. I don't think this needs to be changed, it is more explicit that way.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: That is something completely different from filters.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ResourceEventSource.java
```diff
@@ -25,11 +25,7 @@ default Optional<R> getSecondaryResource(P primary) {
 
   Set<R> getSecondaryResources(P primary);
 
-  void setOnAddFilter(Predicate<R> onAddFilter);
```
csviri, line 28: Not sure why is not initFilters better than having setters?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractEventSourceHolderDependentResource.java
csviri, line 67: When this is intended to be called? Since it should be after`initEventSource`, but that is called after the event source initialized. Can we add an integration test for this?

PR #1317, opened 2022-07-05
feat: decouple configuration of rate limiting and retry from controller

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -34,7 +34,10 @@ default boolean isGenerationAware() {
   String getAssociatedReconcilerClassName();
 
   default Retry getRetry() {
-    return GenericRetry.fromConfiguration(getRetryConfiguration()); // NOSONAR
+    final var configuration = getRetryConfiguration();
+    return RetryConfiguration.DEFAULT != configuration
```
csviri, line 38: Shouldn't be equals here?

csviri (commented):
I have mixed feelings about this concept (well that includes positive ones :) ) , for now we had every configuration in `@ControllerConfiguration` configutation, here some of the configuration would be not part of it but a separate annotation. It makes sense if users want to configure own implementation of these concepts (retry, rate limit). However the default implementations should cover most of the cases. The extensibility is mostly there for sake of generality. 

I'm not opposing, I see what problem this solves, just pointing out that separate annotations, are subject of developers annoyance, since you have to learn this feature from the docs, not directly visible from the code.
Therefore we have to document this very carefully both in docs and javadoc

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/RetryConfiguration.java
```diff
@@ -1,26 +1,23 @@
 package io.javaoperatorsdk.operator.api.config;
 
-public interface RetryConfiguration {
```
csviri, line 3: This is a non backwards compatible change, pls create a separate annotation, and mark the former interfaces deprecated

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/LimitingRateOverPeriod.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.concurrent.TimeUnit;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface LimitingRateOverPeriod {
```
csviri, line 11: Would rather have it names `RateLimited` or `RateLimitedOverPeriod`, but such rate limiting is always over period, so maybe just the simplest possible.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java
```diff
@@ -92,13 +96,15 @@ ReconciliationMaxInterval reconciliationMaxInterval() default @ReconciliationMax
...
-
   /**
    * Optional list of {@link Dependent} configurations which associate a resource type to a
    * {@link io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource} implementation
    *
    * @return the list of {@link Dependent} configurations
    */
   Dependent[] dependents() default {};
+
+  Class<? extends Retry> retry() default GenericRetry.class;
```
csviri, line 107: Maybe have some javadoc here? link how to configur it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -111,12 +110,6 @@ public ControllerConfigurationOverrider<R> withRetry(Retry retry) {
     return this;
   }
 
-  @Deprecated
-  public ControllerConfigurationOverrider<R> withRetry(RetryConfiguration retry) {
```
csviri, line 115: This should be also backwards compatible, so should not be removed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/PeriodRateLimiter.java
```diff
@@ -6,38 +6,37 @@
...
 import java.util.Optional;
 
+import io.javaoperatorsdk.operator.api.config.AnnotationConfigurable;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 
 /**
  * A Simple rate limiter that limits the number of permission for a time interval.
  */
-public class PeriodRateLimiter implements RateLimiter {
+public class PeriodRateLimiter
```
csviri, line 15: Rate limiting is always for period, maybe name it as `LinearRateLimiter` (other one way is exponential, but actually usually not for this case how we use it here)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/PeriodRateLimiter.java
```diff
@@ -58,4 +57,23 @@ public Optional<Duration> acquirePermission(ResourceID resourceID) {
...
   }
+
+  @Override
+  public void initFrom(LimitingRateOverPeriod configuration) {
+    this.refreshPeriod = Duration.of(configuration.within(),
+        configuration.unit().toChronoUnit());
+    this.limitForPeriod = configuration.maxReconciliations();
+  }
+
+  public boolean isActivated() {
```
csviri, line 68: Don;t think this needs to be exposed, thus should not be public.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/retry/RetryingGradually.java
```diff
@@ -0,0 +1,28 @@
+package io.javaoperatorsdk.operator.processing.retry;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RetryingGradually {
```
csviri, line 10: Maybe `GradualRetry`, but this implementation is such generic, that maybe just `Retry` ?

csviri (commented):
Made comments, mostly about naming.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/PeriodRateLimiter.java
metacosm: It's used in a test in a different package so no choice… :(

csviri (approved):

PR #1319, opened 2022-07-07
refactor: manage state associated with a resource in a single spot

### LINKED ISSUE #1314: Investigate if all state associated with a given resource could be managed in a single spot
Candidate states to be unified: retry, event marker, rate limiting…
The most likely spot would be `EventProcessor`. 

Pros:
- state is managed in a single spot and `EventProcessor` knows when the state needs to be modified / cleaned-up
- would simplify moving to a multi-cluster model since how resource state is tracked would be more localised instead of spread in different spots

Cons:
- probably needs to abstract state in a generic way

metacosm:
This is a preliminary exploration to address #1314.

Thinking about this some more, this has additional advantages:
- we could pass the state along once it's initially retrieve to the `EventProcessor` methods instead of retrieving it all the time from the `ResourceID`
- this might allow to remove some synchronisation needs because the full state needed for processing would be passed around
- if we think it's a good idea, it probably is better to do it now because of the impact it has on the `RateLimiter` interface


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/RateLimiter.java
```diff
@@ -3,20 +3,17 @@
 import java.time.Duration;
 import java.util.Optional;
 
-import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.rate.RateLimiter.RateLimitState;
 
-public interface RateLimiter {
+public interface RateLimiter<S extends RateLimitState> {
+  interface RateLimitState {
```
csviri, line 9: Can't this be just an Object? Do we need an explicit interface for this. 
I mean just use `RateLimiter<S>`

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/RateLimiter.java
csviri: Otherwise LGTM

metacosm: Yeah, not sure either… Was thinking that perhaps we needed some some of marker interface to be able to trace where this state is used more easily than tracking `Object`… Will review.

metacosm: I think we should keep it this way so that it's more obvious what the associated state is for, even though technically `Object` would work just as well.

csviri: Can't remember where but we had the same situation, where we decided to go with the object and don't use marker interfaces. 
So I lean more to remove it, but I'm fine having that if you think it has benefits.

metacosm: It also has the advantage of having an interface already in case we need to somehow restrict what can be used…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/PeriodRateLimiter.java
```diff
@@ -36,11 +30,12 @@ public PeriodRateLimiter(Duration refreshPeriod, int limitForPeriod) {
   }
 
   @Override
-  public Optional<Duration> acquirePermission(ResourceID resourceID) {
-    if (limitForPeriod <= 0) {
+  public Optional<Duration> isLimited(RateLimitState rateLimitState) {
```
metacosm, line 33: This is essentially where having a marker interface helps for documentation purposes as the alternative would be: `isLimited(Object state)` which wouldn't be as explicit. It's not a big deal but I think it's nicer this way.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/rate/RateLimiter.java
metacosm: See also https://github.com/java-operator-sdk/java-operator-sdk/pull/1319/files#r917897536

PR #1320, opened 2022-07-07
refactor: put all filters in one interface

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/filter/EventFilter.java
```diff
@@ -0,0 +1,22 @@
...
+
+  default boolean acceptsUpdating(R from, R to) {
+    return true;
+  }
+
+  default boolean acceptsDeleting(R resource) {
+    return true;
+  }
+
+  default boolean rejects(R resource) {
```
csviri, line 19: Why is this method works with inverted logic?

csviri:
It is strange that integration tests are not running on this PR.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/filter/EventFilter.java
metacosm: Because I think it is more explicit that way because the generic filter is about overriding whatever the other filters say, rejecting resources even if they would be accepted by the other filtering methods…

csviri: I think that brings more harm then good, imho even goind with this one interface, that should not be inverted. It is quite straigforwad that there is a conjuction between those rules.

csviri:
More and more I think of this, I lean to have it to separate interfaces. If forces users more to think what they want to achieve with the filters.
Basically a generic filter is about ignoring the resources. And the others really makes think the users in what situation it should be used.

PR #1323, opened 2022-07-08
feat: add metadata support for Metrics

### LINKED ISSUE #1322: Add resource type/kind to `Metrics` method to distinguish between identically named resources of different types

metacosm:
Right now, the only metadata that's provided is the group/version/kind
information associated with the resource being processed.

Fixes #1322.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -22,4 +25,21 @@ default <T> Optional<T> getSecondaryResource(Class<T> expectedType) {
   ControllerConfiguration<P> getControllerConfiguration();
 
   ManagedDependentResourceContext managedDependentResourceContext();
+
+  ResourceID currentlyReconciledResourceID();
```
csviri, line 29: Why are we exposing this to the user? This is not usable at all for him. Could be very confusing? Is this this for metrics?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -22,4 +25,21 @@ default <T> Optional<T> getSecondaryResource(Class<T> expectedType) {
...
+
+  Map<String, Object> metadata();
```
csviri, line 31: Same here why is this in the context and not hidden from the user?

csviri (approved):
LGTM

PR #1326, opened 2022-07-11
fix: naming on configuration override

csviri:


metacosm:
I don't see the point of this change and actually disagree with it. The previous style reads better, imo.

csviri:
> I don't see the point of this change and actually disagree with it. The previous style reads better, imo.

I don't think we use present continues tense on any other place, actually if I saw it in a java ever :) 
Basically this is just a builder should use standard wording IMHO.

Also that is mixing the two styles then. Actually this was pointed out one of the users. At least for me feels quite strange. Maybe we can discuss with broader audience...

PR #1328, opened 2022-07-11
refactor (#1327): Remove autoservice

iocanel:
The pull request removes google autoservice as a dependency.
Not 100% sure if this is what @csviri meant in #1327 but either way it's a small change that removes a compile time dependency. So win win.

### DIFF
--- a/operator-framework/pom.xml
+++ b/operator-framework/pom.xml
@@ -27,12 +27,6 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.auto.service</groupId>
-      <artifactId>auto-service</artifactId>
-      <scope>compile</scope>
-    </dependency>
-
-    <dependency>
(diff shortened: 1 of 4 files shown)


operator-framework/pom.xml
```diff
@@ -65,7 +59,6 @@
     <dependency>
       <groupId>io.fabric8</groupId>
       <artifactId>crd-generator-apt</artifactId>
-      <scope>test</scope>
```
csviri, line 68: The CRD generator is just required for the tests, is it required to remove this? (I guess not)

csviri:
Wow, yes, exactly what I mean, thx!
made just a small comment.

operator-framework/pom.xml
iocanel: Removed that while experimenting. I'll put it back!

csviri (approved):
LGTM

PR #1332, opened 2022-07-13
chore: apt processors now suggests the latest version

iocanel:
Using a fixed version works ok-ish, till you upgrade your jdk version. Then annoying warnings start creeping into the build.
This is a more permanent solution.


metacosm (approved):
LGTM apart from format (unused imported class).

Issue #1334, opened 2022-07-13
ReconcilerUtils.loadYaml should allow loading resources ready to build.


iocanel:
My understanding is that the the `ReconcilerUtils.loadYaml` is meant to load an initial yaml file and then programmatically modify it.
It would make sense for me to allow loading that resource directly as builder:

```java
DeploymentBuilder builder = ReconcilerUtils.loadYaml(DeploymentBuilder.class, getClass(), "deployment.yaml");
Deployment deployment = builder
    .accept(ContainerBuilder.class, c -> c.withImage("my-image")) 
    // More changes here
    .build();
```

metacosm:
I've been thinking about implementing something like this for Quarkus… so that the YAML parsing / reflection could be done at build time. Not too sure how that would look like, though.

PR #1335, opened 2022-07-13
feat: allow loading resources as builders

iocanel:
Resolves: #1334 

`loadYaml` checks if the target class is a builder and if so, it initalizes the builder using the deserialized yaml.


csviri (approved):
LGTM

PR #1345, opened 2022-07-15
feat: update control update methods with change check

csviri:
Provides alternative methods for update control to check if the resource and/or status sub-resource actually changed, and do the requests just in that case.


csviri:
We decided to park this issue intil v4 where this behavior will be the default.

PR #1354, opened 2022-07-19
fix: should only create index if primary to secondary mapper is not null

metacosm:


csviri (changes requested):
I think this is actually right in the version, not needs to be changed

metacosm:
Indeed. My bad!

PR #1375, opened 2022-07-26
fix: replace label selector mock server test with IT

csviri:
Note that the former test was doing actually more than it should be. So this is a replacement but also change in scope.


metacosm:
Is what the previous test was covering covered in other tests?

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/labelselector/LabelSelectorTestCustomResource.java
```diff
@@ -0,0 +1,15 @@
...
+import io.fabric8.kubernetes.client.CustomResource;
+import io.fabric8.kubernetes.model.annotation.Group;
+import io.fabric8.kubernetes.model.annotation.ShortNames;
+import io.fabric8.kubernetes.model.annotation.Version;
+
+@Group("sample.javaoperatorsdk")
+@Version("v1")
+@ShortNames("lst")
+public class LabelSelectorTestCustomResource
+    extends CustomResource<Void, LabelSelectorTestCustomResourceStatus>
```
metacosm, line 13: Why create a status class when it's not used anywhere and empty?

metacosm (changes requested):

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/labelselector/LabelSelectorTestCustomResource.java
csviri: Good point, fixed. (I think I had an issue with this in the past when status was void, but not anymore)

csviri:
> Is what the previous test was covering covered in other tests?

Yes it should.

metacosm (approved):

metacosm (approved):

PR #1382, opened 2022-08-02
feat: use fabric8 client json mapper by default

csviri:
The json mapper might be fine tuned for application in some cases, probably in the default case we should use that everywhere if possible.

see also: https://github.com/fabric8io/kubernetes-client/issues/1308


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -126,7 +127,7 @@ default boolean closeClientOnStop() {
   }
 
   default ObjectMapper getObjectMapper() {
-    return OBJECT_MAPPER;
```
metacosm, line 129: Shouldn't the constant be removed altogether, as well, then?

metacosm (approved):

PR #1384, opened 2022-08-03
refactor: remove smoke-test-samples

csviri:
IMO these are pretty obsolete (at least I did not open them for a very long time)


metacosm:
Sure but aren't they used to check backwards compatibility?

csviri:
> Sure but aren't they used to check backwards compatibility?

Probably not, at least that is also covered by the normal samples. 
Basically the default samples cover that.

metacosm (approved):

PR #1388, opened 2022-08-08
refactor: clean-up WorkflowReconcileResult and Condition

metacosm:
- fix: grammar
- refactor: clean up WorkflowReconcileResult
- refactor: make it easier to create Conditions


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java
```diff
@@ -1,10 +1,11 @@
...
+import java.util.Optional;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
-import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
 public interface Condition<R, P extends HasMetadata> {
 
-  boolean isMet(DependentResource<R, P> dependentResource, P primary, Context<P> context);
+  boolean isMet(P primary, Optional<R> secondary, Context<P> context);
```
csviri, line 10: Optionals should not be used as an argument. Can we just have it as a resource what is nullabel?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
```diff
@@ -11,14 +11,11 @@
...
-  Class<? extends Condition> readyPostcondition() default VoidCondition.class;
+  Class<? extends Condition> readyPostcondition() default Condition.class;
 
-  @SuppressWarnings("rawtypes")
-  Class<? extends Condition> reconcilePrecondition() default VoidCondition.class;
+  Class<? extends Condition> reconcilePrecondition() default Condition.class;
 
-  @SuppressWarnings("rawtypes")
-  Class<? extends Condition> deletePostcondition() default VoidCondition.class;
+  Class<? extends Condition> deletePostcondition() default Condition.class;
```
csviri, line 18: TBH before it was more expressive before, kinda describing that what is there it is a special value - so presumably handled specially subsequently. Taking a look on this is little confusing? maybe document it?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java
csviri, line 10: I'm not sure about this change, various dependent resource implementations can have some additional information. Like dependent resources for external resources with state, information about the state.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java
```diff
@@ -1,10 +1,11 @@
 package io.javaoperatorsdk.operator.processing.dependent.workflow;
 
+import java.util.Optional;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
-import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
 public interface Condition<R, P extends HasMetadata> {
```
csviri, line 9: Paramters should not be Optional, maybe just document it that it can be `null`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java
csviri: Also this is a non-backwards compatible change. But that not on issue for now IMHO :)

csviri: Paramters should not be Optional, maybe just document it that it can be null.

metacosm: DependentResource classes should not be accessible directly to user-controlled code, imo. While unlikely, someone could call `reconcile` on it from the condition, for example… State should be conveyed via the `Context`, imo.

metacosm: While I generally agree, in this case, Optional conveys exactly the fact that the secondary resource might not exist (in the delete case) without needing additional documentation and calling this method is not under the user's control so there is no issue where a user could potentially pass `null` instead of an `Optional`… so I think it's better this way, though I guess we should document the method, either way :)

csviri: Agree with the context part.  But specifically speking about not that part, although don't see it clearly yet. So let's park that for now. 
If you think about it will be more complex than that where there an additional state in a config map for example for an external dependent resource 
But fine with this now.

csviri: Except the Optional part, that should not be there according Java conventions.

csviri: So I would stick with standard Java convention, it feels quite weird, you could have this argument basically for any case. Having it just nullable, documented would be nice enough.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java
metacosm: I don't think we lost any expressiveness with this change: the default is the non-instantiable interface so it fulfils the same role as a default value (i.e. we can effectively check that the provided value is not the default one) without the need to create an additional class.

csviri: Making `VoidCondition` abstract would help with that, since it would instantly fail. Also now it throws always an exception, so it will instantly recognized in any test something is wrong.

But again I'm fine with both, I like that there are no more dedicated classes, just needs to be documented that this is the intention putting the interface there as default, that is a good enough alternative for me :)

csviri (approved):

PR #1393, opened 2022-08-10
feat: make it possible to also check annotations/labels when matching

### LINKED ISSUE #1392: Make it possible for `GenericKubernetesResourceMatcher` consider metadata as well
More specifically annotations and labels should be checked since the rest of the fields are either irrelevant (e.g. `name`) or not under a user's control (e.g. `managedFields`).

metacosm:
Fixes #1392

### DIFF
--- a/docs/documentation/dependent-resources.md
+++ b/docs/documentation/dependent-resources.md
@@ -86,6 +86,35 @@ possible to not implement any of these traits and therefore create read-only dep
 that will trigger your reconciler whenever a user interacts with them but that are never
 modified by your reconciler itself.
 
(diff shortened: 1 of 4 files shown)


csviri (changes requested):
What is not clear to me in this impl, how it should be used in a real life dependent resource, that should override the `match` method? 
Should we showcase that somehow?

docs/documentation/dependent-resources.md
```diff
@@ -169,26 +198,26 @@ instances are managed by JOSDK, an example of which can be seen below:
 ```java
 
 @ControllerConfiguration(
-        labelSelector = SELECTOR,
```
csviri, line 172: I don't get this, we are both using the same formatter from maven, how was this not formatted? It should fail also on build if not properly formatted

metacosm: I don't know but I've seen weird things going on with format… it doesn't appear as deterministic as it should be (or rather, the determinism is not obvious to me) 😓

csviri (approved):
LGTM

Issue #1395, opened 2022-08-11
isLastAttempt() is false on the single call to handle error when max-attempts is zero


scrocquesel:
## Bug Report

To track the issue raised by https://github.com/quarkiverse/quarkus-operator-sdk/issues/380

scrocquesel:
The problem is that the first reconcile loop has no information from the RetryConfiguration if the builded RetryInfo will actually provide a nextDelay.

As it seems that retryConfiguration cannot be null (there is always a default one), Instead of

```
@Override
public boolean isLastAttempt() {
  return controller.getConfiguration().getRetryConfiguration() == null;
}
```

may be something like 

```
@Override
public boolean isLastAttempt() {
  return controller.getConfiguration().getRetryConfiguration().isActive()l;
}
```

The GenericRetryConfiguration may return false when max-attemps is 0.

scrocquesel:
A repro https://github.com/scrocquesel/java-operator-sdk/blob/gh-1395/operator-framework/src/test/java/io/javaoperatorsdk/operator/ErrorStatusHandlerLastAttemptIT.java

csviri:
I think it should be rather nullable, thus `controller.getConfiguration().getRetryConfiguration()` should return an `Optional`?

scrocquesel:
> I think it should be rather nullable, thus `controller.getConfiguration().getRetryConfiguration()` should return an `Optional`?
> 
> That is better describing the domain, that there might not be a retry.

It means that the configuration process is aware of why a given config for a given implementation would not produce a retry so it can set an empty RetryConfiguration instead of the given config/implementation.

csviri:
@scrocquesel fixed this here: https://github.com/java-operator-sdk/java-operator-sdk/pull/1595 pls take a look if you have some time.

PR #1397, opened 2022-08-12
fix: fix lastAttempt value on first try when maxAttempt is zero

scrocquesel:
Fix https://github.com/java-operator-sdk/java-operator-sdk/issues/1395

I start adding `enabled()` on Retry interface but it feels wrong. Retry is a functional interface so it needs a default and neither true or false is a good choice. Moreover, implementation can wrongly implement it. For eg, return true while not returning at least one next delay.

Seems like ReconciliationDispatcher may call initExecution but if we don't want to create an instance  only to test nextDelay result, it should be propagated up to the EventProcessor to be cached and I don't see how to do that.


operator-framework/src/test/java/io/javaoperatorsdk/operator/config/runtime/AnnotationControllerConfigurationTest.java
```diff
@@ -342,6 +350,16 @@ public UpdateControl<ConfigMap> reconcile(ConfigMap resource, Context<ConfigMap>
     }
   }
 
+  @ControllerConfiguration(retry = NoRetry.class)
```
scrocquesel, line 353: @csviri Tried something with Optional but as expected, configuration is a bit weird. We need a marker class because the given class is not a factory but the actual strategy. Then, there is something wrong because nothing prevent from adding `@GradualRetry` with `NoRetry.class`.

Does something like 
```
@ControllerConfiguration(retry = @GradualRetry(maxAttempts = 0))

@ControllerConfiguration(retry = @NoRetry)
````

with the annotation being a factory returning an Optional<Retry> based on the given configuration could be better ?
I think this has been discussed already when introducing GradualRetry annotation in https://github.com/java-operator-sdk/java-operator-sdk/pull/1317 and I think we hit some limitations.

Maybe we can keep the annotation out of controllerConfiguration but need a way to discover annotations that would configure the controllerConfiguration and let them override default configuration.

```
@GradualRetry(maxAttempts = 0)
@ControllerConfiguration() // retry is no longer exposed

@NoRetry
@ControllerConfiguration() // retry is no longer exposed
````

File: operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java [hunk 2]
```diff
@@ -173,7 +173,9 @@ public int getAttemptCount() {
 
           @Override
           public boolean isLastAttempt() {
-            return controller.getConfiguration().getRetry() == null;
+            // on first try, we can only rely on the configured behavior
+            // if enabled, will at least produce one RetryExecution
+            return !controller.getConfiguration().getRetry().isPresent();
```
scrocquesel, line 178: This needs to be optimized to avoid multiple call to configuration doing initialization each time

csviri:
@scrocquesel what is the state of this PR? Do you intend to continue with this, in that case probably should be rebased

csviri:
Jus out of curiosity why wouldn't you use a retry?

scrocquesel:
> Jus out of curiosity why wouldn't you use a retry?

Generally, for testing purpose or when you set periodic reconciliation to lower impact on external services

scrocquesel:
> @scrocquesel what is the state of this PR? Do you intend to continue with this, in that case probably should be rebased.
> 
> Or I can take a look on the original issue.

Help is welcome, I tried different things but i'm not satisfied.

csviri:
Uhh I think the next branch deletion, closed this.

PR #1402, opened 2022-08-16
Add bom module

honnix:
When a project has multiple modules, publishing a bom is considered to be a good practice for users to more easily manage dependencies. For example: https://search.maven.org/artifact/io.fabric8/kubernetes-client-bom and https://search.maven.org/artifact/io.micrometer/micrometer-bom.

I'm not sure of the actual artifact deployment process, so hopefully this makes sense.

I tested locally and it works.

### DIFF
--- a/operator-framework-bom/pom.xml
+++ b/operator-framework-bom/pom.xml
@@ -6,7 +6,7 @@
     <groupId>io.javaoperatorsdk</groupId>
     <artifactId>operator-framework-bom</artifactId>
     <version>3.1.2-SNAPSHOT</version>
-    <name>Operator SDK for Java Bill of Materials</name>
+    <name>Operator SDK for Java - Bill of Materials</name>
     <packaging>pom</packaging>
 
     <dependencyManagement>
--- a/operator-framework-bom/pom.xml
+++ b/operator-framework-bom/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+  
+    <groupId>io.javaoperatorsdk</groupId>
+    <artifactId>operator-framework-bom</artifactId>
+    <version>3.1.2-SNAPSHOT</version>
+    <name>Operator SDK for Java Bill of Materials</name>
+    <packaging>pom</packaging>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.javaoperatorsdk</groupId>
+                <artifactId>operator-framework-core</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.javaoperatorsdk</groupId>
+                <artifactId>operator-framework</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.javaoperatorsdk</groupId>
+                <artifactId>micrometer-support</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.javaoperatorsdk</groupId>
+                <artifactId>operator-framework-junit-5</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <distributionManagement>
+        <snapshotRepository>
+            <id>ossrh</id>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+        </snapshotRepository>
+    </distributionManagement>
+</project>
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,8 +25,8 @@ jobs:
           cache: 'maven'
       - name: Check code format
         run: |
-          ./mvnw ${MAVEN_ARGS} formatter:validate -Dconfigfile=$PWD/contributing/eclipse-google-style.xml --file pom.xml
-          ./mvnw ${MAVEN_ARGS} impsort:check --file pom.xml
+          ./mvnw ${MAVEN_ARGS} formatter:validate -Dconfigfile=$PWD/contributing/eclipse-google-style.xml -pl '!operator-framework-bom' --file pom.xml
+          ./mvnw ${MAVEN_ARGS} impsort:check -pl '!operator-framework-bom' --file pom.xml
       - name: Run unit tests
         run: ./mvnw ${MAVEN_ARGS} -B test --file pom.xml
 
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
     </properties>
 
     <modules>
+        <module>operator-framework-bom</module>
         <module>operator-framework-core</module>
         <module>operator-framework-junit5</module>
         <module>operator-framework</module>


metacosm (approved):
LGTM. Thank you!

honnix:
Not full sure I understand the error. It seems the build is checking code format, including `pom.xml`, right? But the `pom.xml` for bom does not have the plugin declared.

honnix:
I tried fixing the issue by skipping the bom module `-pl '!operator-framework-bom'` for some of the maven runs.

metacosm:
Upon further review, I'm wondering if the BOM should be part of the project itself (i.e. use the parent) or stays standalone as is currently done… Any informed opinion on the subject?

honnix:
> Upon further review, I'm wondering if the BOM should be part of the project itself (i.e. use the parent) or stays standalone as is currently done… Any informed opinion on the subject?

@metacosm Thanks for replying. That's a good question. It is considered to be a good practice not using project parent in BOM, to avoid leaking dependencies.

metacosm:
> > Upon further review, I'm wondering if the BOM should be part of the project itself (i.e. use the parent) or stays standalone as is currently done… Any informed opinion on the subject?
> 
> @metacosm Thanks for replying. That's a good question. It is considered to be a good practice not using project parent in BOM, to avoid leaking dependencies.

That's kind of what I was thinking but couldn't find a definitive answer on the topic after an admittedly rather short online search… Do you have any reference on the subject?

honnix:
> > > Upon further review, I'm wondering if the BOM should be part of the project itself (i.e. use the parent) or stays standalone as is currently done… Any informed opinion on the subject?
> > 
> > 
> > @metacosm Thanks for replying. That's a good question. It is considered to be a good practice not using project parent in BOM, to avoid leaking dependencies.
> 
> That's kind of what I was thinking but couldn't find a definitive answer on the topic after an admittedly rather short online search… Do you have any reference on the subject?

I can try to something. We had a few leakages internally that caused trouble, but I unfortunately cannot reveal much.

The problem is, if bom has parent pointing to the project root, all the managed deps of the project root will be imported to user project if user imports the bom.

operator-framework-bom/pom.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+  
+    <groupId>io.javaoperatorsdk</groupId>
+    <artifactId>operator-framework-bom</artifactId>
+    <version>3.1.2-SNAPSHOT</version>
+    <name>Operator SDK for Java Bill of Materials</name>
```
metacosm, line 9:
```suggestion
    <name>Operator SDK for Java - Bill of Materials</name>
```

metacosm (approved):

metacosm:
> The problem is, if bom has parent pointing to the project root, all the managed deps of the project root will be imported to user project if user imports the bom.

Yes so let's go with the current organisation, just made a naming change.

honnix:
Just for the record, I created a project with minimum deps.

```xml
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.javaoperatorsdk</groupId>
        <artifactId>operator-framework-bom</artifactId>
        <version>3.1.2-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>

  <dependencies>
    <dependency>
      <groupId>io.javaoperatorsdk</groupId>
      <artifactId>operator-framework-core</artifactId>
    </dependency>
    <dependency>
      <groupId>io.javaoperatorsdk</groupId>
      <artifactId>micrometer-support</artifactId>
    </dependency>
    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-lang3</artifactId>
      <version>3.11.0</version>
    </dependency>
  </dependencies>
```

Note that when not defining parent for `operator-framework-bom`, I will have to give `commons-lang3` a version otherwise maven fails.

```shell
$ mvn dependency:tree

[INFO] com.example:foo:jar:0.0.1-SNAPSHOT
[INFO] +- io.javaoperatorsdk:operator-framework-core:jar:3.1.2-SNAPSHOT:compile
[INFO] |  +- io.fabric8:kubernetes-client:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-core:jar:5.12.3:compile
[INFO] |  |  |  +- io.fabric8:kubernetes-model-common:jar:5.12.3:compile
[INFO] |  |  |  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.1:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-rbac:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-admissionregistration:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-apps:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-autoscaling:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-apiextensions:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-batch:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-certificates:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-coordination:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-discovery:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-events:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-extensions:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-flowcontrol:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-networking:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-metrics:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-policy:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-scheduling:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-storageclass:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-node:jar:5.12.3:compile
[INFO] |  |  +- com.squareup.okhttp3:okhttp:jar:3.12.12:compile
[INFO] |  |  |  \- com.squareup.okio:okio:jar:1.15.0:compile
[INFO] |  |  +- com.squareup.okhttp3:logging-interceptor:jar:3.12.12:compile
[INFO] |  |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.13.1:compile
[INFO] |  |  |  \- org.yaml:snakeyaml:jar:1.28:compile
[INFO] |  |  +- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.13.1:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.1:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-core:jar:2.13.1:compile
[INFO] |  |  +- io.fabric8:zjsonpatch:jar:0.3.0:compile
[INFO] |  |  \- com.github.mifmif:generex:jar:1.0.2:compile
[INFO] |  |     \- dk.brics.automaton:automaton:jar:1.11-8:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.36:compile
[INFO] +- io.javaoperatorsdk:micrometer-support:jar:3.1.2-SNAPSHOT:compile
[INFO] |  \- io.micrometer:micrometer-core:jar:1.9.3:compile
[INFO] |     +- org.hdrhistogram:HdrHistogram:jar:2.1.12:compile
[INFO] |     \- org.latencyutils:LatencyUtils:jar:2.0.3:runtime
[INFO] \- org.apache.commons:commons-lang3:jar:3.11.0:compile        <--------------
```

If I define parent for `operator-framework-bom`, I can simply do:
```xml
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.javaoperatorsdk</groupId>
        <artifactId>operator-framework-bom</artifactId>
        <version>3.1.2-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>

  <dependencies>
    <dependency>
      <groupId>io.javaoperatorsdk</groupId>
      <artifactId>operator-framework-core</artifactId>
    </dependency>
    <dependency>
      <groupId>io.javaoperatorsdk</groupId>
      <artifactId>micrometer-support</artifactId>
    </dependency>
    <dependency>
      <groupId>org.apache.commons</groupId>
      <artifactId>commons-lang3</artifactId>  <!-- version leaked from parent pom -->
    </dependency>
  </dependencies>
```

And I get:

```shell
$ mvn dependency:tree

[INFO] com.example:foo:jar:0.0.1-SNAPSHOT
[INFO] +- io.javaoperatorsdk:operator-framework-core:jar:3.1.2-SNAPSHOT:compile
[INFO] |  +- io.fabric8:kubernetes-client:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-core:jar:5.12.3:compile
[INFO] |  |  |  +- io.fabric8:kubernetes-model-common:jar:5.12.3:compile
[INFO] |  |  |  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.1:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-rbac:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-admissionregistration:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-apps:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-autoscaling:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-apiextensions:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-batch:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-certificates:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-coordination:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-discovery:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-events:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-extensions:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-flowcontrol:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-networking:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-metrics:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-policy:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-scheduling:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-storageclass:jar:5.12.3:compile
[INFO] |  |  +- io.fabric8:kubernetes-model-node:jar:5.12.3:compile
[INFO] |  |  +- com.squareup.okhttp3:okhttp:jar:4.10.0:compile
[INFO] |  |  |  +- com.squareup.okio:okio-jvm:jar:3.0.0:compile
[INFO] |  |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib-common:jar:1.5.31:compile
[INFO] |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib:jar:1.6.20:compile
[INFO] |  |  |     \- org.jetbrains:annotations:jar:13.0:compile
[INFO] |  |  +- com.squareup.okhttp3:logging-interceptor:jar:4.10.0:compile
[INFO] |  |  |  \- org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.6.10:compile
[INFO] |  |  |     \- org.jetbrains.kotlin:kotlin-stdlib-jdk7:jar:1.6.10:compile
[INFO] |  |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.13.1:compile
[INFO] |  |  |  \- org.yaml:snakeyaml:jar:1.28:compile
[INFO] |  |  +- com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.13.1:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.1:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-core:jar:2.13.1:compile
[INFO] |  |  +- io.fabric8:zjsonpatch:jar:0.3.0:compile
[INFO] |  |  \- com.github.mifmif:generex:jar:1.0.2:compile
[INFO] |  |     \- dk.brics.automaton:automaton:jar:1.11-8:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.36:compile
[INFO] +- io.javaoperatorsdk:micrometer-support:jar:3.1.2-SNAPSHOT:compile
[INFO] |  \- io.micrometer:micrometer-core:jar:1.9.3:compile
[INFO] |     +- org.hdrhistogram:HdrHistogram:jar:2.1.12:compile
[INFO] |     \- org.latencyutils:LatencyUtils:jar:2.0.3:runtime
[INFO] \- org.apache.commons:commons-lang3:jar:3.12.0:compile        <--------------
```
So there `org.apache.commons:commons-lang3:jar:3.12.0` is leaked from parent pom.

metacosm:
Yes, we should try to remove that dependency altogether…

honnix:
> Yes, we should try to remove that dependency altogether…

I just randomly picked up a dependency from parent, and this leakage applies to any managed dependency in parent.

Please feel free to merge this whenever suitable. Thank you so much for reviewing and providing great feedback.

metacosm:
> > Yes, we should try to remove that dependency altogether…
> 
> I just randomly picked up a dependency from parent, and this leakage applies to any managed dependency in parent.

Yes, I understood that point but this particular dependency is something we've been considering getting rid of for a while, actually :)

> 
> Please feel free to merge this whenever suitable. Thank you so much for reviewing and providing great feedback.

Thank you for your contribution to this project!

honnix:
> Yes, I understood that point but this particular dependency is something we've been considering getting rid of for a while, actually :)

Nice!. Fewer is better. :D

csviri (approved):
LGTM

operator-framework-bom/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+            </dependency>
+            <dependency>
+                <groupId>io.javaoperatorsdk</groupId>
+                <artifactId>operator-framework-junit-5</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <distributionManagement>
```
csviri, line 37: Just to my understanding, why distribution management is here?

csviri: And why just snapshot?

csviri (changes requested):
We should probably use this in the examples.

operator-framework-bom/pom.xml
honnix: Good question. Since we cannot inherit from parent for bom as explained in previous comments, and parent pom has https://github.com/java-operator-sdk/java-operator-sdk/blob/main/pom.xml#L190-L195, so I just replicated here.

honnix:
> We should probably use this in the examples.

@csviri Do you mean doing that in this PR? That might require slightly bigger changes since the sample module inherits those dependencies from parent, so if we start importing from bom, it's better to cut that off (the same dep leakage issue as described above).

honnix:
Hmm, it's actually discarding `dependencyManagement` https://github.com/java-operator-sdk/java-operator-sdk/blob/main/sample-operators/tomcat-operator/pom.xml#L28

operator-framework-bom/pom.xml
csviri: Hmm, I guess this should be defined in the end project where the framework is used. Not here. But not sure tbh.

csviri:
> > We should probably use this in the examples.
> 
> @csviri Do you mean doing that in this PR? That might require slightly bigger changes since the sample module inherits those dependencies from parent, so if we start importing from bom, it's better to cut that off (the same dep leakage issue as described above).

Not necessarily in this PR. But in that case would be better to target the `next` branch with it. (Just rebase it onto `next` and edit the PR to target `next`.) Since it's not a patch, rather a new feature for 3.2. Also we would like to keep main releasable with patches all the time.

operator-framework-bom/pom.xml
honnix: https://central.sonatype.org/publish/publish-maven/#other-prerequisites seems to explain why `snapshotRepository` is needed when using `nexus-staging-maven-plugin` (this project uses it in `release` profile).

PR #1403, opened 2022-08-18
refactor: make it easier to initialize mapper & cloner in subclasses

metacosm:
- feat: use fabric8 client json mapper by default (#1382)
- refactor: make it easier to initialize mapper & cloner in subclasses


csviri (approved):
LGTM

PR #1420, opened 2022-08-24
fix: improve ConfigurationServiceOverrider's resilience

metacosm:
- fix: make it easier to support new field when overriding configuration
- refactor: simplify constructors chaining


csviri (approved):
LGTM

Issue #1424, opened 2022-08-26
Improvements on `CachingInboundEventSource`


csviri:
Currently [`CachingInboundEventSource`](https://github.com/java-operator-sdk/java-operator-sdk/blob/95d25fe9ea6426f5a340f68544c199ac29a13575/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/inbound/CachingInboundEventSource.java) has an issue with initialization of state. Currently the cache is only populated explicitly calling the `handleResourceEvent`. Therefore when related primary resources is reconciled the state might not be initialized. 

To fix this a resource fetching mechanism could be implemented, similarly as in `PerResourcePollingEventSource`, thus first time the primary resource appears it will initialize the state by fetching the target resource.

Optionally could have a cache expiration period, for the cases if some event is lost, making sure that if a longer period the resource is not refreshed, it is fetched it again.

metacosm:
More generally speaking, I think our caches should be able to transparently check the API server if a resource is not found locally…

csviri:
I'nm not sure how that would look like, also it really depends on type of resource and event source.  There are different behaviors, like `PerResourceEventSource` has completely different behavior than `PollingEventSource`. And I think it is good. So that the event sources populate the caches, with their specific behavior is the way to IMO.

scrocquesel:
> More generally speaking, I think our caches should be able to transparently check the API server if a resource is not found locally…

Regarding external resource caching, at least, the cache should be able to handle staleness. Currently, the resource must be removed from the cache and it become complicated to detect if an event should be triggered especially on deleted resource.

PR #1427, opened 2022-08-27
feat: add resource fetcher to CachingInboundEventSource (#1424)

scrocquesel:
I don't see real practical use case where we don't want a `ResourceFetcher`. There is a breaking change due to signature change but if  `ResourceFetcher` is left optional, then then current behavior is flawed and I doubt any one use it yet.
Otherwise, we can keep the current behavior but it means being able to handle that  `Set<R> getSecondaryResources(P primary);` may not return a valid set, i.e. return an Optional.empty to tell it doesn't know anything about the secondary ressource associated to primary yet but it is a bigger design change.

If ResourceFetcher is mandatory, then there is a lot of code shared with `PerResourcePollingEventSource`. The code is refactored so that the public interface of both EventSource remains clearly separated and do not leak the base implementation. Only the logger categoy would change.

Be aware that `PerResourcePollingEventSource` now throw an argument exception when period is invalid (instead of when calling timer.schedule).


scrocquesel:
@csviri First thought on that one ?

scrocquesel:
Oh, I guess, I should invalid the cache and let the fetcher be called if needed instead of polling Nevermind this one. I'll rework the timer task, no need to share code with `PerResourcePollingEventSource` .

PR #1428, opened 2022-08-28
feat: add resource fetcher to CachingInboundEventSource

scrocquesel:
Refs: https://github.com/java-operator-sdk/java-operator-sdk/issues/1424

Signature change but I doubt anyone use this event source as it is not currently working well.

Expiration needs a single fetcher by secondary, So I prefered seperating the behavior in a another event source as not everyone will need this.

### DIFF
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/inbound/CachingInboundEventSourceTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/inbound/CachingInboundEventSourceTest.java
@@ -1,0 +1,89 @@
+package io.javaoperatorsdk.operator.processing.event.source.inbound;
+
+import java.util.Set;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import io.javaoperatorsdk.operator.TestUtils;
+import io.javaoperatorsdk.operator.processing.event.EventHandler;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.AbstractEventSourceTestBase;
+import io.javaoperatorsdk.operator.processing.event.source.CacheKeyMapper;
+import io.javaoperatorsdk.operator.processing.event.source.SampleExternalResource;
+import io.javaoperatorsdk.operator.sample.simple.TestCustomResource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class CachingInboundEventSourceTest extends
+    AbstractEventSourceTestBase<CachingInboundEventSource<SampleExternalResource, TestCustomResource>, EventHandler> {
+
+  public static final int PERIOD = 150;
+  private CachingInboundEventSource.ResourceFetcher<SampleExternalResource, TestCustomResource> supplier =
+      mock(
+          CachingInboundEventSource.ResourceFetcher.class);
+  private TestCustomResource testCustomResource = TestUtils.testCustomResource();
+  private CacheKeyMapper<SampleExternalResource> cacheKeyMapper =
+      r -> r.getName() + "#" + r.getValue();
+
+  @BeforeEach
+  public void setup() {
+    when(supplier.fetchResources(any()))
+        .thenReturn(Set.of(SampleExternalResource.testResource1()));
+
+    setUpSource(new CachingInboundEventSource<>(supplier,
+        SampleExternalResource.class, cacheKeyMapper));
+  }
+
+  @Test
+  void getSecondaryResourceFromCacheOrSupplier() throws InterruptedException {
+    when(supplier.fetchResources(any()))
+        .thenReturn(Set.of(SampleExternalResource.testResource1()));
+
+    var value = source.getSecondaryResources(testCustomResource);
+
+    verify(supplier, times(1)).fetchResources(eq(testCustomResource));
+    verify(eventHandler, never()).handleEvent(any());
+    assertThat(value).hasSize(1);
+
+    value = source.getSecondaryResources(testCustomResource);
+
+    assertThat(value).hasSize(1);
+    verify(supplier, times(1)).fetchResources(eq(testCustomResource));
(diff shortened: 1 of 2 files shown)


csviri (changes requested):
Thank you! 
The implementation looks good to me, could you add unit tests pls?

csviri:
Maybe take a look on:
https://github.com/java-operator-sdk/java-operator-sdk/blob/7603d0374400ac804ac85c09c7b866851f2d5ad7/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ResourceEventAware.java#L5-L5

(Used in PerResourcePollingEventSource)

This might be an interesting aspect of it, so the cleanup is called if custom resource is deleted. But the first fetching (maybe event an optional periodic fetching in the background might be some nice addition. But probably in an other PR :)

scrocquesel:
> Maybe take a look on:
> 
> https://github.com/java-operator-sdk/java-operator-sdk/blob/7603d0374400ac804ac85c09c7b866851f2d5ad7/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ResourceEventAware.java#L5-L5
> 
> (Used in PerResourcePollingEventSource)
> 
> This might be an interesting aspect of it, so the cleanup is called if custom resource is deleted. 

That's already there.


> But the first fetching (maybe event an optional periodic fetching in the background might be some nice addition. But probably in an other PR :)

Indeed, I make another PR but it's more complicated than it appears. Doing a period fetching will become no more than a PerResourcePolling (with inbound event on top to maybe get fresh date quicker) and it's not really what this event source should do IMHO. It should only fetch data in the cache when required and not update the cache proactively when it become stale.

scrocquesel:
> Thank you! The implementation looks good to me, could you add unit tests pls?

I know you will ask for but as there wasn't yet, I secretly hope you will not :)
I think it also miss an integration test to show how to use it. Should the class be derived and start/stop overriden to start/stop a web listener and bind request to the public handle mthod. Or should it be share somehow between the operator and a web listener. I'm especially wondering how it could be integrated in a quarkus operator with quarkus rest services or websocket.

csviri:
> Or should it be share somehow between the operator and a web listener. 

It should, I think operator should not deal at this point with servlets.

csviri:
> > Maybe take a look on:
> > https://github.com/java-operator-sdk/java-operator-sdk/blob/7603d0374400ac804ac85c09c7b866851f2d5ad7/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ResourceEventAware.java#L5-L5
> > 
> > (Used in PerResourcePollingEventSource)
> > This might be an interesting aspect of it, so the cleanup is called if custom resource is deleted.
> 
> That's already there.

Ahh nice.
Shoudn't we fetch the resource for `onResourceCreated` ? So the reconciliation is faster. 

> > But the first fetching (maybe event an optional periodic fetching in the background might be some nice addition. But probably in an other PR :)
> 
> Indeed, I make another PR but it's more complicated than it appears. Doing a period fetching will become no more than a PerResourcePolling (with inbound event on top to maybe get fresh date quicker) and it's not really what this event source should do IMHO. It should only fetch data in the cache when required and not update the cache proactively when it become stale.

Ok, agree, let's  stay with the simple approach for now.

scrocquesel:
> > > Maybe take a look on:
> > > https://github.com/java-operator-sdk/java-operator-sdk/blob/7603d0374400ac804ac85c09c7b866851f2d5ad7/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ResourceEventAware.java#L5-L5
> > > 
> > > (Used in PerResourcePollingEventSource)
> > > This might be an interesting aspect of it, so the cleanup is called if custom resource is deleted.
> > 
> > 
> > That's already there.
> 
> Ahh nice. Shoudn't we fetch the resource for `onResourceCreated` (async way, so the Thread is not blocked) ? So the reconciliation is faster.

It will be fetched by the getSecondaryResources ? What would be the race condition in this case ? Will it wait for the parrallel fetch to complete ?

> 
> > > But the first fetching (maybe event an optional periodic fetching in the background might be some nice addition. But probably in an other PR :)
> > 
> > 
> > Indeed, I make another PR but it's more complicated than it appears. Doing a period fetching will become no more than a PerResourcePolling (with inbound event on top to maybe get fresh date quicker) and it's not really what this event source should do IMHO. It should only fetch data in the cache when required and not update the cache proactively when it become stale.
> 
> Ok, agree, let's stay with the simple approach for now.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/inbound/CachingInboundEventSourceTest.java
```diff
@@ -0,0 +1,89 @@
...
+    verify(eventHandler, never()).handleEvent(any());
+    assertThat(value).hasSize(1);
+
+    value = source.getSecondaryResources(testCustomResource);
+
+    assertThat(value).hasSize(1);
+    verify(supplier, times(1)).fetchResources(eq(testCustomResource));
+    verify(eventHandler, never()).handleEvent(any());
+
+    source.handleResourceEvent(ResourceID.fromResource(testCustomResource),
```
scrocquesel, line 62: @csviri Had to use the Set.of version to avoid replacing the existing secondary. I guess the correct behavior should be additive.

```
source.handleResourceEvent(primaryID, a)
source.handleResourceEvent(primaryID, b)
source.handleResourceEvent(primaryID, c)
source.handleResourceDeleteEvent(primaryID, c)
source.getSecondaryResources(testCustomResource) // should return (a, b)
source.handleResourceEvent(primaryID, Set.of(x,y))
source.getSecondaryResources(testCustomResource) // should return (x,y)
```

Wait for your confirmation and will add javadoc so it's more clear to the user

csviri: What do you mean by additive exactly?

csviri: No it should not be. At least that is not what I had on my mind.

csviri: Probably that would be good to document.
We could add overloaded or rather differerently named versions for additiveness, that would be useful for sure. But this is on purpose, now.

csviri (approved):
LGTM

csviri:
@metacosm this LGTM, pls take a look if you will find some time.

csviri:
@scrocquesel could you rebase this to the next branch? and target next with the PR. `next` contains the next minor release with new features, `main` is usually there for patching.

metacosm (approved):

PR #1429, opened 2022-08-28
feat: add RecoverableFromMissingEventCachingInboundEventSource

scrocquesel:
@csviri TDB but I doubt we really need this.


csviri (commented):
Yep, maybe if there is a direct request for it. Would not implement this until that time.

PR #1440, opened 2022-09-02
WIP: unify configuration

metacosm:
- refactor: cleaner separation between instantiation and configuration
- refactor: move instantiation functionality to Utils
- refactor: unify context information creation, move it to Utils
- refactor: improve getFirstTypeArgumentFromInterface
- feat: introduce AnnotationDependentResourceConfigurator concept

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
@@ -1,5 +1,9 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent.managed;
 
+import java.util.Optional;
+
 public interface DependentResourceConfigurator<C> {
   void configureWith(C config);
+
+  Optional<C> configuration();
 }
--- a/sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
+++ b/sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
@@ -9,9 +9,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.Secret;
+import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.Deleter;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.EventSourceProvider;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.managed.AnnotationDependentResourceConfigurator;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.managed.DependentResourceConfigurator;
 import io.javaoperatorsdk.operator.processing.dependent.Creator;
 import io.javaoperatorsdk.operator.processing.dependent.external.PerResourcePollingDependentResource;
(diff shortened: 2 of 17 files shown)


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
```diff
@@ -24,13 +26,18 @@
 import static io.javaoperatorsdk.operator.sample.dependent.SecretDependentResource.MYSQL_SECRET_USERNAME;
 import static java.lang.String.format;
 
+@SchemaConfig(pollPeriod = 700, host = "127.0.0.1",
+    port = SchemaDependentResource.LOCAL_PORT,
+    user = "root", password = "password") // NOSONAR: password is only used locally, example only
 public class SchemaDependentResource
     extends PerResourcePollingDependentResource<Schema, MySQLSchema>
     implements EventSourceProvider<MySQLSchema>,
     DependentResourceConfigurator<ResourcePollerConfig>,
```
csviri, line 35: I guess this can be removed, it's enough just to have `AnnotationDependentResourceConfigurator` ?

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
```diff
@@ -24,13 +26,18 @@
 import static io.javaoperatorsdk.operator.sample.dependent.SecretDependentResource.MYSQL_SECRET_USERNAME;
 import static java.lang.String.format;
 
+@SchemaConfig(pollPeriod = 700, host = "127.0.0.1",
+    port = SchemaDependentResource.LOCAL_PORT,
+    user = "root", password = "password") // NOSONAR: password is only used locally, example only
```
csviri, line 31: I guess the password here should have some placeholder value for "not set"

csviri (commented):
Overall about the dynamic configuration and unification, what is a very good idea IMO, just few points / tech details (other than the comments):

1. The `DependentResourceConfigurator` does not seems to be used anywhere directly, it's always at the end it's subinterface: `AnnotationDependentResourceConfigurator` 

2. What `AnnotationDependentResourceConfigurator` does it basically two responsibilities, creates the configuration from from annotation, and marks that which annotation is related to that configuration.
Could we decouple this? Like having `DependentResourceConfigurator` interface, that receives the configuration. It does not need to be annotation it can go just from an override or anything.  But would separate the creation maybe of the confoguration from annotation. One way would be maybe with a nice(?) is to put an new annotation for the configuration annotation, that says how to produce a config from this annotation.  Like for `@SchemaConfig`:

```
@ConfigInput(configFactory = AnnotationSchemaConfigFactory.class)
@Target({ElementType.TYPE})
public @interface SchemaConfig {
 
  int pollPeriod() default DEFAULT_POLL_PERIOD;

 // omitted 
}
```

It will create the configuration instance from this annotation, calling a method on the factory: `AnnotationSchemaConfigFactory`

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
```diff
@@ -1,5 +1,9 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent.managed;
 
+import java.util.Optional;
+
 public interface DependentResourceConfigurator<C> {
```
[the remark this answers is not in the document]
metacosm: Backwards compatibility.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
metacosm: It is, but I wanted to make things backwards-compatible.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
metacosm: Yes, this isn't a realistic implementation, just showing here how it is possible to also use annotations to configure dependents.

metacosm:
> Overall about the dynamic configuration and unification, what is a very good idea IMO, just few points / tech details (other than the comments):
> 
> 1. The `DependentResourceConfigurator` does not seems to be used anywhere directly, it's always at the end it's subinterface: `AnnotationDependentResourceConfigurator`

It was kept for backwards compatibility.

> 2. What `AnnotationDependentResourceConfigurator` does it basically two responsibilities, creates the configuration from from annotation, and marks that which annotation is related to that configuration.
>    Could we decouple this? Like having `DependentResourceConfigurator` interface, that receives the configuration. It does not need to be annotation it can go just from an override or anything.  But would separate the creation maybe of the confoguration from annotation. One way would be maybe with a nice(?) is to put an new annotation for the configuration annotation, that says how to produce a config from this annotation.  Like for `@SchemaConfig`:
> 
> ```
> @ConfigInput(configFactory = AnnotationSchemaConfigFactory.class)
> @Retention(RetentionPolicy.RUNTIME)
> @Target({ElementType.TYPE})
> public @interface SchemaConfig {
>  
>   int pollPeriod() default DEFAULT_POLL_PERIOD;
> 
>  // omitted 
> }
> ```
> 
> It will create the configuration instance from this annotation, calling a method on the factory: `AnnotationSchemaConfigFactory`

That's an interesting idea but won't that make things complex to process because you'd need to do meta-annotation processing?

PR #1442, opened 2022-09-05
fix: bom version after release

csviri:


metacosm (approved):

honnix:
Not sure what I missed. I thought this should be updated automatically by maven release plugin. Also it seems 3.2.0 did not publish the bom artifact to maven central.

honnix:
OK I see this now https://github.com/java-operator-sdk/java-operator-sdk/pull/1444. Sorry about that.

An alternative way could be setting `-DprocessAllModules=true` for `mvn versions:set`, but having a dedicated step seems to be better.

I guess there would be no backfill release for bom until 3.2.1 is out, right?

csviri:
probably not, but that will be done probably soon, since anyways there was a small bug with worfklows.

PR #1445, opened 2022-09-07
chore: restore fabric8 5.12 compatibility

metacosm:


metacosm:
This work has been ported to the `v3` branch. If you wish to keep using JOSDK with Fabric8 5.x, please use the 3.x releases of JOSDK though we encourage everyone to upgrade to JOSDK 4 and Fabric8 6 since it contains numerous improvements and will be the target of future improvements while only bug fixes will be done on the v3 branch.

PR #1449, opened 2022-09-09
fix: issue with post condition in workflow execution delete

csviri:


metacosm (commented):
LGTM but it'd be interesting to add a unit test to cover this case since we didn't catch that issue with the existing tests.

csviri:
> LGTM but it'd be interesting to add a unit test to cover this case since we didn't catch that issue with the existing tests.

The related issue is about the faling test

PR #1450, opened 2022-09-11
Try to get lease namespace if unspecified

honnix:
This PR makes providing lease namespace optional. If unspecified, it will try to read from a file pointed by system property or env var `kubenamespace` with default value as `/var/run/secrets/kubernetes.io/serviceaccount/namespace` which is injected by k8s. The code is inspired by https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java.

The idea is to simplify use code because in most cases users will need to do the same. Please let me know if this makes sense. Thanks.

@csviri @metacosm

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -39,14 +39,15 @@ public class LeaderElectionManager {
   public void init(LeaderElectionConfiguration config, KubernetesClient client) {
     this.identity = identity(config);
     final var leaseNamespace =
-        config.getLeaseNamespace().or(LeaderElectionManager::tryNamespaceFromPath);
-    if (leaseNamespace.isEmpty()) {
+        config.getLeaseNamespace().orElseGet(
+            () -> ConfigurationServiceProvider.instance().getClientConfiguration().getNamespace());
+    if (leaseNamespace == null) {
       final var message =
           "Lease namespace is not set and cannot be inferred. Leader election cannot continue.";
       log.error(message);
       throw new IllegalArgumentException(message);
     }
-    final var lock = new LeaseLock(leaseNamespace.orElseThrow(), config.getLeaseName(), identity);
+    final var lock = new LeaseLock(leaseNamespace, config.getLeaseName(), identity);
     // releaseOnCancel is not used in the underlying implementation
     leaderElector =
         new LeaderElectorBuilder(
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -39,7 +39,7 @@ public class LeaderElectionManager {
   public void init(LeaderElectionConfiguration config, KubernetesClient client) {
     this.identity = identity(config);
     final var leaseNamespace =
-        config.getLeaseNamespace().or(LeaderElectionManager::tryGetInClusterNamespace);
+        config.getLeaseNamespace().or(LeaderElectionManager::tryNamespaceFromPath);
     if (leaseNamespace.isEmpty()) {
       final var message =
           "Lease namespace is not set and cannot be inferred. Leader election cannot continue.";
@@ -96,7 +96,7 @@ public class LeaderElectionManager {
     return id;
   }
 
-  private static Optional<String> tryGetInClusterNamespace() {
+  private static Optional<String> tryNamespaceFromPath() {
     log.info("Trying to get namespace from Kubernetes service account namespace path...");
 
     final var serviceAccountNamespace =
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -107,16 +107,17 @@ public class LeaderElectionManager {
     if (serviceAccountNamespaceExists) {
       log.info("Found service account namespace at: [{}].", serviceAccountNamespace);
       try {
-        return Optional.of(Files.readString(serviceAccountNamespacePath));
+        return Optional
+            .of(Files.readString(serviceAccountNamespacePath).strip());
       } catch (IOException e) {
         log.error(
             "Error reading service account namespace from: [" + serviceAccountNamespace + "].", e);
         return Optional.empty();
       }
-    } else {
-      log.warn("Did not find service account namespace at: [{}].", serviceAccountNamespace);
-      return Optional.empty();
     }
+
+    log.warn("Did not find service account namespace at: [{}].", serviceAccountNamespace);
+    return Optional.empty();
   }
 
   public void start() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -1,9 +1,5 @@
 package io.javaoperatorsdk.operator;
 
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
@@ -16,12 +12,8 @@ import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectionConfig
 import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElector;
 import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectorBuilder;
 import io.fabric8.kubernetes.client.extended.leaderelection.resourcelock.LeaseLock;
-import io.fabric8.kubernetes.client.utils.Utils;
 import io.javaoperatorsdk.operator.api.config.ConfigurationServiceProvider;
 import io.javaoperatorsdk.operator.api.config.LeaderElectionConfiguration;
-
-import static io.fabric8.kubernetes.client.Config.KUBERNETES_NAMESPACE_FILE;
-import static io.fabric8.kubernetes.client.Config.KUBERNETES_NAMESPACE_PATH;
 
 public class LeaderElectionManager {
 
@@ -95,30 +87,6 @@ public class LeaderElectionManager {
       id = UUID.randomUUID().toString();
     }
     return id;
-  }
-
-  private static Optional<String> tryNamespaceFromPath() {
-    log.info("Trying to get namespace from Kubernetes service account namespace path...");
-
-    final var serviceAccountNamespace =
-        Utils.getSystemPropertyOrEnvVar(KUBERNETES_NAMESPACE_FILE, KUBERNETES_NAMESPACE_PATH);
-    final var serviceAccountNamespacePath = Path.of(serviceAccountNamespace);
-
-    final var serviceAccountNamespaceExists = Files.isRegularFile(serviceAccountNamespacePath);
-    if (serviceAccountNamespaceExists) {
-      log.info("Found service account namespace at: [{}].", serviceAccountNamespace);
-      try {
-        return Optional
-            .of(Files.readString(serviceAccountNamespacePath).strip());
-      } catch (IOException e) {
-        log.error(
-            "Error reading service account namespace from: [" + serviceAccountNamespace + "].", e);
-        return Optional.empty();
-      }
-    }
-
-    log.warn("Did not find service account namespace at: [{}].", serviceAccountNamespace);
-    return Optional.empty();
   }
 
   public void start() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -1,5 +1,9 @@
 package io.javaoperatorsdk.operator;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 
@@ -12,9 +16,12 @@ import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectionConfig
 import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElector;
 import io.fabric8.kubernetes.client.extended.leaderelection.LeaderElectorBuilder;
 import io.fabric8.kubernetes.client.extended.leaderelection.resourcelock.LeaseLock;
-import io.fabric8.kubernetes.client.extended.leaderelection.resourcelock.Lock;
+import io.fabric8.kubernetes.client.utils.Utils;
 import io.javaoperatorsdk.operator.api.config.ConfigurationServiceProvider;
 import io.javaoperatorsdk.operator.api.config.LeaderElectionConfiguration;
+
+import static io.fabric8.kubernetes.client.Config.KUBERNETES_NAMESPACE_FILE;
+import static io.fabric8.kubernetes.client.Config.KUBERNETES_NAMESPACE_PATH;
 
 public class LeaderElectionManager {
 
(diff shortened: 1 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/LeaderElectionConfiguration.java
```diff
@@ -59,8 +68,8 @@ public LeaderElectionConfiguration(
     this.identity = identity;
   }
 
-  public String getLeaseNamespace() {
-    return leaseNamespace;
+  public Optional<String> getLeaseNamespace() {
```
honnix, line 71: Note: this is a breaking change. I can mark it as deprecated and provide another method `getOptionalLeaseNamespace`, if needed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -31,24 +38,42 @@ public LeaderElectionManager(ControllerManager controllerManager) {
...
+    }
+    final var lock = new LeaseLock(leaseNamespace.orElseThrow(), config.getLeaseName(), identity);
     // releaseOnCancel is not used in the underlying implementation
-    leaderElector = new LeaderElectorBuilder(client,
-        ConfigurationServiceProvider.instance().getExecutorService())
-        .withConfig(
-            new LeaderElectionConfig(lock, config.getLeaseDuration(), config.getRenewDeadline(),
-                config.getRetryPeriod(), leaderCallbacks(), true, config.getLeaseName()))
-        .build();
+    leaderElector =
```
honnix, line 51: This is only format change done by `mvn compile`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -31,24 +38,42 @@ public LeaderElectionManager(ControllerManager controllerManager) {
...
 
   public boolean isLeaderElectionEnabled() {
     return leaderElector != null;
   }
 
   private LeaderCallbacks leaderCallbacks() {
-    return new LeaderCallbacks(this::startLeading, this::stopLeading, leader -> {
-      log.info("New leader with identity: {}", leader);
-    });
+    return new LeaderCallbacks(
```
honnix, line 71: This is only format change done by `mvn compile`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -31,24 +38,42 @@ public LeaderElectionManager(ControllerManager controllerManager) {
...
   public void init(LeaderElectionConfiguration config, KubernetesClient client) {
     this.identity = identity(config);
-    Lock lock = new LeaseLock(config.getLeaseNamespace(), config.getLeaseName(), identity);
+    final var leaseNamespace =
+        config.getLeaseNamespace().or(LeaderElectionManager::tryGetInClusterNamespace);
+    if (leaseNamespace.isEmpty()) {
+      final var message =
+          "Lease namespace is not set and cannot be inferred. Leader election cannot continue.";
+      log.error(message);
+      throw new IllegalArgumentException(message);
```
honnix, line 47: Not sure which way is better: throwing an exception to fail or just return so `leaderElector` is still `null`. I prefer to fail loudly, but fine with the other approach.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -31,24 +38,42 @@ public LeaderElectionManager(ControllerManager controllerManager) {
...
+    }
+    final var lock = new LeaseLock(leaseNamespace.orElseThrow(), config.getLeaseName(), identity);
```
honnix, line 49: Not sure which code style this repo prefers, so this change just tries to align with the majority cases.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
metacosm: Failing explicitly is better, imo as well.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -64,13 +89,36 @@ private void stopLeading() {
...
   private String identity(LeaderElectionConfiguration config) {
-    String id = config.getIdentity().orElse(System.getenv("HOSTNAME"));
+    var id = config.getIdentity().orElse(System.getenv("HOSTNAME"));
     if (id == null || id.isBlank()) {
       id = UUID.randomUUID().toString();
     }
     return id;
   }
 
+  private static Optional<String> tryNamespaceFromPath() {
```
csviri, line 99: Not sure if this is needed, could you try instead just use the client's facilities, like:
`Config.autoConfigure(null).getNamespace()` ?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
csviri: yes, fail loudly that I would prefer too.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/LeaderElectionConfiguration.java
csviri: Ahh normally I would prefer backwards compatible with deprecated. `getOptionalLeaseNamespace` does not sound too good, maybe just stick with this this time.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/LeaderElectionConfiguration.java
metacosm: Probably would be better, yes.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
metacosm, line 99: I think we could reasonably ask for the related method to become public in the Fabric8 client so that we don't have to maintain it here…

metacosm (commented):
On the fence regarding the change in API, LGTM otherwise.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
metacosm: 🤔 Would this lead to the same NS? And actually, we should probably get it from `ConfigurationServiceProvider.instance().getClientConfiguration()` instead of creating a new `Config` object that might not match the configured one.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/LeaderElectionConfiguration.java
honnix: I think for users, direct usage of this API will be rare even they do use leader election feature. Probably only in unit test.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
honnix:
```java
default Config getClientConfiguration() {
  return Config.autoConfigure(null);
}
```
This seems create a new config as well.

I can give it a try and see whether those tests still work.

honnix: I could get tests pass by setting `System.setProperty(KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, "false");`, but I worry that loading a full config here might not be a correct approach because it confuses users if they do not expect `.kube/config` to be read at all. The intention of this change is to only make it work in-cluster.

honnix: Yeah that would be better. Do you think I could give it a shot?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
honnix: 👍

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
csviri: Good point, maybe we can then just stick with the current impl for now

csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
honnix: I will give it a try. On the other hand, if we want to get this shipped, we don't need to get blocked because the roundtrip might take some time.

csviri: yep, we can adjust it here after, and ship this as it is. Maybe just create an issue so we don't forget it.

csviri:
Shouldn't we consider this rather a bug? and put it to main?

wdyt @metacosm @honnix  ?

honnix:
@csviri I guess we can. :) Strictly speaking it is still more like a feature, but given this small breaking change, shipping it sooner seems to be better.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
metacosm: > This seems create a new config as well.

That's just the default implementation, though. The Quarkus extension, for example, returns the configuration associated with the client (i.e. the one the client uses to connect to the cluster). 

Are there use cases where the config would be configured to use a different namespace than the one associated with the service account? Do we specifically only want the SA namespace regardless of how the Kubernetes config is set up?

honnix:
I'm making a small tweak.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
honnix: @metacosm OK I see what you mean now. Let me try something.

honnix: So something like https://github.com/java-operator-sdk/java-operator-sdk/pull/1450/commits/0c1294753dba5cdca429e8f1487f792b4cf62bb3 ? If the config was loaded from `.kube/config`, there is for sure no namespace; if it is in-cluster config, the namespace will be read from `/var/run/secrets/kubernetes.io/serviceaccount/namespace` if no overridden.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/LeaderElectionManagerTest.java
```diff
@@ -0,0 +1,75 @@
...
+    assertTrue(leaderElectionManager.isLeaderElectionEnabled());
+  }
+
+  @Test
+  void testInitInferLeaseNamespace(@TempDir Path tempDir) throws IOException {
+    var namespace = "foo";
+    var namespacePath = tempDir.resolve("namespace");
+    Files.writeString(namespacePath, namespace);
+
+    System.setProperty(KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, "false");
```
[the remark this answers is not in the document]
honnix: It is because of https://github.com/fabric8io/kubernetes-client/blob/0742e942eb09f8ed2672f7f82e176b76e595ad50/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/Config.java#L819. Otherwise it will try to read `.kube/config`. By setting this the test case makes sure it tries to read the file.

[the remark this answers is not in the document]
honnix: I'm however still not sure this gets the correct client config.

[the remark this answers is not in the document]
honnix: The order also seems to be wrong:

```java
configurationService.getLeaderElectionConfiguration()
    .ifPresent(c -> leaderElectionManager.init(c, this.kubernetesClient));
ConfigurationServiceProvider.set(configurationService);
```

Now the PR as it is, tries to access `ConfigurationServiceProvider.instance()` before `ConfigurationServiceProvider.set(configurationService)` is called.

For the use case creating k8s client manually, this seems to be required but maybe it is supposed to be so?

```java
var operator =
        new Operator(
            kubernetesClient,
            configurationServiceOverrider ->
                configurationServiceOverrider
                    .withCloseClientOnStop(false)
                    .withClientConfiguration(kubernetesClient.getConfiguration()));
```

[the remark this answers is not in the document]
honnix: Hmm, the instance is already set in `ConfigurationServiceProvider::overrideCurrent`, so `ConfigurationServiceProvider.set(configurationService);` doesn't do anything.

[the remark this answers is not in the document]
honnix: I have a feeling coupling reading namespace from inside a pod with client config seems to complicate things.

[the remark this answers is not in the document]
honnix: Will this be simpler?

```java
final var leaseNamespace =
        config.getLeaseNamespace().orElseGet(
            () -> client.getConfiguration().getNamespace());
```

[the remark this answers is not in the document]
csviri:
```
final var leaseNamespace =
        config.getLeaseNamespace().orElseGet(
            () -> client.getConfiguration().getNamespace());
```
This sounds reasonable for me, also it is good that there is a null check after. Since client namespace can be null too.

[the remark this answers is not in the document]
csviri: It might be good to do an Integration test.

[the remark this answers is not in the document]
honnix: To me there seems to be two separated issues:

1. In case of manually created k8s client, it is better to enforce setting `withClientConfiguration` otherwise there could be a mismatch
2. Which is considered to be a better way retrieving configuration, from client directly or from `ConfigurationServiceProvider`?

I can try to add an integration test.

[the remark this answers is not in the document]
csviri: > In case of manually created k8s client, it is better to enforce setting withClientConfiguration otherwise there could be a mismatch

What do you mean manually created?  In the background it always created, if the namespace is not set it should use always one where the pod is running.

> Which is considered to be a better way retrieving configuration, from client directly or from ConfigurationServiceProvider?

I would say the config service provider since that can be overridden. But I don't really understand that method, shouldn't that call the client in the implementation?
` operator.getKubernetesClient().getConfiguration() `. @metacosm  ? 

So the client should be part of configuration service rather. WDYT?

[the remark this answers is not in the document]
csviri: That might a separate issue to improve this config part probably, I would just use the config service for the sake of this issue.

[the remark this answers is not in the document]
honnix: > What do you mean manually created? In the background it always created, if the namespace is not set it should use always one where the pod is running.

For example:

```java
var kubernetesClient = ...; # build the client; may not be from .kube/config
var operator =
        new Operator(
            kubernetesClient,
            configurationServiceOverrider ->
                configurationServiceOverrider
                    .withCloseClientOnStop(false)
                    .withClientConfiguration(kubernetesClient.getConfiguration()));
```

If here `withClientConfiguration` is not called, I think a default configuration will be created, which might be different than the configuration used to build `kubernetesClient`. Then we have an inconsistency. I'm not sure of the consequence though.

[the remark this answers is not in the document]
csviri: Yes, that is absolutely true. The thing is that on the other hand on quarkus extension it would make issues when the config from config service not used. Note however that if the namespace is configured for a client, it's not our api.  
So we have api for namespace config on controller level. And we have api for leader election to config namespace. So I would expect that the leader election is either uses the pod namespace or a setting explicitly for it.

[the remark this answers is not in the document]
honnix: Yeah I understand that. I think as you said this is a separated issue to track, not related to leader election. I can create a new issue.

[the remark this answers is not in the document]
honnix: I modified the existing e2e test a bit. PTAL, thanks.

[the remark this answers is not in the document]
metacosm: We probably need to clean things up or, at the very least, document them better. The intention is to create client instance(s) based on the configuration retrieved from the `ConfigurationService`. I'm not sure that we do follow that ourselves in our code but we should review it. You should only create client instances with a different configuration only for very specific purposes.

metacosm (approved):

honnix:
Could you please help merge this so it doesn't stay too long and get conflict? Thank you.

csviri:
yep, fine by me, @metacosm could you take a look pls?

metacosm:
I already approved it… 😁

metacosm:
Thé only question that trains is whether we should merge it in `main` or `next`…

honnix:
If going to `main` results a bug fix release right way, it might make sense due to the small breaking change. It feels better to ship it sooner than later.

csviri:
I think this is kinda gray area, it's not a bug fix, rather an improvement IMHO. But fine with merging it to the main, and have a release.

honnix:
> I think this is kinda gray area, it's not a bug fix, rather an improvement IMHO. But fine with merging it to the main, and have a release.

Yeah agreed. I don't have a strong opinion. Either way is ok for me.

csviri (approved):

csviri:
merged to main, thank you @honnix !!

honnix:
Thank you all!

metacosm:
Thank you!

PR #1456, opened 2022-09-13
feat: cicd for client v5.x releases

csviri:


.github/workflows/pr.yml
```diff
@@ -10,7 +10,7 @@ on:
   pull_request:
     paths-ignore:
       - 'docs/**'
-    branches: [ main, v1, v2, next ]
+    branches: [ main, main-c5, v1, v2, next ]
```
metacosm, line 13: I think a more explicit name would be better, like `main-f8-5.12` or something similar?

csviri: ok, make sense, maybe not put the minor number of the fabric8 client version, since it might change, something like

`main-f8-5.x` ?

metacosm: Done.

metacosm (approved):

PR #1459, opened 2022-09-14
fix: new target branch name for clien v5.x version flavor

csviri:


.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v3
         if: ${{ endsWith(github.event.release.tag_name, 'f8-5.x') }}
```
metacosm, line 20:
```suggestion
        if: ${{ endsWith(github.event.release.tag_name, 'f8v5') }}
```

.github/workflows/release.yml
```diff
@@ -64,7 +64,7 @@ jobs:
...
```
metacosm, line 65:
```suggestion
        if: ${{ endsWith(github.event.release.tag_name, 'f8v5') }}
```

.github/workflows/release.yml
```diff
@@ -99,7 +99,7 @@ jobs:
...
```
metacosm, line 99:
```suggestion
        if: ${{ endsWith(github.event.release.tag_name, 'f8v5') }}
```

metacosm (changes requested):

PR #1461, opened 2022-09-14
fix: revert to using Fabric8 Client 5.12.3

metacosm:
- chore: restore fabric8 5.12 compatibility
- chore: add f8-5x maven classifier
- chore: change version to 3.2.0-f8v5-SNAPSHOT


csviri:
The integration tests are not running. Could you pix that pls?

csviri (approved):
We should run the integration tests before merge but otherwise LGTM

PR #1466, opened 2022-09-15
feat: add reusable release workflow and use it to also release BOM

metacosm:


.github/workflows/release.yml
```diff
@@ -8,101 +8,31 @@ jobs:
...
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: "v3"
-      - name: Push changes v4
-        uses: ad-m/github-push-action@master
-        if: ${{ startsWith(github.event.release.tag_name, 'v4.' ) }}
+          version-branch: "v3"
+      - if: ${{ startsWith(github.event.release.tag_name, 'v4.' ) }}
+        env:
+          version-branch: "main"
+      - name: Fail if version-branch is not set
```
csviri, line 23: These steps does not seems to have a name

csviri:
I did think we need this until we did not try the current approach, what is much simpler

csviri (approved):
LGTM

PR #1479, opened 2022-09-19
Optional eventsource on dependent resources

csviri:
depends on: https://github.com/java-operator-sdk/java-operator-sdk/pull/1448


metacosm:
Let's merge this so that we can progress faster. To me, at least 2 things need to be addressed: the reliance on index-based for bulk resources needs be somehow removed or at the very least isolated (#1504) and the optional handling of event sources needs to be simplified (#1513). Then we can clean things up some more… Right now, we have too many alternatives.

metacosm (approved):

PR #1484, opened 2022-09-19
fix: config service override executor service concurrent reconciliation numbers

csviri:
Fixes a bug that if only the number overwrite did not end up with correct executor service.


csviri:
cc @morhidi

PR #1493, opened 2022-09-21
feat: enable configuring a handler to listen to informers stopping

metacosm:
Requires https://github.com/fabric8io/kubernetes-client/pull/4412
- chore: update to SNAPSHOT version of fabric8 client
- feat: enable configuring a handler to listen to informers stopping

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/InformerStoppedHandler.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/InformerStoppedHandler.java
@@ -1,0 +1,9 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+public interface InformerStoppedHandler {
+
+  @SuppressWarnings("rawtypes")
+  void onStop(SharedIndexInformer informer, Throwable ex);
+}
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
@@ -13,6 +13,7 @@ import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.ReconcilerUtils;
+import io.javaoperatorsdk.operator.api.config.ConfigurationServiceProvider;
 import io.javaoperatorsdk.operator.processing.LifecycleAware;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 import io.javaoperatorsdk.operator.processing.event.source.IndexerResourceCache;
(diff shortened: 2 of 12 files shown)


csviri (approved):
LGTM,

would it be possible to add an integration test? (not sure how though)

metacosm:
> LGTM,
> 
> would it be possible to add an integration test? (not sure how though)

I'm not sure either apart from what we already test in this PR. Note also that the informer part is already tested in the fabric8 client, we just need to test our integration, which I believe we do somewhat in this PR.

csviri:
> > LGTM,
> > would it be possible to add an integration test? (not sure how though)
> 
> I'm not sure either apart from what we already test in this PR. Note also that the informer part is already tested in the fabric8 client, we just need to test our integration, which I believe we do somewhat in this PR.

Yep, no doubt this is good enough. Just meatn in a way if that would be easy to do, would not harm,

csviri:
cc @andreaTP 
could you review this pls, should be support your use case.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
```diff
@@ -25,6 +26,18 @@
 
   public InformerWrapper(SharedIndexInformer<T> informer) {
     this.informer = informer;
+
+    // register
+    ConfigurationServiceProvider.instance().getInformerStoppedHandler()
+        .ifPresent(ish -> {
+          final var stopped = informer.stopped();
+          if (stopped != null) {
```
andreaTP, line 34: in which condition `stopped` is going to be `null` ?
With the current implementation in the `client` it doesn't seem to be possible.

I would leave this check out to fail early if this erroneous situation appears.

andreaTP (commented):
Great work! Thanks a lot!

One little comment and a question.

How do we inject this from a `Reconciler`?
Would it be possible to use it in one of the examples?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
metacosm: Good point. This happens during tests, though, where it would require quite a bit of setup just to mock the future…

andreaTP: I would rather prefer more setup in the tests as opposed to the risk of this to happen and be swallowed in a real operator.
The risk here is to end up with a non-working callback without noticing.

If you are really, really against doing it, I should, at least, ask for a loud warning when this situation occurs .

vmuzikar (approved):
I played around with it with Keycloak Operator and can confirm it works well for our use-case. Thanks!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/InformerStoppedHandler.java
```diff
@@ -0,0 +1,9 @@
+package io.javaoperatorsdk.operator.api.config;
+
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+public interface InformerStoppedHandler {
```
csviri, line 5: Would rather rename it to `InformerStopHandler`

csviri (changes requested):
just one tiny naming issue

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/InformerStoppedHandler.java
metacosm: I think the current name is better because it reflects the fact that the handler is called *after* the informer is stopped.

csviri (approved):

PR #1496, opened 2022-09-23
fix: remove problematic leader election unit test

csviri:
this test would fail for anyone that has kube config inited locally that makes is quite problematic, would rather remove it


csviri:
cc @honnix

honnix:
OK I see, so it would fail if kubeconfig does have `namespace` set.

csviri:
> OK I see, so it would fail if kubeconfig does have `namespace` set.

yep, checking how to unset it, but did not found anything smart to do that :(

honnix:
Let's kill it then. It's not worth. :)

honnix (approved):

metacosm (approved):

PR #1506, opened 2022-09-30
docs: adr regarding the cr deserialization problem

csviri:
Created an ADR for this quesiton. So we are all aligned. 
(Feel free to improve the content and/or wording)

### DIFF
--- a/adr/002-Custom-Resource-Deserialization-Problem.md
+++ b/adr/002-Custom-Resource-Deserialization-Problem.md
@@ -6,7 +6,7 @@ accepted
 
 ## Context
 
-In case there are multiple versions of a custom resource in can happen that a controller/informer tracking
+In case there are multiple versions of a custom resource it can happen that a controller/informer tracking
 such a resource might run into deserialization problem as shown
 in [this integration test](https://github.com/java-operator-sdk/java-operator-sdk/blob/07aab1a9914d865364d7236e496ef9ba5b50699e/operator-framework/src/test/java/io/javaoperatorsdk/operator/MultiVersionCRDIT.java#L55-L55)
 .
@@ -24,10 +24,10 @@ Alternatives:
 
 ## Decision
 
-From the JOSDK perspective is fine if the informer stops, and the users decides if the whole operator should stop
+From the JOSDK perspective, it is fine if the informer stops, and the users decides if the whole operator should stop
 (usually the preferred way). The reason, that this is an obvious issue on platform level (not on operator/controller
 level). Thus, the controller should not receive such custom resources in the first place, so the problem should be
-addressed on platform level. Possibly introducing conversion hooks, or labeling for the target resource.
+addressed at the platform level. Possibly introducing conversion hooks, or labeling for the target resource.
 
 ## Consequences
 
--- a/adr/002-Custom-Resource-Deserialization-Problem.md
+++ b/adr/002-Custom-Resource-Deserialization-Problem.md
@@ -9,8 +9,11 @@ accepted
 In case there are multiple versions of a custom resource in can happen that a controller/informer tracking
 such a resource might run into deserialization problem as shown
 in [this integration test](https://github.com/java-operator-sdk/java-operator-sdk/blob/07aab1a9914d865364d7236e496ef9ba5b50699e/operator-framework/src/test/java/io/javaoperatorsdk/operator/MultiVersionCRDIT.java#L55-L55)
-. In the mentioned case two versions of a custom resource are not compatible with each other. The informer receives
-both, but naturally not able to deserialize one of them.
+.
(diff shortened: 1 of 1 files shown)


csviri:
cc @manusa @shawkins @andreaTP

csviri:
cc @vmuzikar

adr/002-Custom-Resource-Deserialization-Problem.md
```diff
@@ -0,0 +1,37 @@
...
+
+accepted
+
+## Context
+
+In case there are multiple versions of a custom resource in can happen that a controller/informer tracking
+such a resource might run into deserialization problem as shown
+in [this integration test](https://github.com/java-operator-sdk/java-operator-sdk/blob/07aab1a9914d865364d7236e496ef9ba5b50699e/operator-framework/src/test/java/io/javaoperatorsdk/operator/MultiVersionCRDIT.java#L55-L55)
+. In the mentioned case two versions of a custom resource are not compatible with each other. The informer receives
+both, but naturally not able to deserialize one of them.
```
manusa, line 13: Can we clarify somewhere in the doc that the informer/watcher receives an event for resource with a given Group Version Kind (GVK) but with the structure of a different (former or newer) version (for the same Group and Kind).

This is what in the end might lead to a deserialization Exception if the `Serializer` is configured in a **strict** mode and some of the fields might have unmatched target types because there was a change in between the two versions of the same resource definition.

Note that the deserializer can be set to be more **lenient** by configuring the Serialization Unmatched Field Type module:
```java
Serialization.UNMATCHED_FIELD_TYPE_MODULE.setRestrictToTemplates(true);
```

adr/002-Custom-Resource-Deserialization-Problem.md
```diff
@@ -0,0 +1,37 @@
...
+
+How should the framework or the underlying informer behave?
+
+Alternatives:
+
+1. The informer should skip the resource and should continue to process the resources with the correct version.
```
manusa, line 19: IMHO I think this shouldn't even be considered as an alternative. It might encourage users to ignore errors that in the long-term will definitely have an impact in their application's lifecycle.

csviri: Just adding since it was working this way if until now, and before was mentioned in other conversations. But completely agree with what you say.

adr/002-Custom-Resource-Deserialization-Problem.md
csviri: > Note that the deserializer can be set to be more lenient by configuring the Serialization Unmatched Field Type module:

I think this is something we want to avoid. So processing resource that are not complete in some cases, that can lead to some more confusion. But might be good to put this into the docs as a note.

adr/002-Custom-Resource-Deserialization-Problem.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Consequences
+
+If an Informer stops on such deserialization error, even explicitly restarting it won't solve the problem, since
+would fail again on the same error.
+
+## Notes
+
+- The informer implementation if fabric8 client changed in this regard, before it was not stopping on deserialization
+  error, but as describe this change in behavior is completely acceptable.
```
vmuzikar, line 40:
```suggestion
- The informer implementation in fabric8 client changed in this regard, before it was not stopping on deserialization
  error, but as described this change in behavior is completely acceptable.
```

adr/002-Custom-Resource-Deserialization-Problem.md
csviri: > Can we clarify somewhere in the doc that the informer/watcher receives an event for resource with a given Group Version Kind (GVK) but with the structure of a different (former or newer) version (for the same Group and Kind).

added some explanation, not sure if you meant it this way

vmuzikar (approved):
Makes sense, just found a few typos. Overall LGTM.

adr/002-Custom-Resource-Deserialization-Problem.md
```diff
@@ -0,0 +1,44 @@
...
+which have different version are stored in the original form (not converted) and are not compatible.
+In this case, if there is no further filtering (by labels) informer receives both, but naturally not able to deserialize
+one of them.
+
+How should the framework or the underlying informer behave?
+
+Alternatives:
+
+1. The informer should skip the resource and should continue to process the resources with the correct version.
+2. Informer stops and makes a notification callback.
```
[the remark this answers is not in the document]
csviri: The problem is that at that point is useless (or rather does not fit very well). So the informer is already stopped, event if the resource is fixed from the callback method, on next start there might a other resource with that problem. So it is a restart per problematic resource.

The solution here is to use conversion hooks.

metacosm (changes requested):

metacosm (approved):

PR #1509, opened 2022-10-03
feat: enable configuring a handler to listen to informers stopping

metacosm:
**DO NOT MERGE before Fabric8 Client 6.2.0 is released.**

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
@@ -16,6 +16,7 @@ import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
 import io.fabric8.kubernetes.client.informers.cache.Cache;
 import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.ReconcilerUtils;
+import io.javaoperatorsdk.operator.api.config.ConfigurationServiceProvider;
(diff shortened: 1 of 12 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
```diff
@@ -32,6 +33,27 @@ public InformerWrapper(SharedIndexInformer<T> informer) {
   public void start() throws OperatorException {
     try {
       informer.run();
+
+      // register stopped handler if we have one defined
+      ConfigurationServiceProvider.instance().getInformerStoppedHandler()
+          .ifPresent(ish -> {
+            final var stopped = informer.stopped();
+            if (stopped != null) {
```
csviri, line 41: Should we have here the null check?

pom.xml
```diff
@@ -196,6 +196,17 @@
         </dependencies>
     </dependencyManagement>
 
+    <repositories>
```
csviri, line 199: This is related to this issue? Why was added here?

metacosm: So that this builds with the Fabric8 snapshot version.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java
metacosm: It should never be `null` but I've seen cases where it was (though I think this was due to a race condition that I fixed)… In any case, it doesn't hurt to have it and we have an explicit error in case it's ever `null`.

PR #1512, opened 2022-10-03
fix: move default client to abstract extension, use config if available

metacosm:


operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/AbstractOperatorExtension.java
```diff
@@ -51,7 +52,9 @@ protected AbstractOperatorExtension(
       boolean preserveNamespaceOnError,
       boolean waitForNamespaceDeletion,
       KubernetesClient kubernetesClient) {
-    this.kubernetesClient = kubernetesClient;
+    this.kubernetesClient = kubernetesClient != null ? kubernetesClient
+        : new KubernetesClientBuilder()
+            .withConfig(configurationService.getClientConfiguration()).build();
```
csviri, line 57: Why we want to use the config of the operator? mainly because of namespaces?

csviri (approved):
Added just a question to clarify, othewise LGTM

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/AbstractOperatorExtension.java
csviri: Ahhh, sorry misread a part, sure, make sense.

PR #1504, opened 2022-09-29
refactor: isolate index handling to BulkDependentResource interface

metacosm:

PR #1513, opened 2022-10-03
refactor: simplify handling of reused event sources

metacosm:
DependentResources that provide an EventSource should implement
EventSourceAware. If they want to reuse an already created event source
instead of providing their own, they need to select the appropriate
event source identified by the name specified by the useEventSourceNamed
method.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
```diff
@@ -1,10 +1,36 @@
...
+   * @param context context of event source initialization
+   * @return an optional event source
+   */
+  default Optional<ResourceEventSource<R, P>> eventSource(EventSourceContext<P> context) {
+    return getUsedEventSourceName().map(
+        name -> context.getEventSourceRetriever().getResourceEventSourceFor(resourceType(), name));
+  }
 
-public interface EventSourceAware<P extends HasMetadata> {
+  Class<R> resourceType();
```
csviri, line 29: It's a little smelly that we have resourceType in both places (also in dependent resource), but no big deal.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
```diff
@@ -1,10 +1,36 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent;
 
+import java.util.Optional;
+
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.javaoperatorsdk.operator.processing.event.EventSourceRetriever;
+import io.javaoperatorsdk.operator.api.reconciler.EventSourceContext;
+import io.javaoperatorsdk.operator.processing.event.source.ResourceEventSource;
+
+public interface EventSourceAware<R, P extends HasMetadata> {
```
csviri, line 9: A couple of remarks with this:

- This is in one hand an elegant solution to couple the event source related methods into one interface
on the other hand one of the primary reasons was to put these methods to Dependent Resource to couple it more, thus to express that a dependent resource ideally (an usually, so if not it is more of an corner case) should provide an event source. With this we are getting back to the position that basically event source is an optional thing in general.

- We don't have parity with the `@Dependent` annotation, so while setting there the `eventSource` parameter won't have effect on the dependent if it does not implement this interface. (Not that big deal just pointing out).

csviri, line 9: not sure  if `EventSourceAware` is the proper name, but cannot think anything better.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
```diff
@@ -1,10 +1,36 @@
...
+  default Optional<ResourceEventSource<R, P>> eventSource(EventSourceContext<P> context) {
+    return getUsedEventSourceName().map(
+        name -> context.getEventSourceRetriever().getResourceEventSourceFor(resourceType(), name));
+  }
 
-public interface EventSourceAware<P extends HasMetadata> {
+  Class<R> resourceType();
 
-  void selectEventSources(EventSourceRetriever<P> eventSourceRetriever);
+  void useEventSourceNamed(String eventSourceName);
```
csviri, line 31: - maybe `useEventSourceByName` ?

csviri, line 31: What little feels smelly that ok actually most of our dependent resources use event sources to fill the cache on create/update. But in this interface it is explicit to use a different event source but implicit to not provide one. While one is configurable (by setting the name on `@Dependent`) other can be achieved only by overriding the `default Optional<ResourceEventSource<R, P>> eventSource(EventSourceContext<P> context)` method. This is smelly I think.

csviri (approved):
Made comments, I have mixed feeling of some parts, especially that it is not explicit to not use event source, but this is fine for now since all our dependents work like this. Just not really future proof. (Actually will dig how this works in controller runtime)

csviri:
@metacosm would be good to create a separate PR on next just to see if integration test passing,
( change the CI later to run integration tests on all targets? )

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
metacosm: Yeah, it's not ideal but this allows to have a default implementation in the interface…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
metacosm: Yes, on both. This was an initial attempt to see where that could lead. We can (and probably should) refine things.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
metacosm: I'd rather use `useEventSourceWithName` then.

metacosm:
Let's merge https://github.com/java-operator-sdk/java-operator-sdk/pull/1479 with the understanding that #1504 and this PR need to be addressed after.

csviri:
Thinking about this, what if we just replace the `doNotProvideEventSource`, with `useEventSourceByName` in dependent resource? would not be little better?

metacosm:
> Thinking about this, what if we just replace the `doNotProvideEventSource`, with `useEventSourceByName` in dependent resource? would not be little better?
> 
> regarding the comments I made, will create a PR for that, how it would look like.

Yes, was thinking about removing `EventSourceAware` altogether and put everything on `DependentResource`…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/EventSourceAware.java
```diff
@@ -0,0 +1,36 @@
...
+   * <li>Some special implementation of an event source. That just execute some action might not
+   * provide one.</li>
+   * </ul>
+   *
+   * @param context context of event source initialization
+   * @return an optional event source
+   */
+  default Optional<ResourceEventSource<R, P>> eventSource(EventSourceContext<P> context) {
+    return getUsedEventSourceName().map(
+        name -> context.getEventSourceRetriever().getResourceEventSourceFor(resourceType(), name));
```
csviri, line 26: Hmm also this is a little smelly that the `EventSourceContext` is now meant to be just register event sources, and support registration, not to read it.

PR #1517, opened 2022-10-04
refactor: isolate index handling to BulkDependentResource interface

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -22,16 +20,16 @@
...
+  private final boolean bulk = this instanceof BulkDependentResource;
 
   protected Creator<R, P> creator;
   protected Updater<R, P> updater;
   protected BulkDependentResource<R, P> bulkDependentResource;
   private boolean returnEventSource = true;
-
-  protected List<ResourceDiscriminator<R, P>> resourceDiscriminator = new ArrayList<>(1);
+  private ResourceDiscriminator<R, P> resourceDiscriminator;
+  private int currentCount;
```
csviri, line 32: Should we address this global variable in this PR, or  rather should make a separate one?

metacosm: I would say in a different PR but we can do it here if it makes more sense.

csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
csviri: sure np

PR #1518, opened 2022-10-04
refactor: simplify handling of reused event sources

csviri:


metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DeferrableEventSourceHolder.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.event.EventSourceRetriever;
+
+public interface DeferrableEventSourceHolder<P extends HasMetadata> {
+
+  default void useEventSourceWithName(String name) {}
+
+  Optional<String> resolveEventSource(EventSourceRetriever<P> eventSourceRetriever);
```
csviri: This Optional part feels a little weird. IMO should rather throw something like `EventSourceNotFoundException`.

csviri: It does not have to be even collected, since this will always fail to start the operator, and needs code change. So it is not even a config error if an exception is thrown here.

metacosm: Yeah, it's a little awkward. I'd like to collect it because, as a user, I prefer to know all the mistakes I made at once if possible instead of failing, fixing one thing only to have an error on another thing and fail again…

metacosm (approved):

PR #1520, opened 2022-10-05
refactor: improve configuration utilities

metacosm:


csviri:
Hmm why are there 2 PRs ? for the main and next? probably just should be for one of them.

metacosm:
> Hmm why are there 2 PRs ? for the main and next? probably just should be for one of them.

Sure as long as it gets in both branches ASAP :)

csviri:
> > Hmm why are there 2 PRs ? for the main and next? probably just should be for one of them.
> 
> Sure as long as it gets in both branches ASAP :)

I think we should instantly rebase next on main, if we added something to main. (Not sure if I instantly do though always, but used to, but will work on my discipline :) )

PR #1527, opened 2022-10-06
feat: external resource with generated id support

csviri:


operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/externalstate/ExternalStateReconciler.java
```diff
@@ -0,0 +1,125 @@
...
+            .withName(resource.getMetadata().getName())
+            .withNamespace(resource.getMetadata().getNamespace())
+            .build())
+        .withData(Map.of(ID_KEY, createdResource.getId()))
+        .build();
+    configMap.addOwnerReference(resource);
+    client.configMaps().resource(configMap).create();
+
+    var primaryID = ResourceID.fromResource(resource);
+    configMapEventSource.handleRecentResourceCreate(primaryID, configMap);
```
metacosm, line 81: This method shouldn't be called directly, imo. It shouldn't even be part of the public interface.

csviri: In this case it is needed, since we want to be strongly consistent regarding the event source. 
(For long term we might want to implement these features directly to the client, similarly as in go)

metacosm: Sure, but that's not up to the user to ensure that consistency. If the user has to enforce that, then we need to redesign things because the user shouldn't even be aware of that part as it's internal to how the SDK works.

csviri: Dependent resources will hide this, but some might do it on lower level. The thing is that without that it just don't work, (In all cases). We need probably do some work on these levels in the future. But this is how it is now, hiding that layer would won't solve the problem. And dependent resources using that, as other abstractions might use that user implement itself, so definitelly would not hide it.

PR #1530, opened 2022-10-07
refactor: isolate bulk dependent resource handling more

metacosm:


csviri (commented):
Went through this briefly, tbh I have mixed feelings. I agree that this is kinda more structured, but also was much harder to follow the flow of the execution, since bidirectional references etc. There are some parts that is not entirely clear for me, will take a deeper look on that tomorrow.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/BulkDependentResourceReconciler.java
```diff
@@ -0,0 +1,122 @@
...
+      return asAbstractDependentResource().resourceType();
+    }
+
+    @Override
+    public R create(R desired, P primary, Context<P> context) {
+      return bulkDependentResource.create(desired, primary, context);
+    }
+  }
+
+  private static class UpdatableBulkDependentResourceInstance<R, P extends HasMetadata>
```
csviri, line 113: I don't get why this is needed.

csviri (approved):
I made a comment where it is not clear to me why that is needed. But otherwise although it makes it harder imho to read/follow the flow of the exection, it seprates the two type of logic, let's see how it work out.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/BulkDependentResourceReconciler.java
metacosm: This is needed so that we can have the exact same interface for a "single" or bulk dependent in `AbstractDependentResource`.

PR #1532, opened 2022-10-07
feat: introduce AnnotationDependentResourceConfigurator concept

metacosm:
The idea is to be able to configure dependents using annotations
directly so that we can remove the special handling of
KubernetesDependentResource from AnnotationControllerConfiguration. This
results in dependent resources being instantiated and configured
directly when processed from the annotations in the managed case, thus
rendering the DependentResourceFactory concept obsolete. This should
also lead to further simplification later.


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
```diff
@@ -23,12 +28,17 @@
 import static io.javaoperatorsdk.operator.sample.dependent.SecretDependentResource.MYSQL_SECRET_USERNAME;
 import static java.lang.String.format;
 
+@SchemaConfig(pollPeriod = 700, host = "127.0.0.1",
+    port = SchemaDependentResource.LOCAL_PORT,
+    user = "root", password = "password") // NOSONAR: password is only used locally, example only
 public class SchemaDependentResource
     extends PerResourcePollingDependentResource<Schema, MySQLSchema>
     implements DependentResourceConfigurator<ResourcePollerConfig>,
```
csviri, line 36: I think we should remove the `DependentResourceConfigurator` from here.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
```diff
@@ -23,12 +28,17 @@
 import static io.javaoperatorsdk.operator.sample.dependent.SecretDependentResource.MYSQL_SECRET_USERNAME;
 import static java.lang.String.format;
 
+@SchemaConfig(pollPeriod = 700, host = "127.0.0.1",
+    port = SchemaDependentResource.LOCAL_PORT,
+    user = "root", password = "password") // NOSONAR: password is only used locally, example only
```
csviri, line 33: We should have real life examples as much it makes sense, could we fix that here ? And show how to pass the password from the config override (and remove from the annotation) ?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
```diff
@@ -1,5 +1,9 @@
 package io.javaoperatorsdk.operator.api.reconciler.dependent.managed;
 
+import java.util.Optional;
+
 public interface DependentResourceConfigurator<C> {
```
csviri, line 5: Does this make sense to have, or only the `AnnotationDependentResourceConfiguration`?

csviri (changes requested):
Made some comments, nothing serious, I still think this could be done using an annotation instead of the interface, but we can experiment with that later. 
There are some parts what is not clear to me, we should clarify document those.

csviri:
> Made some comments, nothing serious, I still think this could be done using an annotation instead of the interface, but we can experiment with that later. There are some parts what is not clear to me, we should clarify document those.

Otherwise looks very good :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AnnotationControllerConfiguration.java
```diff
@@ -171,6 +168,22 @@ private <T> void configureFromAnnotatedReconciler(T instance) {
     }
   }
 
+  @SuppressWarnings("unchecked")
+  private void configureFromCustomAnnotation(Object instance) {
+    if (instance instanceof AnnotationDependentResourceConfigurator) {
```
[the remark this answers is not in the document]
metacosm: I don't think so, as this just means that the associated dependent resource doesn't need configuration.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
metacosm: This is kept for backwards-compatibility… though I guess we're breaking it with the other method…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
```diff
@@ -1,5 +1,9 @@
...
   void configureWith(C config);
+
+  Optional<C> configuration();
```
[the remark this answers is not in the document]
metacosm: This is meant to retrieve the configuration that was used previously. This implies that the configuration needs to be kept around by implementations so that it can be retrieved at some later time.

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
metacosm: The whole point here is to show how the dependent resource can be configured from the annotation. I don't have time to create another example.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -159,42 +158,39 @@ public ControllerConfigurationOverrider<R> withGenericFilter(GenericFilter<R> ge
...
-              .map(KubernetesDependentResourceConfig.class::cast)
-              .filter(Predicate.not(KubernetesDependentResourceConfig::wereNamespacesConfigured))
-              .map(c -> updateSpec(drsEntry.getKey(), spec, c))
-              .orElse(drsEntry.getValue());
-        }).collect(Collectors.toUnmodifiableList());
+    final var newDependentSpecs = namedDependentResourceSpecs.values().stream()
+        .peek(spec -> {
+          // if the dependent resource has a NamespaceChangeable config
+          // update the namespaces if needed, otherwise, do nothing
+          if (hasModifiedNamespaces) {
```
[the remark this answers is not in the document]
csviri: Hmm maybe not on this level.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -159,42 +161,59 @@ public ControllerConfigurationOverrider<R> withGenericFilter(GenericFilter<R> ge
...
+      var configurator = (DependentResourceConfigurator) dependentResource;
+      configurator.configureWith(dependentResourceConfig);
+    }
+
+    return this;
   }
 
   @SuppressWarnings("unchecked")
   public ControllerConfiguration<R> build() {
+    // todo: this should be abstracted by introducing an interface to deal with listening to
```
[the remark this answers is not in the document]
csviri: This is not entirely clear for me, but since there is an issue we can discuss there :)

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
csviri: I don't mean creating an another example, just passing the password from for the controller as it should be and not setting here by annotation.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
csviri: Yeah I guess. So basically the `DependentResourceConfigurator` could be deleted I guess and just have the Annotation version of it?

sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
metacosm: Again, the whole point is to show how the configuration can now be done via the annotation. I agree that passing the password in the annotation is dumb but that's what the configuration for this particular example is. Feel free to provide an alternative.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DependentResourceConfigurator.java
metacosm: Well, I'm not sure. That's what I thought originally but then people who don't want to use annotations will be forced to implement an interface that won't make sense to them…  Maybe we still need to clean some things up here. I'd be in favor of doing this in subsequent PRs, though.

csviri (approved):

PR #1536, opened 2022-10-12
wip: alternative to #1527

metacosm:
- feat: external resource with generated id support
- reconciler for IT
- fixes
- IT progress
- wp
- wip
- refactor: add default implementations
- feat: make it possible to retrieve a future event source when ready
- refactor: rework external state IT


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/EventSourceContext.java
```diff
@@ -13,14 +17,14 @@
  */
 public class EventSourceContext<P extends HasMetadata> {
 
-  private final IndexerResourceCache<P> primaryCache;
   private final ControllerConfiguration<P> controllerConfiguration;
   private final KubernetesClient client;
+  private final EventSourceManager<P> eventSourceManager;
 
-  public EventSourceContext(IndexerResourceCache<P> primaryCache,
+  public EventSourceContext(EventSourceManager<P> eventSourceManager,
```
csviri, line 24: This is still smelly, probably we should not add there the whole manager

metacosm:
Obsolete.

PR #1537, opened 2022-10-12
fix: avoid computing alternate value if not needed in Optionals

### LINKED ISSUE #1535: AbstractDependentResource reconciles two times on update
#### What did you do?

I created Reconciler on primary resource and CRUDKubernetesDependentResource connected with it. On the next step i triggered reconciliation which should update dependent resource. 

Single call of `desired` method

Two calls of `desired` method - second is redundant.

#### Additional context

After analysis i assume the root cause is [HERE](https://github.com/java-operator-sdk/java-operator-sdk/blob/0d57629830cdd9ac1c3665438a1493ca438afdc4/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java#L47). The `desired` method is called here even if Optional returned from `computedDesired` has value. It's a pretty common mistake during using optionals, that `orElse` is called instead of `orElseGet` when we really need lazy evaluation.

#### Possible Solution

Just make `desired` call lazily evaluated by using `orElseGet` ;) 

#### Environment

3.1.1

metacosm:
Fixes #1535

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
@@ -64,9 +64,7 @@ public class LeaderElectionManager {
     return new LeaderCallbacks(
         this::startLeading,
         this::stopLeading,
-        leader -> {
-          log.info("New leader with identity: {}", leader);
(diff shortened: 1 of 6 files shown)


csviri (approved):
LGTM

Issue #1221, opened 2022-05-18
Support External Resource State Management for Dependent Resources


csviri:
When  external resources that requires state, it is stored into a ConfigMap, Secret or a dedicated CustomResource. Would be nice to support this patter in Dependent Resources.

csviri:
Regarding design:
- It would be the best if the event sources would transparently provide such translation, if the resource is created it's ID is stored to an (let's say) ConfigMap, the event source would be aware of such config maps. So if the resource is polled (from a polling event source), it will do the the translation for a resource to the target ID. (For the polling event sources, actually the resource fetcher). 

- In terms of dependent resource, the dependent resource will store the ID (on create the ConfigMap) and on cleanup (delete the config map).

Issue #1511, opened 2022-10-03
Clear Approach How Do We Prefer Configuring Dependent Resources - Behavior and Properties


csviri:
Currently there could be two approaches to take how to specify some behavior / properties of a Dependent Resource. One is to specify it as a property, or implement it as an interface. To see on concrete example, the discriminator here is a property:

https://github.com/java-operator-sdk/java-operator-sdk/blob/9c1be9d59151f1673b5b2a5bea60543021fabd34/operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/multiplemanageddependentsametype/MultipleManagedDependentResourceConfigMap2.java#L15-L16

While we set a SecondaryToPrimary mapper for a dependent resource, by implementing an interface:

https://github.com/java-operator-sdk/java-operator-sdk/blob/9c1be9d59151f1673b5b2a5bea60543021fabd34/operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primaryindexer/DependentPrimaryIndexerTestReconciler.java#L33-L34

we should clearly define how to and when to use an approach.

## Notes

The result of this issues should be an ADR (But let's agree on guidlines) , also the adjustment to the implementation of any.

metacosm:
IMO, everything that can be set on a generic object should be a property so that you don't need to create an implementation just to change one thing. Anything that's central to how the dependent works should probably be implemented as an interface… So the distinction would be made on behavior as opposed to something that is more a "using" relationship (i.e. difference between inheritance and composition).

csviri:
> IMO, everything that can be set on a generic object should be a property so that you don't need to create an implementation just to change one thing. Anything that's central to how the dependent works should probably be implemented as an interface… So the distinction would be made on behavior as opposed to something that is more a "using" relationship (i.e. difference between inheritance and composition).

On the other hand we could say that the `trait` (C(R)UD) interfaces could be modelled also as a property. 

But we can live with this definition. Basically then the `Deleter`, `Updater`, `Creator`, `BulkDependentResource`, `DependentResourceWithExplicitState` should be interfaces.

But `GenericResourceFetcher` should be rather a property instead of implemented by the dependent resource?

https://github.com/java-operator-sdk/java-operator-sdk/blob/2cb616c4c4fd0094ee6e3a0ef2a0ea82173372bf/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/PollingDependentResource.java#L13-L13

[...]

PR #1543, opened 2022-10-13
Persistent state dependent resources

csviri:


csviri:
Note that this works now with combination of Bulk resources.

csviri:
closes #1221

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -154,7 +154,7 @@ public void delete(P primary, Context<P> context) {
     dependentResourceReconciler.delete(primary, context);
   }
 
-  protected void handleDelete(P primary, Context<P> context) {
+  protected void handleDelete(P primary, R secondary, Context<P> context) {
     throw new IllegalStateException("delete method be implemented if Deleter trait is supported");
```
metacosm, line 158:
```suggestion
    throw new IllegalStateException("handleDelete method must be implemented if Deleter trait is supported");
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
```diff
@@ -0,0 +1,120 @@
...
+import io.javaoperatorsdk.operator.api.reconciler.dependent.RecentOperationCacheFiller;
+import io.javaoperatorsdk.operator.processing.event.EventSourceRetriever;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.ResourceEventSource;
+import io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource;
+
+public abstract class AbstractExternalDependentResource<R, P extends HasMetadata, T extends ResourceEventSource<R, P>>
+    extends AbstractEventSourceHolderDependentResource<R, P, T> {
+
+  private final boolean isDependentResourceWithExplicitState =
```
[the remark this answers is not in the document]
metacosm: I'm not sure renaming the class will fix the problem… Why would someone extend `AbstractExternalDependentResource` and not implement `DependentResourceWithExplicitState`? Could you detail a use case where someone would extend this class and not want an explicit external state with it?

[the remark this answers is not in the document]
csviri: yep, so we have already samples for that (like the mysql sample):
If somebody want a PollingDependentResource, that does not need an external state.

Note that not every external resource needs an explicit state, for example if it can be addressed only by the information in the custom resource. However some might need one, like if a github PR is managed. Therefore the current implementation allows to a dependent resource to extend PollingDependentResource (and others) and to decide if it needs an external state or not implementing the related `DependentResourceWithExplicitState`.

I thing that we might address in a separate PR and could have an impact on this is this issue:
https://github.com/java-operator-sdk/java-operator-sdk/issues/1511
(but not necessarily)

However I recognize that Polling and PerResourcePolling DRs are quite a thin layer, that basically just creates an instance of a EventSource on top of a standard functionality. (althoug might be extended later for example with an improved configuration). So the structure might change in the future if we for example make it less explic? "Thus just an external event source that provides an event source" abstraction.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/AbstractPollingDependentResource.java
```diff
@@ -2,11 +2,14 @@
...
 import io.javaoperatorsdk.operator.api.reconciler.Ignore;
+import io.javaoperatorsdk.operator.processing.dependent.AbstractExternalDependentResource;
 import io.javaoperatorsdk.operator.processing.event.source.CacheKeyMapper;
+import io.javaoperatorsdk.operator.processing.event.source.ExternalResourceCachingEventSource;
 
 @Ignore
 public abstract class AbstractPollingDependentResource<R, P extends HasMetadata>
-    extends AbstractCachingDependentResource<R, P> implements CacheKeyMapper<R> {
+    extends AbstractExternalDependentResource<R, P, ExternalResourceCachingEventSource<R, P>>
+    implements CacheKeyMapper<R> {
```
metacosm, line 13: Is a polling dependent resource necessarily external?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -1,16 +1,13 @@
 package io.javaoperatorsdk.operator.processing.dependent.kubernetes;
 
 import java.util.HashMap;
-import java.util.Optional;
 import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
metacosm, line 7: Should we throw an exception if secondary is null?

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/external/AbstractPollingDependentResource.java
csviri: Yes, we don't poll kubernetes API.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/externalstate/ExternalWithStateDependentResourceDependentResourceWith.java
```diff
@@ -0,0 +1,99 @@
...
+import io.javaoperatorsdk.operator.processing.dependent.DependentResourceWithExplicitState;
+import io.javaoperatorsdk.operator.processing.dependent.Matcher;
+import io.javaoperatorsdk.operator.processing.dependent.Updater;
+import io.javaoperatorsdk.operator.processing.dependent.external.PerResourcePollingDependentResource;
+import io.javaoperatorsdk.operator.support.ExternalIDGenServiceMock;
+import io.javaoperatorsdk.operator.support.ExternalResource;
+
+import static io.javaoperatorsdk.operator.sample.externalstate.ExternalStateDependentReconciler.ID_KEY;
+
+public class ExternalWithStateDependentResourceDependentResourceWith extends
```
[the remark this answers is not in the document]
csviri: updated

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
csviri: this is an outdated place, could you pls point me again where do you mean it?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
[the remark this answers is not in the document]
csviri: Yep, but don't see how we would cover otherwise just optionally having this external state functionality for Polling Event sources.

metacosm (approved):

PR #1551, opened 2022-10-19
fix: release - change bom version for release release

csviri:


csviri:
This might fix the release since bom will be installed for local mvn cache

metacosm (approved):

PR #1554, opened 2022-10-19
feat: introduce AnnotationDependentResourceConfigurator concept

metacosm:
The idea is to be able to configure dependents using annotations
directly so that we can remove the special handling of
KubernetesDependentResource from AnnotationControllerConfiguration. This
results in dependent resources being instantiated and configured
directly when processed from the annotations in the managed case, thus
rendering the DependentResourceFactory concept obsolete. This should
also lead to further simplification later.


csviri (approved):
I Assume this is the same PR :)

metacosm:
Yes, meant to replace #1532 which was just too annoying to rebase & address conflicts… 😓

PR #1557, opened 2022-10-19
chore: update project version to 4.1.0-SNAPSHOT

metacosm:


csviri (approved):

csviri:
This is fine, although it is not needed necessarily since the release changes the version.

metacosm:
> This is fine, although it is not needed necessarily since the release changes the version.

Not sure I follow what you mean but this is actually needed so that snapshot versions are actually coherent as there shouldn't be anymore `4.0.2-SNAPSHOT` versions being built. For that matter, we should have changed the version earlier because some of these 4.0.2 snapshots were actually referring to what's in `next` while some others to what's in `main`…

csviri:
> > This is fine, although it is not needed necessarily since the release changes the version.
> 
> Not sure I follow what you mean but this is actually needed so that snapshot versions are actually coherent as there shouldn't be anymore `4.0.2-SNAPSHOT` versions being built. For that matter, we should have changed the version earlier because some of these 4.0.2 snapshots were actually referring to what's in `next` while some others to what's in `main`…

Ahh yes, sorry forgot that we have snapshots also from here

PR #1570, opened 2022-10-25
chore: update k8s versions to test against, add Java 19

metacosm:


csviri (approved):

metacosm:
Annoying thing is that this augments the build time quite significantly… Maybe we should remove some older k8s versions from the set? Or maybe only run the full tests as part of the release process as a last check to make sure things still work on older versions?

csviri:
> Annoying thing is that this augments the build time quite significantly… Maybe we should remove some older k8s versions from the set? Or maybe only run the full tests as part of the release process as a last check to make sure things still work on older versions?

hmm, good question, maybe the one oldest kubernetes version?

PR #1572, opened 2022-10-26
feat: add default stop handler

csviri:
I think we should have here system exit by default.

cc @andreaTP @vmuzikar


csviri:
have to double check it but stop handler is called now when the informer is stopped explicitly.

andreaTP (approved):

andreaTP:
> when the informer is stopped explicitly

does this happen? Does JOSDK exposes API to stop watching specific resources?

csviri:
> > when the informer is stopped explicitly
> 
> does this happen? Does JOSDK exposes API to stop watching specific resources?

No, I think just , on s sig term , when we stop the operators, we do a graceful shutdown.

metacosm (approved):

vmuzikar (approved):

PR #1576, opened 2022-10-27
feat: moving cache sync timeout to controller level

csviri:
this makes sense if there are multiple controlelrs, where this value might vary

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -16,6 +16,7 @@ import io.fabric8.kubernetes.api.model.KubernetesResourceList;
 import io.fabric8.kubernetes.client.dsl.MixedOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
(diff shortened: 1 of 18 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -133,4 +135,8 @@ public Stream<R> list(Predicate<R> predicate) {
     return cache.list(predicate);
   }
 
+  public void setControllerConfiguration(ControllerConfiguration<P> controllerConfiguration) {
```
metacosm, line 138: This PR is missing some tests because if you remove this method, everything still builds fine and yet, the `controllerConfiguration` is not properly initialized.

csviri: uhh, good catch, forgot t remove that, before it was used.

csviri: removed.

(The `InformerRelatedBehaviorITS` would fail if this would not work, although it is not very explicit test)

metacosm: Still not working right, though, because now the `controllerConfiguration` is never set… So why even record it and pass it to `initSources`?

metacosm (approved):

PR #1587, opened 2022-11-01
feat: add configurable 'or' event filter to allow reconciling on non-…

Bijnagte:
…generation change update events.

I am not sure I really like the name of the configuration but naming things is hard.

This adds the ability to optionally configure an or filter that can include events filtered by the default filters. My use case is to run a reconciler when there are specific metadata changes (labels and annotations) that don't impact the spec and thus are still for the same generation.


csviri:
Hi @Bijnagte ,
this probably is not the way to go, please turn of generation aware filtering:
https://github.com/java-operator-sdk/java-operator-sdk/blob/fb9dc1aeef2a5b0c9faf69286d39f16409f332f7/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java#L43-L43

And implement a custom filter, that handles your usecase.

(The other internal filters are valid in all situations, those are solving the special cases that always should be solved that way )

Bijnagte:
Not needed since it can be accomplished already with out the added complexity

PR #1595, opened 2022-11-09
fix: isLastAttempt on error handler correct if max attempt 0

csviri:


csviri:
closes #1395

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -38,11 +38,17 @@
...
+  // Usually for testing purposes.
+  private final boolean retryConfigurationHasZeroAttempts;
 
   ReconciliationDispatcher(Controller<P> controller,
       CustomResourceFacade<P> customResourceFacade) {
     this.controller = controller;
     this.customResourceFacade = customResourceFacade;
+
+    var retry = controller.getConfiguration().getRetry();
+    retryConfigurationHasZeroAttempts = retry == null || retry.initExecution().isLastAttempt();
```
scrocquesel, line 51: Great. Didn't think of calling it in the ctor. I was convince that it would only work in a reconciliation context. LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -38,11 +38,17 @@
 
   private final Controller<P> controller;
   private final CustomResourceFacade<P> customResourceFacade;
+  // this is tp handle corner case, when there is a retry, but it is actually limited to 0.
```
metacosm, line 41:
```suggestion
  // this is to handle corner case, when there is a retry, but it is actually limited to 0.
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -173,7 +179,8 @@ public int getAttemptCount() {
 
           @Override
           public boolean isLastAttempt() {
-            return controller.getConfiguration().getRetry() == null;
+            return controller.getConfiguration().getRetry() == null ||
+                retryConfigurationHasZeroAttempts; // check also if the retry is limited to 0
```
metacosm, line 183: I would just reverse the order of the check for optimisation purposes… (not a big deal, obviously 😅)

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
csviri: makes sense, done!

Issue #1599, opened 2022-11-15
cleanup is only called on restart


qinzhichao01:
I am going to use java-operator-sdk to implement my own operator, but I found that when I implemented the Cleaner method, I found that when deleting a pod, the cleanup method was not called, and the k8s pod would not be deleted, only in The cleanup method will be called after restarting the application, and the pod will be deleted. My implementation is as follows. Is my implementation wrong?

```
@ControllerConfiguration
public class CloudIdeOperator implements Reconciler<Pod>, Cleaner<Pod> {
    private volatile int count = 0;
    private volatile int count2 = 0;
    @Override
    public DeleteControl cleanup(Pod pod, Context<Pod> context) {
        count2++;
        System.out.println("cleanUp times:" + count2+"   "+ pod.getMetadata().getName() + ": " + pod.getStatus().getPodIP());
        return DeleteControl.defaultDelete();
    }

    @Override
    public UpdateControl<Pod> reconcile(Pod pod, Context<Pod> context) throws Exception {
        System.out.println(pod.getMetadata().getName() + ": " + pod.getStatus().getPodIP());
        count++;
        System.out.println("crete pod times:" + count);

        return UpdateControl.noUpdate();

    }
}
```
calll method
```
    public static void main(String[] args) {
        KubernetesClient client = new DefaultKubernetesClient();
        Operator operator = new Operator(client);
        operator.register(new CloudIdeOperator());

        operator.installShutdownHook();
        operator.start();
//        operator.stop();

    }
```

csviri:
@qinzhichao01 are you using the newest version?

qinzhichao01:
>do you have any idea to save this problem?

csviri:
will take a look into it

qinzhichao01:
> will take a look into it
that`s

csviri:
@qinzhichao01 added a PR, this will fix the issue. The problem was with something else, the generation filter, the pod does not seems to have `generation` in metadata. This was unexpected, see how it is fixed in the pr.

Until released you can turn off generation aware processing:
https://github.com/java-operator-sdk/java-operator-sdk/blob/2e7fac2a25ea470696de719b52baa880ec252bc4/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java#L43-L43

by setting this flag to false. That should fix you issue.

qinzhichao01:
[...]

PR #1600, opened 2022-11-15
fix: cleaner issue for pod caused by missing generation in metadata

csviri:
closes: https://github.com/java-operator-sdk/java-operator-sdk/issues/1599


metacosm (approved):

csviri:
At the end changed the integration tests to use Service instead of Pod (what has the same issue), because it results in much more stable test.

PR #1606, opened 2022-11-16
fix: run event source start on specific thread pool

### LINKED ISSUE #1603: Move start methods out of the common thread pool
Using parallel streams induces using the common thread pool by default and I/O operations (such as ones implied by informer startup) shouldn't be run on these threads.

metacosm:
Fixes #1603

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
@@ -86,7 +86,9 @@ public class ExecutorServiceManager {
     final var name = thread.getName();
     try {
       thread.setName(threadNamePrefix + "-" + thread.getId());
-      instrumented.submit(task).get(instance().terminationTimeoutSeconds, TimeUnit.SECONDS);
+      instrumented.submit(task)
+          .get(ConfigurationServiceProvider.instance().cacheSyncTimeout().toSeconds(),
+              TimeUnit.SECONDS);
       shutdown(instrumented);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
       throw new OperatorException("Couldn't execute task", e);
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
@@ -70,10 +70,26 @@ public class ExecutorServiceManager {
     return workflowExecutor;
   }
 
+  /**
+   * Runs the specified I/O-bound task and waits for its completion using the ExecutorService
+   * provided by {@link #workflowExecutorService()}
+   * 
+   * @param task task to run concurrently
+   */
   public static void executeAndWaitForCompletion(Runnable task) {
+    executeAndWaitForCompletion(task, instance().workflowExecutorService());
+  }
+
+  /**
+   * Executes the specified I/O-bound task using the specified ExecutorService and waits for its
+   * completion for at most {@link #terminationTimeoutSeconds} seconds.
+   * 
+   * @param task task to run concurrently
+   * @param executor ExecutorService used to run the task
+   */
+  public static void executeAndWaitForCompletion(Runnable task, ExecutorService executor) {
     try {
-      instance().workflowExecutorService().submit(task)
-          .get(instance().terminationTimeoutSeconds, TimeUnit.SECONDS);
+      executor.submit(task).get(instance().terminationTimeoutSeconds, TimeUnit.SECONDS);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
       throw new OperatorException("Couldn't execute task", e);
     }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
@@ -73,7 +73,7 @@ public class ExecutorServiceManager {
   /**
    * Runs the specified I/O-bound task and waits for its completion using the ExecutorService
    * provided by {@link #workflowExecutorService()}
-   * 
+   *
    * @param task task to run concurrently
    */
   public static void executeAndWaitForCompletion(Runnable task) {
@@ -83,7 +83,7 @@ public class ExecutorServiceManager {
   /**
    * Executes the specified I/O-bound task using the specified ExecutorService and waits for its
    * completion for at most {@link #terminationTimeoutSeconds} seconds.
-   * 
+   *
    * @param task task to run concurrently
    * @param executor ExecutorService used to run the task
    */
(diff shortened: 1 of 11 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
```diff
@@ -17,6 +18,8 @@ public class ExecutorServiceManager {
   private static ExecutorServiceManager instance;
   private final ExecutorService executor;
   private final ExecutorService workflowExecutor;
+  private final ForkJoinPool threadPool =
+      new ForkJoinPool(Runtime.getRuntime().availableProcessors());
```
csviri, line 22: 1. `Runtime.getRuntime().availableProcessors()` is not needed, empty constructor is equivalent to this.
2. We probably should not use ForkJoinPool for this purpose, since event sources are IO bound operations. A normal executor service would be a good choice imo, maybe a configurable size; or a size that is same as event source number (but with an upper bound)
see:
https://www.geeksforgeeks.org/java-forkjoinpool-vs-executorservice/

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
metacosm: Good point, I originally was using the executor services defined by the `ConfigurationService` but wasn't seeing changes due to me being dumb (i.e. not using the proper version of the SDK in my testing code) and then forgot to go back to it when I finally managed to get it to work as I was expecting. 😅

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
```diff
@@ -41,7 +46,9 @@
 
   @Override
   public void start() throws OperatorException {
-    sources.values().parallelStream().forEach(LifecycleAware::start);
+    // make sure informers are all started before proceeding further
+    ExecutorServiceManager.executeAndWaitForCompletion(
```
csviri, line 50: This can cause a deadlock. For example if the Thread count is 1 (but also others). So the event source is an InformerEventSource and on start it also want to start event sources from the same thread pool. This should not use the same executor service.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerManager.java
metacosm: Not sure I'm following…

csviri: Lets say there are 2 threads in the in the pool. We start 2 `InformerEventSources` at tha same time. Both started but then both of them starts with the same executor service two InformerManager, but those don't have available threads anymore at that point. Because all the threads available in the executor service are already blocked.

This can be verified if you just create an operator with 1 thread and 1 informer event source, that I assume will never start.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
```diff
@@ -68,6 +70,31 @@ public ExecutorService workflowExecutorService() {
...
   }
 
+  /**
+   * Runs the specified I/O-bound task and waits for its completion using the ExecutorService
+   * provided by {@link #workflowExecutorService()}
+   *
+   * @param task task to run concurrently
+   */
+  public static void executeAndWaitForCompletion(Runnable task) {
+    executeAndWaitForCompletion(task, instance().workflowExecutorService());
```
csviri, line 80: It is a little smelly that this workflow executor is reused for a general purpose executor service IMO.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
```diff
@@ -68,20 +71,51 @@ public ExecutorService workflowExecutorService() {
...
+   *        way
+   */
+  public static void executeAndWaitForCompletion(Runnable task, String threadNamePrefix) {
+    final var executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
+    ExecutorService instrumented = new InstrumentedExecutorService(executor);
+    // change thread name for easier debugging
+    final var thread = Thread.currentThread();
+    final var name = thread.getName();
+    try {
+      thread.setName(threadNamePrefix + "-" + thread.getId());
```
csviri, line 88: Why is the thread name changes here? It's not propagated to the executor, and basically there is just submits tasks and shutdows the executor.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
```diff
@@ -68,20 +71,51 @@ public ExecutorService workflowExecutorService() {
...
 
+  /**
+   * Runs the specified I/O-bound task and waits for its completion using the new ExecutorService
+   *
+   * @param task task to run concurrently
+   * @param threadNamePrefix the prefix with which to prefix thread names when tasks are run this
+   *        way
+   */
+  public static void executeAndWaitForCompletion(Runnable task, String threadNamePrefix) {
+    final var executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
csviri, line 82: This does not seems to be right. Creating a new threadpool for all tasks.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -64,20 +65,30 @@ public void postProcessDefaultEventSourcesAfterProcessorInitializer() {
...
-        .parallel().forEach(this::startEventSource);
-    eventSources.additionalNamedEventSources()
-        .filter(es -> es.priority().equals(EventSourceStartPriority.DEFAULT))
-        .parallel().forEach(this::startEventSource);
+
+    // starting event sources on the workflow executor which shouldn't be used at this point
+    ExecutorServiceManager.executeAndWaitForCompletion(
+        () -> eventSources.additionalNamedEventSources()
+            .filter(es -> es.priority().equals(EventSourceStartPriority.RESOURCE_STATE_LOADER))
+            .parallel()
```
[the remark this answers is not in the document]
metacosm: No, this won't.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
```diff
@@ -64,20 +65,30 @@ public void postProcessDefaultEventSourcesAfterProcessorInitializer() {
...
+            .forEach(this::startEventSource),
+        "LowLevelEventSourceStart");
```
csviri, line 75:
```suggestion
        "ResourceStateLoaderEventSourceStart");
```

csviri (approved):

PR #1613, opened 2022-11-21
feat: improve workflow error reporting

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
```diff
@@ -1,31 +1,28 @@
...
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 public class AggregatedOperatorException extends OperatorException {
 
-  private final List<Exception> causes;
+  private final Map<String, Exception> causes;
```
csviri, line 9: Why is the key String here? It could be `DependentResource` I guess.

csviri (commented):
Just one question / design decision to clarify other than that looks good.

csviri (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/AggregatedOperatorException.java
metacosm: The exception is not necessarily meant only for workflow errors so the key could be anything identifying "what" caused that specific error so that user has some idea of what's going on.

PR #1617, opened 2022-11-22
feat: expose event source metadata

metacosm:


csviri:
Could you make a summary what is the problem being solved with this PR?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/NamedEventSource.java
```diff
@@ -1,19 +1,27 @@
...
 import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.processing.event.source.EventSource;
 import io.javaoperatorsdk.operator.processing.event.source.EventSourceStartPriority;
 
-class NamedEventSource implements EventSource {
+class NamedEventSource implements EventSource, EventSourceMetadata {
 
   private final EventSource original;
   private final String name;
+  private final AssociatedDependentMetadata dependentMetadata;
```
csviri, line 14: Porbably not fully understand the PR, but really don't like this part. The event sources should have no knowledge at all about dependent resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceMetadata.java
```diff
@@ -0,0 +1,9 @@
+package io.javaoperatorsdk.operator.processing.event;
+
+import io.javaoperatorsdk.operator.processing.event.source.EventSource;
+
+public interface EventSourceMetadata {
```
csviri, line 5: Shouldn't we just simply expose `NamedEventSource` ?
This is not really metadata, it is THE event source with name. 
Or rename it to `EventSourceWithMetadata`.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceMetadata.java
metacosm: I debated only using `NamedEventSource`, don't recall why I didn't actually… 😅

metacosm: Ah yes, didn't want to expose `start` / `stop` but since the `EventSource` is exposed… 🤷🏼 
The alternative would be to expose only what's needed from the `EventSource` via another interface.

csviri (approved):
LGMT

PR #1629, opened 2022-11-24
fix: url for ci build

csviri:


metacosm (approved):
No idea what the correct path should be so I trust you on this ;)

csviri:
> No idea what the correct path should be so I trust you on this ;)

THis is just one the relavant builds selected. But checked and the URL at least works now :)

PR #1636, opened 2022-11-29
fix: leader election manager init when not using config overrider

nicoloboschi:
If you create an Operator instance without using the `overrider` to set the leader election config, the configuration singleton is set twice leading to a ISE. 

The method `LeaderElectionManager#init` relies on the singleton config but the config is not set yet at that moment.

Note that I found out this issue by using [Quarkus Operator SDK 5.0.2.Beta](https://github.com/quarkiverse/quarkus-operator-sdk)


csviri (approved):
LGTM thx

metacosm (approved):

PR #1649, opened 2022-12-05
feat: feat controller queue size, execution thread count

csviri:


micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -5,28 +5,55 @@
...
+import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Timer;
 
+import static io.javaoperatorsdk.operator.api.reconciler.Constants.CONTROLLER_NAME;
+
 public class MicrometerMetrics implements Metrics {
 
   private static final String PREFIX = "operator.sdk.";
   private static final String RECONCILIATIONS = "reconciliations.";
+  private static final String RECONCILER_EXECUTING_THREADS = "reconciler.executions.";
```
metacosm, line 30: Why aren't these under `reconciliations`?

metacosm: Also, there's already a `controllers.executions` prefix. How is this new `reconciler.executions.` different? Finally, all metrics should be prefixed by `PREFIX` to ensure consistency and name spacing.

metacosm (changes requested):

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
csviri: good point will fix

metacosm: Unless I'm mistaken, the new metrics are still not using  the `PREFIX` constant as a prefix…

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -118,6 +168,12 @@ public void failedReconciliation(HasMetadata resource, Exception exception,
     return registry.gaugeMapSize(PREFIX + name + ".size", Collections.emptyList(), map);
   }
 
+  private List<Tag> gvkTags(Class<? extends HasMetadata> resourceClass) {
```
metacosm, line 171: There are several spots where GVK information is added to tags. Could this method be reused elsewhere? Also, it looks like sometimes it's prefixed with `resource`, sometimes not…

metacosm (changes requested):
Looks OK apart maybe the GVK tags (see comment).

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
csviri: `gauge` is a little different than others, since it is first needs to initialized. Then incremented or decremented. This is specially needed for that. But made it at least public static for now.

csviri: >  Also, it looks like sometimes it's prefixed with resource, sometimes not…

what do you mean exactly?

metacosm (approved):

PR #1657, opened 2022-12-13
chore: upgrade k8s for CI, only keeping supported versions

metacosm:


csviri (approved):

metacosm:
Looks like the Minikube action (or possibly Minikube itself) is not yet compatible with 1.26.

csviri:
updated to most recent supported versions

.github/workflows/pr.yml
```diff
@@ -35,20 +35,20 @@ jobs:
     strategy:
       matrix:
         java: [ 11, 17 ]
-        kubernetes: [ 'v1.23.15', 'v1.24.9', 'v1.25.5' ]
+        kubernetes: [ 'v1.25.13', 'v1.26.8', 'v1.27.5', 'v1.28.1' ]
```
metacosm, line 38: So I guess we're dropping older versions completely?

csviri: it's a good quesiton, before we said (if ai remember correctly, that we will test on supported versions only), but not sure we also say that we will support last `n` major versions. That might be wiser.

metacosm: The problem is that this will increase the CI run time. We could test older versions on main maybe once a day instead of on each PR?

csviri: Yeah, it's always a compromise, a shorter build is always better, maybe optimize that when we see that it is too long, not sure if it will be. 
It depends also how much of those version we want to run, maybe the last 6 major versions?

PR #1659, opened 2022-12-13
feat: make it possible to disable leader election via configuration

### LINKED ISSUE #1658: Make it possible to disable leader election support via configuration

metacosm:
Fixes #1658


csviri:
Why is this needed? 
First of all this config should be immutable, there is no such scenario that the leader eleciton is started and then it will be turned off. And just not turning on is basically not providing the configuration.

metacosm:
This could be handled differently and, indeed, more work would be needed here to make sure leader election is indeed not used when disabled. Closing this.

PR #1667, opened 2022-12-16
Generate a Versions file to be consumed at runtime

andreaTP:
This will enable compile time checks of the versions in the Quarkus Operator SDK.
ref: https://github.com/quarkiverse/quarkus-operator-sdk/pull/460

cc. @metacosm

### DIFF
--- a/operator-framework-core/pom.xml
+++ b/operator-framework-core/pom.xml
@@ -47,6 +47,19 @@
           <commitIdGenerationMode>full</commitIdGenerationMode>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>templating-maven-plugin</artifactId>
+        <version>1.0.0</version>
+        <executions>
+          <execution>
+            <id>filtering-java-templates</id>
+            <goals>
+              <goal>filter-sources</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
--- a/operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/api/config/Versions.java
+++ b/operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/api/config/Versions.java
@@ -1,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config;
+
+public final class Versions {
+
+  private Versions() {}
+
+  protected static final String JOSDK = "${project.version}";
+  protected static final String KUBERNETES_CLIENT = "${fabric8-client.version}";
+
+}
--- a/operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/Versions.java
+++ b/operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/Versions.java
@@ -1,0 +1,8 @@
+package io.javaoperatorsdk.operator;
+
+public final class Versions {
+
+  public static final String JOSDK = "${project.version}";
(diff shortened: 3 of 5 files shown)


metacosm:
Note that this should probably be integrated with https://github.com/java-operator-sdk/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Version.java, which already provides the SDK version.

andreaTP:
@metacosm I took a look and the scope of `Version` seems to be quite different, as well as the way to achieve it.
Do you have something in mind?

cc. @csviri

metacosm:
> @metacosm I took a look and the scope of `Version` seems to be quite different, as well as the way to achieve it. Do you have something in mind?
> 
> cc. @csviri

What I meant is that all these versions should be accessible via `Version`… We could probably keep the current `Versions` however, it shouldn't be public, imo, and its information accessed via the `Version` API (which is also reused by the Quarkus extension).

andreaTP:
@metacosm I have attempted to follow your comment, unsure if this is what you meant ...

operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/api/config/Versions.java
```diff
@@ -0,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config;
+
+public final class Versions {
+
+  private Versions() {}
+
+  protected static final String JOSDK = "${project.version}";
```
csviri, line 7: Could you pls enlighten me how this supposed to work? Where are those placeholders filled?

operator-framework-core/pom.xml
```diff
@@ -47,6 +47,19 @@
           <commitIdGenerationMode>full</commitIdGenerationMode>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
```
csviri, line 51: Why is this needed?

operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/api/config/Versions.java
andreaTP: This file is filled by the Maven Plugin as it is in `java-templates` folder as opposed to `java`.

operator-framework-core/pom.xml
andreaTP: To fill in the information that is available in Maven.

csviri (approved):
I have no objections, if this helps with the quarkus extension fine by me.

operator-framework-core/src/main/java-templates/io/javaoperatorsdk/operator/api/config/Versions.java
csviri: would it be possible to write a unit test for this?

csviri: great, thx!

metacosm:
@andreaTP could you give me access to push changes to your branch, please? Or would you prefer I open a different PR based on this one with my changes?

andreaTP:
@metacosm as you prefer, you have an invite for my fork or you can open another PR to target the improvements or whatever.
No strong opinions at all! 👍

metacosm:
Didn't easily manage to push changes to the original branch so created a new PR: #1706.

andreaTP:
No worries, thanks @metacosm !

PR #1656, opened 2022-12-12
chore: upgrade to fabric8 client 6.3.0

metacosm:
DO NOT MERGE: this PR is meant to run CI with this fabric8 client version.

PR #1669, opened 2022-12-19
chore: fabric8 client v6.3.1

csviri:


metacosm:
Please close #1656, then. Also note that this probably shouldn't be merged until Quarkus provides a public version that uses the same Fabric8 client.

metacosm:
Re-target to `next` branch. OK to merge there so that we can have a snapshot with it for downstream consumption.

PR #1674, opened 2022-12-21
feat: use github token for minikube

csviri:


csviri:
@metacosm will merge this now, since it's trivial, and fixes some blocking issue.

PR #1681, opened 2022-12-25
fix: remove exception on delete event

csviri:


csviri:
Delete event should always get propagated to `EventProcessor`. See also related issue

csviri:
(will add unit test and/or make this more explicit when back from holidays)

Issue #1687, opened 2022-12-30
Managed BulkDependentResource fails with reconcilePrecondition


grawinkel:
## Bug Report

On Operator SDK 4.2.0 (commit: 061db9c):

I want to implement a Resource

```
public class SomeResourceDependentResource extends CRUDKubernetesDependentResource<X, Y>
    implements BulkDependentResource<X, Y>
{
... desiredResources(...)
... getSecondaryResources(...)
}
```
And use it as a Managed Resource with a reconcilePrecondition

```
@ControllerConfiguration(
    dependents = {
        @Dependent(
            name = "SomeResource",
            type = SomeResourceDependentResource.class
            reconcilePrecondition = SomePrecondition.class
            ),
    })
```

As soon as the reconcilePrecondition is configured, the SomeResourceDependentResource fails, since it is not handled as a BulkDependentResource but as a normal DependentReousource. So the getSecondaryResource() is called, instead the getSecondaryResources(), which leads to errors.

#### What did you expect to see?

Being able to use reconcilePrecondition with BulkDependentResource.

csviri:
Thanks @grawinkel for reporting this, will take a look.

csviri:
> On Operator SDK 4.2.0 (commit: https://github.com/java-operator-sdk/java-operator-sdk/commit/061db9cc2241eb1079924a77a528ce8cd9473f37):

I can't see how this is related tbh.

csviri:
The problem is with condition API, at some point we added there the resource created by the dependent resource, that is clearly not fitting in this case.

https://github.com/java-operator-sdk/java-operator-sdk/blob/d83b9f817fe79c81b4601c41401058b57492b41e/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java#L20-L20

So we probably need to pass here the dependent resource not the resource. So then the user will can decide, which method to call. Will create a issue for that to discuss.

For now it can be handled in the PR, so the resource in the condition will be a Map of resources:

```java
public class SamplePrecondition
    implements Condition<Map<String, ConfigMap>, BulkDependentTestCustomResource> {

  public static final String SKIP_RESOURCE_DATA = "skipThis";

  @Override
  public boolean isMet(BulkDependentTestCustomResource primary, Map<String, ConfigMap> secondary,
      Context<BulkDependentTestCustomResource> context) {
[...]

PR #1688, opened 2023-01-02
fix: condition for bulk resources

csviri:
Note that this is a temp fix for 4.2.  A proper fix with API change will come in 4.3 probably with this PR:
https://github.com/java-operator-sdk/java-operator-sdk/pull/1690

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
@@ -9,12 +9,12 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 
-import io.javaoperatorsdk.operator.processing.dependent.BulkDependentResource;
 import org.slf4j.Logger;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
+import io.javaoperatorsdk.operator.processing.dependent.BulkDependentResource;
 
 @SuppressWarnings("rawtypes")
 public abstract class AbstractWorkflowExecutor<P extends HasMetadata> {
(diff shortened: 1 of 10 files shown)


csviri:
closes #1687

csviri:
The test for some reason does not fail. Will investigate also that.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
```diff
@@ -101,9 +102,15 @@ protected synchronized void handleNodeExecutionFinish(
...
+      return true;
+    }
+    var resources = dependentResource instanceof BulkDependentResource
+        ? ((BulkDependentResource) dependentResource).getSecondaryResources(primary, context)
+        : dependentResource.getSecondaryResource(primary, context).orElse(null);
+
     return condition.map(c -> c.isMet(primary,
-        dependentResource.getSecondaryResource(primary, context).orElse(null),
-        context))
+        (R) resources, context))
```
metacosm, line 113: This shouldn't work because `R` can never be a `Map` in a `DependentResource` declaration so this should most likely cause a `ClassCastException`.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
csviri: That thing does not get into the bytecode, but also there is a test for this in the PR. Will try to make it more nice though.

csviri: Checked but this is still the simplest way to do locally, this will change probably within this issue:
https://github.com/java-operator-sdk/java-operator-sdk/issues/1689

csviri: will improve this sample so it shows more the issue with the Bulk resources.

csviri:
@metacosm improved the test, commented it, would be quite unefficient to do a real life example, but I think here the main point is the API to show in the IT.

metacosm (approved):
LGTM after we make it clearer that this is a temporary fix to be addressed better in #1690

PR #1690, opened 2023-01-02
feat: dependent resource in the condition instead of resource

csviri:


csviri:
Will add migration guide.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/Condition.java
```diff
@@ -10,12 +11,10 @@
...
+   * @param dependentResource for which the condition applies to
    * @param primary the primary resource being considered
-   * @param secondary the secondary resource associated with the specified primary resource or
-   *        {@code null} if no such secondary resource exists (for example because it's been
-   *        deleted)
    * @param context the current reconciliation {@link Context}
    * @return {@code true} if the condition holds, {@code false} otherwise
    */
-  boolean isMet(P primary, R secondary, Context<P> context);
+  boolean isMet(DependentResource<R, P> dependentResource, P primary, Context<P> context);
```
metacosm, line 19: Maybe keep the order that was there before (i.e. primary as first param)?

csviri: Hmm, don't, have a strong opinion on that. Changed this because the condition applies primarily to a dependent resource, so that's why it is on first place (kinda leading or priority)

metacosm:
We should add an example where the dependent resource is actually used because, as it stands, this change only seems to complicate things while not actually addressing any particular issue…

csviri:
> We should add an example where the dependent resource is actually used because, as it stands, this change only seems to complicate things while not actually addressing any particular issue…

when this will be merged: https://github.com/java-operator-sdk/java-operator-sdk/pull/1688
I will adjust that sample.

PR #1699, opened 2023-01-04
feat: provide more exception context on workflow errors

metacosm:
- feat: provide more exception context on workflow errors
- refactor: simplify
- docs: improve wording


csviri (approved):
LGTM

PR #1704, opened 2023-01-05
fix: separated job with special integration tests

csviri:
This fixes also issue with minikube. If started twice in a job, it causes issues.


manusa:
Minikube issues fixed in https://github.com/manusa/actions-setup-minikube/pull/80 pending release

metacosm (approved):

PR #1724, opened 2023-01-19
chore: fabric8 client 6.4

csviri:


csviri:
@metacosm I think we can merge this now, since the client is already in quarkus.

metacosm:
> @metacosm I think we can merge this now, since the client is already in quarkus.

Only in `main`, though. I'm OK merging it with the understanding that it'll be more difficult to release a new version if we also need that version to be used in Quarkus…

csviri:
kk, let's wait until quarkus released.

metacosm:
I think we should merge this now to target Quarkus 3.

metacosm (approved):

PR #1731, opened 2023-01-24
feat: fail-safe event listing

csviri:


csviri:
It's quite a corner case, so decided to not do this.

PR #1734, opened 2023-01-25
fix: do not change namespaces if they were manually set

metacosm:


csviri (commented):
I'm not getting this, `KubernetesDependentResourceConfig` should no be namespace changeable, why is that? namespace changeable should only apply for informers

metacosm:
This was needed at some point: https://github.com/java-operator-sdk/java-operator-sdk/pull/1554/files#diff-6fca1d1644dadba71e06623b843eff387374279a921562c0ea6532d98b53d820R181-R193
A similar problem occurs in the Quarkus extension actually: namespaces can be set on the controller configuration at different times during the build process (i.e. before this even gets to the point where users can do anything) and, if the namespaces haven't been set explicitly on the `@KubernetesDependent` annotation, it might be needed to propagate namespace changes that occur during the runtime phase of the configuration process to the dependent config because they might have recorded a now obsolete version of the namespaces during an earlier phase. We could probably remove the `NamespaceChangeable` interface from `KubernetesDependentResourceConfig` now, though. Still need to be able to change the namespaces in certain situations, though.

csviri:
Ok, somehow I have I feeling its a bit too complex with these configs.
But I guess this is just setting not really changing in the runtime sense. So would remove that, since it's quite confusing there.

csviri (approved):

csviri (approved):

PR #1739, opened 2023-01-27
fix: remove unused cache impl

csviri:


metacosm:
Don't recall what this was used for…

metacosm (approved):

csviri:
At some point this was heavily refactored around the external resource caches. I think we just forgot to delete this.

PR #1744, opened 2023-01-30
Primary to seconday dr informer sample

csviri:


operator-framework/src/test/java/io/javaoperatorsdk/operator/PrimaryToSecondaryDependentIT.java
```diff
@@ -0,0 +1,76 @@
...
+    operator.create(testCustomResource());
+
+    await().pollDelay(Duration.ofMillis(250)).untilAsserted(() -> {
+      assertThat(reconciler.getNumberOfExecutions()).isPositive();
+      assertThat(operator.get(Secret.class, TEST_CR_NAME)).isNull();
+    });
+
+    cm.setData(Map.of(DATA_KEY, TEST_DATA));
+    operator.replace(cm);
+    var executions = reconciler.getNumberOfExecutions();
```
metacosm, line 47: It would make more sense to record the number of executions before updating the CR, though in practice it probably doesn't matter much.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
```diff
@@ -0,0 +1,79 @@
...
+            .byIndex(CONFIG_MAP_INDEX, indexKey(cm.getMetadata().getName(),
+                cm.getMetadata().getNamespace()))
+            .stream().map(ResourceID::fromResource).collect(Collectors.toSet()))
+        .build(),
+        context);
+
+    return Map.of(CONFIG_MAP_EVENT_SOURCE, cmES);
+  }
+
+  private String indexKey(String clusterName, String namespace) {
```
metacosm, line 76: This should be the `configMapName` instead of `clusterName`

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
```diff
@@ -0,0 +1,79 @@
...
+    numberOfExecutions.addAndGet(1);
+    return UpdateControl.noUpdate();
+  }
+
+  public int getNumberOfExecutions() {
+    return numberOfExecutions.get();
+  }
+
+  @Override
+  public Map<String, EventSource> prepareEventSources(
```
metacosm, line 56: This method needs more documentation to explain what is going on. Also, this is weird because usually all the event sources are prepared in this method while here some are managed while others are created manually so this feels like exploiting a loophole in the implementation rather than a properly established mechanism…

metacosm (changes requested):

operator-framework/src/test/java/io/javaoperatorsdk/operator/PrimaryToSecondaryDependentIT.java
csviri: ahh right, changed.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: fixed

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: Yes, will document it. 

So this serves as an example, for a rather corner case; and shows how it is possible to do. As mentioned in the related issue, this is pretty much a corner case where it is questionable if dependent resources should be used, but there is this one case when it make sense to better describe the workflow.

csviri: TBH I'm not sure if we need a nice API for this, or maybe we should keep an eye on this, if pops up more often might make sense.

csviri: added comment.

metacosm:
This should be rebased on `next` since it includes unrelated changes…

docs/documentation/v4-3-migration.md
```diff
@@ -0,0 +1,36 @@
...
+layout: docs
+permalink: /docs/v4-3-migration
+---
+
+# Migrating from v4.2 to v4.3
+
+## Condition API Change
+
+In Workflows the target of the condition was the managed resource itself, not a dependent resource. This changed, from
+not the API contains the dependent resource.
```
metacosm, line 13: I don't understand that part…

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
```diff
@@ -52,6 +52,12 @@ public int getNumberOfExecutions() {
     return numberOfExecutions.get();
   }
 
+  /**
+   * Here just simply creating an Event Source, and setting it for the Dependent Resource. Since it
+   * is not possible to do this setup within the bounds of the KubernetesDependentResource API.
+   * However, this is quite a corner case; might be covered more out of the box in the future if
```
metacosm: Yes, I think we should provide a way for dependents to initialize their event sources if needed in the future.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Creating an Event Source and setting it for the Dependent Resource. Since it is not possible to
+   * do this setup within the bounds of the KubernetesDependentResource API. However, this is quite
+   * a corner case; might be covered more out of the box in the future if there will be demand for
+   * it.
+   **/
+  @Override
+  public Map<String, EventSource> prepareEventSources(
+      EventSourceContext<PrimaryToSecondaryDependentCustomResource> context) {
```
metacosm, line 63: It would be helpful to add comments here as to how the different resources depend on one another and how the mappers make these relations explicit.

metacosm (approved):

metacosm (changes requested):
LGTM apart some more documentation improvements

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: So this is doable also now just not in an elegant way, and did not want to confuse users. One way it would be just to override this method: 

https://github.com/java-operator-sdk/java-operator-sdk/blob/81f7e30ced5bbfe097ca9482e5c7b760ed883320/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java#L47-L50

The SecondaryToPrimary is supported explicitly, but would not be useful to support the PrimaryToSecondary in same way, since for that (as in the example) also access to the context is needed.

We can think about this in the future more, how to cover also this case more elegantly in KubernetesDependentResource, however it's kinda rare case as mentioned. Since not sure if users might want to use dependent resources for these read-only cases.

docs/documentation/v4-3-migration.md
csviri: changed

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: not sure what do you mean, you mean how the secondary depends on primary? and why the mappers needed?

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
metacosm: I would use dependent resources everywhere I can, in a declarative way, so I'm all in favor of making everything possible declaratively (as long as we can do it in an elegant way)… 😄

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
metacosm: Explain how the secret is found from the CR via the mapper and why the mapper is needed, yes, as this would help users figure out how this scenario can (or not) be applied to their own use case.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: The thing is that in this cases this is a ready only resource, not reallly reconciled, and that is what dependent resources primary does, so at least it is smelly here a little bit. 

Tbh I'm not big fan of annotations, but in this would be good to have possibility to do event sources by annotation as we prototyped in the past - to be consistent at least.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/primarytosecondaydependent/PrimaryToSecondaryDependentReconciler.java
csviri: I guess you mean the config map. kk.

csviri: added

metacosm (approved):

PR #1746, opened 2023-01-31
feat: sample of how to do unmodifiable part of dependent

csviri:


metacosm:
All these samples would be even better with a description of what they're supposed to be doing so that people can get an idea before looking at the code.

csviri:
> All these samples would be even better with a description of what they're supposed to be doing so that people can get an idea before looking at the code.

Yep, should be probably a separate issue, either we could generate a summary page, or just link the test from the docs and better describe them there too. 

But kind of an index + generated description would be beneficial.

metacosm (approved):

PR #1749, opened 2023-02-01
fix: improvements on informer start and logging

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
```diff
@@ -82,7 +84,16 @@ public static <T> void executeAndWaitForAllToComplete(Stream<T> stream,
           // restore original name
           thread.setName(name);
         }
-      }).collect(Collectors.toList()));
+      }).collect(Collectors.toList())).forEach(f -> {
+        try {
+          // to find out any exceptions
+          f.get();
```
metacosm, line 90: Isn't this defeating the purpose? I mean this will block each callable in turn…

metacosm: Ah yes, you are right!

metacosm (approved):

PR #1781, opened 2023-02-23
feat: configure resource class and name for controller

csviri:
this is mostly needed so we are able to nicely upgrade the spring boot starter to 4.2.x of JOSDK


csviri:
see https://github.com/java-operator-sdk/operator-framework-spring-boot-starter/pull/32

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -152,6 +156,16 @@ public ControllerConfigurationOverrider<R> withGenericFilter(GenericFilter<R> ge
     return this;
   }
 
+  public ControllerConfigurationOverrider<R> withName(String name) {
+    this.name = name;
+    return this;
+  }
+
+  public ControllerConfigurationOverrider<R> withResourceClass(Class<R> resourceClass) {
+    this.resourceClass = resourceClass;
```
metacosm, line 165: It feels weird to override a given controller configuration with a different resource so I'm not sure about this one…

metacosm: The problem is that it opens the door for improper usage so I'd rather see this support moved somewhere outside of the overrider. The goal of the overrider is to change the configuration provided by the annotations at runtime, not to create a new configuration from scratch. As such, since the resource type is mandated by the parameter type of the `Reconciler` implementation, it shouldn't be possible to override it at runtime so the more I think about it, the more against this change I am.

csviri: Hmm, that is right. The problem is it's not possible to specify it by `@ControllerConfiguration`. Shall we put it rather there? 
Would be a better solution.

metacosm: That would be a huge change and a redundant information with the `Reconciler` implementation.

csviri: Would be empy value. But on the other hand, users might want some hierarchy that the current approach does not detect. I guess that was the reason it was added to the starter at first place.

metacosm: I'm not sure how that would be the case… the `Reconciler` implementation should be parameterised with the resource type. Changing this just for a framework support is not a good reason to change it, especially considering that it should be a pretty rare corner case, imo.

csviri: It's not for framework support, might be the case for anybody use just core. But we can emit this for now and see if somebody asks for it, and discuss the reasons.

csviri: kk, let's remove this option

csviri: revisited this again, and we would break this functionality provided by the users:
https://github.com/java-operator-sdk/operator-framework-spring-boot-starter/blob/182b1111990f99d6516d2965f54dddca8c1cf6de/starter/src/main/java/io/javaoperatorsdk/operator/springboot/starter/OperatorAutoConfiguration.java#L103-L107

if we remove this.

csviri: Basically there is no other way to configure this through our public API, just maybe hacking it with some custom config service, what is problematic. So I think we should allow this to be configured and document it, that this is just for framework support.

metacosm: > Basically there is no other way to configure this through our public API, just maybe hacking it with some custom config service, what is problematic. So I think we should allow this to be configured and document it, that this is just for framework support.

That is not true: the current way to do it is to provide your own `ControllerConfiguration` implementation (or extend the existing `ConfigurationService` implementations) but I do agree that we could make things easier.

metacosm:
Note that this is technically a breaking API change.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -113,8 +113,5 @@ default ConfigurationService getConfigurationService() {
 
   @SuppressWarnings("unchecked")
   @Override
-  default Class<P> getResourceClass() {
-    return (Class<P>) Utils.getFirstTypeArgumentFromSuperClassOrInterface(getClass(),
-        ControllerConfiguration.class);
-  }
+  Class<P> getResourceClass();
```
metacosm, line 116: We should keep a default implementation to avoid breaking the API.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
csviri: Well, yes, althoug not really the API, it's the implementation. But we should remove it in the future, to make it more explicit that this is not used

csviri: It is ok to do ti for next? (I mean a separate PR, wdyt?)

metacosm: No, this method should stay as it's needed by some implementations, Quarkus in particular. And it is breaking the API because implementations that were relying on the provided default implementation won't compile anymore with this change.

csviri: reverted and commented it for now, but I guess we can adjust also quarkus in the future if it will be removed.

For now fine to have it there IMO. 

This PR should be ready now, also the adjusted spring boot part.

metacosm: I don't think that this will change in Quarkus because the resource class is resolved at build time and is therefore fixed then, not resolved at runtime, which is why we need this method to stay as-is.

metacosm (approved):

Issue #1783, opened 2023-02-24
KubernetesDependentResource implementing Matcher causes StackOverflowError


xintellion:
If sub classes of KubernetesDependentResource implement Matcher the matcher field of the class is set to this in the constructor

https://github.com/java-operator-sdk/java-operator-sdk/blob/f1b02f7e8a62242399aed7e8e2b10e948f67e341/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L51

This produces an infinit recursion when the match method is called which throws an StackOverflowError error.
https://github.com/java-operator-sdk/java-operator-sdk/blob/f1b02f7e8a62242399aed7e8e2b10e948f67e341/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L143

I have a standalone reconciler with a manuall worflow configure using operator-framework-spring-boot-starter (4.1.1).

Since the execution of the reconciler is in a separate thread the application keeps running but somehow the Error got totaly swallowed by the try catch block of the NodeExecuter run method. So there is space for improvement in the error handling.

csviri:
We did not see this issue since the subsequent updates, will close.

PR #1786, opened 2023-02-27
fix: make explicit that Updater extends Matcher

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/Updater.java
```diff
@@ -2,10 +2,7 @@
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
-import io.javaoperatorsdk.operator.processing.dependent.Matcher.Result;
 
-public interface Updater<R, P extends HasMetadata> {
+public interface Updater<R, P extends HasMetadata> extends Matcher<R, P> {
```
csviri, line 6: I'm not sure we want to do this. Matcher is something we are just using somewhere in Kubernetes Dependent Resource, where it makes sense to configure different kind of matchers. But that servers completely different purpose.
What purpose it serves having it here?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/Updater.java
metacosm: I understand what you mean but this isn't actually true since `Updater` uses `Matcher.Result` so to be really coherent, we should extract `Result` from `Matcher`. That said, having the `Matcher` interface here doesn't really hurt, imo, and makes it explicit that `Updater` needs to provide that exact same behavior.

csviri: It just confusing, so basically if the abstraction is not used anywhere directly, we should not define that abstraction. Is anywhere this Matcher referenced in the code? Is this useful somewhere?  IMO this is kinda anti-pattern, at least I don't see what purpose it serves. 

>  so to be really coherent, we should extract Result from Matcher

that is the point, it does not need to be coherent.

I think this is the same issue as we do with `LifecycleAware` , it was defined because it was nice, but actually it is never referenced, then it is limiting if something does not fit. But again that abstraction is not explicitly used to abstract away a problem

metacosm: In my opinion, the reverse is actually happening as evidenced by #1783: people think they need to implement `Matcher` when they don't actually don't need to because the current code makes it implicit and difficult to figure out. Making `Updater` extends `Matcher` makes things more explicit. It also solves the issue that `Updater` depends on `Matcher` for no obvious reason in the current code while providing the exact same behavior.

csviri: We can check explicitly also that if implements matcher and throw an exceptio if does.  I think is much clear if we don't compose interfaces with no real reason, and just have the method in update.

csviri: Not saying that having this matcher in Updater would be very bad or something, then just we have to probably create BulkMatcher, and those 2 are API changes actually that we should consider.

metacosm: I don't agree that there is no reason to compose the interfaces: the Updater needs to perform the match behavior, so it makes sense that it extends the Matcher interface.

csviri: I mean no strong opinon if it is done also to bulkmatcher elegantly, and released as part of minor (not patch)

PR #1793, opened 2023-03-02
fix: remove okhttp dependency

metacosm:


csviri (approved):
Just mad one comment. Also not sure if the main pom.xml should be reformatted, but otherwise LGTM

.github/workflows/e2e-test.yml
```diff
@@ -50,6 +50,8 @@ jobs:
       - name: Run integration tests in local mode
         working-directory: ${{ matrix.sample_dir }}
         run: |
+          ls /home/runner/.m2/repository/io/fabric8/kubernetes-httpclient-*
+          mvn dependency:tree
```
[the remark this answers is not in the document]
metacosm: ah no, we can remove this now, just wanted to check what the hell was going on with the dependencies at some point ;)

operator-framework/pom.xml
```diff
@@ -17,7 +17,11 @@
       <groupId>io.javaoperatorsdk</groupId>
       <artifactId>operator-framework-core</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-httpclient-jdk</artifactId>
```
csviri, line 22: Should not be this the vertx client?

metacosm: As mentioned, I would like to have tests with different clients to see if there is any impact

caffeine-bounded-cache-support/pom.xml
```diff
@@ -54,6 +54,11 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-httpclient-okhttp</artifactId>
```
csviri, line 59: Why is okhttp in this module? Probably should be vertx

sample-operators/tomcat-operator/pom.xml
```diff
@@ -38,6 +38,10 @@
             <groupId>io.javaoperatorsdk</groupId>
             <artifactId>operator-framework</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>kubernetes-httpclient-okhttp</artifactId>
```
csviri, line 43: Should not be just okhttp in one of these?

caffeine-bounded-cache-support/pom.xml
metacosm: Let's test with all the different clients in integration tests in a different PR. In this PR, I wanted to have "some" coverage of the different http client implementations without getting into all possible combinations.

csviri (approved):

PR #1805, opened 2023-03-09
feat: remove CR meters when they are deleted (after a delay)

### LINKED ISSUE #1803: Metrics for deleted CRs may lead to OOM when scraping
#### What did you do?

Request to `/metrics` endpoint in Quarkus (using JOSDK w/Quarkus extension) resulted in an OOM error. Further research revealed that JOSDK reconcile metrics accumulate in the Micrometer registry and are not removed for CRs no longer present in the system.

Expect metrics for deleted CRs be removed from the registry after a configurable amount of time, with a default of -1/never.

Metrics for deleted CRs are present until the process ends

#### Environment

**Kubernetes cluster type:**

JOSDK: 3.2.4 (via quarkus-operator-sdk 4.0.7) on Quarkus 2.13.7

`$ java -version`

11.0.x

`$ kubectl version`

```
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4+a34b9e9", GitCommit:"b6d1f054747e9886f61dd85316deac3415e2726f", GitTreeState:"clean", BuildDate:"2023-01-10T15:55:28Z", GoVersion:"go1.19.4", Compiler:"gc", Platform:"linux/amd64"}
```

#### Possible Solution

Clear/remove metrics for deleted CRs after a configurable amount of time, defaulting to "never" for compatibility.

metacosm:
Fixes #1803.

### DIFF
--- a/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
+++ b/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
@@ -32,16 +32,17 @@ public class MicrometerMetrics implements Metrics {
   private final MeterRegistry registry;
   private final Map<String, AtomicInteger> gauges = new ConcurrentHashMap<>();
   private final Map<ResourceID, Set<Meter.Id>> metersPerResource = new ConcurrentHashMap<>();
-  private final ScheduledExecutorService metersCleaner = Executors.newScheduledThreadPool(10);
+  private final ScheduledExecutorService metersCleaner;
   private final int cleanUpDelayInSeconds;
 
   public MicrometerMetrics(MeterRegistry registry) {
-    this(registry, 300);
+    this(registry, 300, Runtime.getRuntime().availableProcessors());
   }
 
-  public MicrometerMetrics(MeterRegistry registry, int cleanUpDelayInSeconds) {
+  public MicrometerMetrics(MeterRegistry registry, int cleanUpDelayInSeconds, int cleaningThreadsNumber) {
     this.registry = registry;
     this.cleanUpDelayInSeconds = cleanUpDelayInSeconds;
+    this.metersCleaner = Executors.newScheduledThreadPool(cleaningThreadsNumber);
   }
 
   @Override
--- a/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
+++ b/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
@@ -1,11 +1,10 @@
 package io.javaoperatorsdk.operator.monitoring.micrometer;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
@@ -17,6 +16,7 @@ import io.javaoperatorsdk.operator.processing.Controller;
 import io.javaoperatorsdk.operator.processing.GroupVersionKind;
 import io.javaoperatorsdk.operator.processing.event.Event;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Timer;
@@ -31,9 +31,17 @@ public class MicrometerMetrics implements Metrics {
   private static final String RECONCILIATIONS_QUEUE_SIZE = PREFIX + RECONCILIATIONS + "queue.size.";
   private final MeterRegistry registry;
   private final Map<String, AtomicInteger> gauges = new ConcurrentHashMap<>();
+  private final Map<ResourceID, Set<Meter.Id>> metersPerResource = new ConcurrentHashMap<>();
+  private final ScheduledExecutorService metersCleaner = Executors.newScheduledThreadPool(10);
+  private final int cleanUpDelayInSeconds;
 
   public MicrometerMetrics(MeterRegistry registry) {
+    this(registry, 300);
+  }
+
+  public MicrometerMetrics(MeterRegistry registry, int cleanUpDelayInSeconds) {
     this.registry = registry;
+    this.cleanUpDelayInSeconds = cleanUpDelayInSeconds;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


csviri:
I guess the target should be the `next` branch for this PR.

metacosm:
Yes

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -31,9 +31,17 @@ public class MicrometerMetrics implements Metrics {
...
+  private final ScheduledExecutorService metersCleaner = Executors.newScheduledThreadPool(10);
+  private final int cleanUpDelayInSeconds;
 
   public MicrometerMetrics(MeterRegistry registry) {
+    this(registry, 300);
+  }
+
+  public MicrometerMetrics(MeterRegistry registry, int cleanUpDelayInSeconds) {
     this.registry = registry;
+    this.cleanUpDelayInSeconds = cleanUpDelayInSeconds;
```
csviri, line 44: Why is there a delay, why cannot be deleted instantly?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -31,9 +31,17 @@ public class MicrometerMetrics implements Metrics {
   private static final String RECONCILIATIONS_QUEUE_SIZE = PREFIX + RECONCILIATIONS + "queue.size.";
   private final MeterRegistry registry;
   private final Map<String, AtomicInteger> gauges = new ConcurrentHashMap<>();
+  private final Map<ResourceID, Set<Meter.Id>> metersPerResource = new ConcurrentHashMap<>();
+  private final ScheduledExecutorService metersCleaner = Executors.newScheduledThreadPool(10);
```
csviri, line 35: Since the clean is a short running task (I assume), wouldn't'  be better to use a Timer?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -116,6 +124,14 @@ public void receivedEvent(Event event, Map<String, Object> metadata) {
   @Override
   public void cleanupDoneFor(ResourceID resourceID, Map<String, Object> metadata) {
     incrementCounter(resourceID, "events.delete", metadata);
+
+    // schedule deletion of meters associated with ResourceID
+    metersCleaner.schedule(() -> {
```
csviri, line 129: Since this is a non trivial logic, we should add unit test.

csviri (changes requested):

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: Because you still might want to get the metrics on the deleted resources for a while before the metrics are removed.

csviri: true, would worth to write it as a javadoc so it is obvious from code.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: Because it's usually better to use `ScheduledExecutorService` than `Timer`. See https://stackoverflow.com/a/409993.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: I'm not sure a unit test makes sense here, though I certainly could add one. What would make more sense would be an integration test to check that the metrics are actually not there anymore, will see how complex that could be to set up.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
csviri: Ok, then the number of Threads should be configurable, and we should use 1 as default.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
csviri: IMO we should tests every non trivial logic with unit tests, and see at least that the API there is called for every ID eventually.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
csviri: Actually the whole ScheduledExecutorService should be configurable.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: The logic is trivial: on delete, call remove on all the meters associated with the resource id. I've added an integration test, anyway.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: Yes, the feature needs to be documented.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: Both delay and thread numbers are configurable. What other aspects would need configuration?

csviri: For example bursting the threads (like the cached executor), but I think it fine for now now

csviri (approved):
LGTM

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -31,9 +32,28 @@ public class MicrometerMetrics implements Metrics {
...
   private final Map<String, AtomicInteger> gauges = new ConcurrentHashMap<>();
+  private final Map<ResourceID, Set<Meter.Id>> metersPerResource = new ConcurrentHashMap<>();
+  private final Cleaner cleaner;
 
   public MicrometerMetrics(MeterRegistry registry) {
+    this(registry, 0);
+  }
+
+  public MicrometerMetrics(MeterRegistry registry, int cleanUpDelayInSeconds) {
+    this(registry, cleanUpDelayInSeconds, 0);
```
csviri, line 43: I think the cleaning thread number should be by default 1, it is basically a very simple operation should not eath up more threads.

csviri (approved):
Made one comment on default otherwise looks good to me

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
metacosm: Makes sense. The default (not obvious from just looking at this line) is actually the number of available processors but will make it 1.

metacosm: Done. Also added javadoc for the constructors.

PR #1825, opened 2023-03-20
docs: readme improvements

csviri:


README.md
```diff
@@ -40,13 +40,13 @@ For all features and their usage see the [related section on the website](https:
...
-
-Quarkus Extension: https://github.com/quarkiverse/quarkus-operator-sdk
-
-Spring Boot Starter: https://github.com/java-operator-sdk/operator-framework-spring-boot-starter
-
-Framework to implement Admission Controllers and Conversion Hooks: https://github.com/java-operator-sdk/kubernetes-webooks-framework
+* Quarkus Extension: https://github.com/quarkiverse/quarkus-operator-sdk
+* Spring Boot Starter: https://github.com/java-operator-sdk/operator-framework-spring-boot-starter
+* jenvtest: https://github.com/java-operator-sdk/jenvtest 
+  Support for integration testing again Kubernetes API Server
```
metacosm, line 53:
```suggestion
  Lightweight Kubernetes API Server, optimized for integration testing in Java
```

README.md
```diff
@@ -5,12 +5,19 @@
 
 # Build Kubernetes Operators in Java Without Hassle 
 
+Java Operator SDK is a framework that makes it easy to implement Kubernetes Operators in Java. 
+
+It provides a controller runtime, support for testing operators, and related tooling. In addition to that implementing
+conversion hooks and dynamic admission controllers are supported as a separate project 
+(and much more,see related projects section).
```
andreaTP, line 12:
```suggestion
(and much more, see related projects section).
```

andreaTP (approved):
A few suggestions but looks good!

README.md
csviri: It's not really a lightweight API Server, it's THE Kubernetes Api server running

README.md
```diff
@@ -5,12 +5,19 @@
 
 # Build Kubernetes Operators in Java Without Hassle 
 
+Java Operator SDK is a framework that makes it easy to implement Kubernetes Operators in Java. 
+
+It provides a controller runtime, support for testing operators, and related tooling. In addition to that implementing
```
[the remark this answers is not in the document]
csviri: I think not here, at least my intention is to have here some one liners, that brings attention to these components. What are the use cases should be under those projects IMO

README.md
metacosm: Fair enough, but I don't think that the current description is very helpful. Why would I want to use this instead of running my tests against a real cluster for example? Your one-liner needs to capture that idea, somehow.

csviri: changed.

Issue #1817, opened 2023-03-14
leader-election support multi cluster


hkmdxlftjf:
#### What did you do?

an instance watch multi-cluster event, but only one election

#### What did you expect to see?

https://github.com/java-operator-sdk/java-operator-sdk/blob/3812a7fe73f81033d87297afd3081a93760a7abd/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java#L76  
can support custom kubernetes-client

csviri:
Hi @hkmdxlftjf , I'm not sure I uderstand what are you asking for, could you describe your use case?

hkmdxlftjf:
we have multi k8s clusters, I want to deploy the operator on a cluster and namespace is xxxx, watch all cluster cr events, 
eg:
`Map<String, Operator> clusterWithOperator = new HashMap<>();`
when isLeaderElectionEnabled is true,  new Operator() will create lease on a different cluster，expect to create a lease only on the deployed cluster, or i can specify the cluster(kubernetes client)

csviri:
Ok, so currently you can do this just not using the leader election capabilities of the JOSDK, and just setup leader election with fabric8 client, and start the operators if becomes the leader.  
The downside is that the non-leader operator does prepares the caches, so when one them becomes the leader it might take some time until the resources are processed on startup on large clusters. 

So if I understand the problem what you asking for is to be able to specify the client for leader election. 
Will have to take a look because this not necessarily will solve the problem. Since you will have multiple operators competing for the lease, even if they have the same ID (in terms of leader election), this might not work. Will have to think about it / check it.

hkmdxlftjf:
can i try add attributein leaderelectionconfiguration(KubernetesClient client), and provide a method to use modify this attribute

csviri:
you are always free to issue a pull request, and we can take a look. But would require a test for this, pls see current leader election tests. 

I will also take a look on the implementation of the leader election, since there might be conceptual issue with this approach. Since in this case it is more like about having multiple leaders instead of one with same `identity` and the leader election was not designed for this.  (But will try the best to support your use case). 

cc @shawkins

csviri:
Will take a look but I think the proper solution here is to share the `LeaderElector` between multiple controllers, not create `LeaderElector` for each `Operator` with the same client. Will take a look how feasible is this to implement.

csviri:
[...]

PR #1828, opened 2023-03-24
add configurable kubernetes client for leader election

hkmdxlftjf:
add configurable kubernetes client for leader election

https://github.com/java-operator-sdk/java-operator-sdk/issues/1817

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -73,7 +73,11 @@ public class Operator implements LifecycleAware {
                 .build();
     ConfigurationServiceProvider.set(configurationService);
     configurationService.getLeaderElectionConfiguration()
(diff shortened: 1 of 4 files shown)


hkmdxlftjf:
#https://github.com/java-operator-sdk/java-operator-sdk/issues/1817

csviri:
@hkmdxlftjf as mentioned in the discussion, and was thinking more since that time, this is not the right way to solve this problem. Since there will be still more instances of operator holding the lock, where it is not the way it was designed, event if now might work it can have issues in future, for example if the leader election implementation changes in fabric8 client - since this is not the case it is tested for.

Will try to take a look and come up with an alternative PR, to share the leader elector, between instances, also discuss it with the fabric8 team more in depth.

csviri:
@hkmdxlftjf I see you closed this, issue, sorry did not have time this week, should I still take a look on this and come up with a solution?

hkmdxlftjf:
Thank you for your attention, I have found another solution for the time being.

PR #1836, opened 2023-03-28
fix: special IT failing test case

csviri:


csviri:
Lowered the timeout. Before I was able to consistently reproduce it locally (with higher timout). This fixed it, also the only explanation from the logs.

PR #1837, opened 2023-03-31
docs: slack badge

csviri:


andreaTP:
How does it work to obtain an account to join?

csviri:
Hmm, this is Kubernetes Slack, yeah that should be somehow described there. It is below, but a join link would be helpful.

csviri:
if you already there it will just open the app / website.

csviri:
will check if i can ad a tooltip or something, little later.

csviri:
added tooltip with invite link. We might add a :grey_question: mark with tooltip also, but this might be enough.

PR #1850, opened 2023-04-05
next version

metacosm:


csviri:
Not sure this should be on the next branch, rather than on an independent PR (maybe for next)

csviri:
Idea is that `next` is also stable and releasable any time.

PR #1856, opened 2023-04-13
fabric8 next version daily check

csviri:
**!! DON'T MERGE !!**

The goal of this PR is to test the framework with the most recent version of fabric8 client.

Will make a daily release of the snapshot with version: `999.F8-SNAPSHOT`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@
     <sonar.host.url>https://sonarcloud.io</sonar.host.url>
     <fabric8-httpclient-impl.name>jdk</fabric8-httpclient-impl.name>
 
-    <junit.version>5.11.3</junit.version>
+    <junit.version>5.11.4</junit.version>
     <fabric8-client.version>7.0.1</fabric8-client.version>
     <slf4j.version>2.0.12</slf4j.version>
     <log4j.version>2.24.3</log4j.version>


.github/workflows/snapshot-releases.yml
```diff
@@ -7,10 +7,13 @@ concurrency:
...
   cancel-in-progress: true
 on:
+  schedule:
+    #   Run on end of the day
+    - cron: '0 0 * * *'
   push:
     paths-ignore:
       - 'docs/**'
-    branches: [ main, v1, v2, v3, next ]
+    branches: [ fabric8-next-version ]
```
metacosm, line 16: I think we should keep releasing snapshots for `main` as well. Ideally, on every merge, even, which would mean splitting this out. Don't know. A snapshot a day is probably enough.

.github/workflows/pr.yml
```diff
@@ -7,6 +7,9 @@ concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
   cancel-in-progress: true
 on:
+  schedule:
+#   Run on end of the day
+    - cron: '0 0 * * *'
```
metacosm, line 12: Why was this added? Is it meant to run a build at least once a day?

csviri: yes, to run a daily build.

.github/workflows/snapshot-releases.yml
csviri: AFAIK it works in a way that, this is not merged, so the github action will for this branch will take into account only the code on this branch. So since this PR never will be merged this should be enough to run the releases daily for this branch, but will work also for main (since that run  uses the code from there)

metacosm:
Why did you close this PR, @csviri?

csviri:
Hmm, I did not, I just updated it for the main, and since there is no snapshot of fabric8 (tried 7.0.2-SNAPSHOT and 7.1.0-SNAPSHOT) I set it for the latest version, so github closed it. Do you know what is the latest snapshot?

metacosm:
Next version is `7.1-SNAPSHOT`.

csviri:
that cannot be found. `Could not find artifact io.fabric8:kubernetes-client-bom:pom:7.1-SNAPSHOT`

metacosm:
It's possible that they haven't been deployed, yet.

csviri:
yeah, I gues so, will recreate this PR with that version.

PR #1858, opened 2023-04-13
chore: jenvtest 0.9.0

csviri:


operator-framework/pom.xml
```diff
@@ -58,12 +58,6 @@
       <groupId>com.google.testing.compile</groupId>
       <artifactId>compile-testing</artifactId>
       <scope>test</scope>
-      <exclusions>
```
metacosm, line 61: Why was this exclusion removed?

csviri: it was included in jenvtest before, but is not anymore.

metacosm (approved):

PR #1873, opened 2023-04-25
fix: logging missing config

csviri:


metacosm (approved):
This was originally done to avoid adding the slf4j dependency to the core module but I guess it's been added for a while.

PR #1882, opened 2023-05-03
improve: jetty log level to info

csviri:
for integration tests this makes issues


metacosm:
Could you elaborate, please?

csviri:
> Could you elaborate, please?

If you take a look into build logs when jetty (but also java http client) used, the logs are polluted with huge amount of debug logs from jetty (600MB), these are useless for us, so this PR will filter them out.

metacosm (approved):

PR #1885, opened 2023-05-03
chore: update to fabric8 client to 6.6.0

metacosm:


csviri:
womdering if this should not target next?

metacosm:
> wondering if this should not target next?

We need a release that uses it soon because it includes fixes for serialization and the vert.x client.

csviri:
> > wondering if this should not target next?
> 
> We need a release that uses it soon because it includes fixes for serialization and the vert.x client.

makes sense, thx

csviri (approved):

PR #1886, opened 2023-05-03
fix: only load Version from properties once

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
```diff
@@ -53,7 +53,7 @@ public BaseConfigurationService(Version version, Cloner cloner) {
   }
 
   public BaseConfigurationService() {
```
csviri, line 55: Actually I'm not sure about this, really try to avoid static stuff.
Also the whole ConfigurationServiceProvider should go away soon (hopefully)

metacosm: Yes, but this is actually a fix until the provider goes away. There is absolutely no reason for the version to be anything but static in any case since it never needs to be reloaded and should not change once the SDK is up.

csviri: So it does not change neither in the current case, since the service is only can be initialized once. (What I achieve is to have static variables, because of the issue we having)

metacosm: It's not true, currently, the properties are reloaded anytime you reset the ConfigurationServiceProvider.

csviri: But the reset is only mean for testing purposes.

metacosm: No, it's used in QOSDK because the static nature of the provider is causing issues when the app restarts in dev mode. So yes, the static provider should be removed because it's causing more issues than helping but in the meantime, I'd like this fix (possibly for a 4.3.2 release soon with Fabric8 6.6.0 as well).

csviri (approved):

PR #1890, opened 2023-05-05
feat: namespace deletion issue

csviri:


csviri:
Actually will convert it to an e2e test, since this would be much nicer with an e2e test.

PR #1911, opened 2023-05-23
feat: use processors for matching as well since usually both are related

metacosm:
- feat: use processors for matching as well since usually both are related
- refactor: remove unneeded parameter that can also be inferred from DR


csviri (approved):

csviri:
Just one note something like this I wanted to do also for 4.4, not sure if we should have a higher level concept for this explicitly. But I think it is ok now, we cant iterate on it for next minor release.

PR #1912, opened 2023-05-23
feat: not start operator if leader elector has no permission to lease object

csviri:


csviri:
will check how elegantly can check other permission (put) on the resources

metacosm (approved):
Don't know enough about leader election to have an informed opinion, so trusting you on this :)

PR #1920, opened 2023-05-29
fix: same short name for multi version crd

csviri:


andreaTP (approved):
LGTM, probably we should make even the CRD generator a little more robust and throw an exception if misalignment is detected.

PR #1924, opened 2023-05-30
improve: Integration test for dependent in different namespace

csviri:


operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/dependentdifferentnamespace/DependentDifferentNamespaceReconciler.java
```diff
@@ -0,0 +1,31 @@
+package io.javaoperatorsdk.operator.sample.dependentdifferentnamespace;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import io.javaoperatorsdk.operator.api.reconciler.*;
+import io.javaoperatorsdk.operator.api.reconciler.dependent.Dependent;
+import io.javaoperatorsdk.operator.support.TestExecutionInfoProvider;
+
+@ControllerConfiguration(
+    dependents = {
```
metacosm, line 10: Perhaps this should be made more explicit by making the reconciler only watch 2 namespaces: the one where the primary resources are deployed and the one where the dependents are supposed to be? It feels like we're somehow "cheating" by having the reconciler watch all namespaces… 😅

csviri: that would require a change to the extension, if you insist can do that, I don't see it that important, on the other hand would be little nicer I agree.

metacosm (approved):

PR #1940, opened 2023-06-13
improve: filters accept super types

csviri:


csviri:
cc @shawkins

shawkins:
LGTM, thanks @csviri

metacosm:
Wondering if this has an impact on backwards compatibility.

csviri:
Probably not, I guess most of the users use just the target resources. Might be more correct to target `next` version probably

shawkins:
> Wondering if this has an impact on backwards compatibility.

There should not be, this is widening, not narrowing what is accepted.  Are you worried about exotic scenarios such as users using reflection to find these methods?

csviri:
(changed it to target `next` )

metacosm:
> > Wondering if this has an impact on backwards compatibility.
> 
> There should not be, this is widening, not narrowing what is accepted. Are you worried about exotic scenarios such as users using reflection to find these methods?

You're right, my concern was unwarranted.

metacosm (approved):

PR #1942, opened 2023-06-14
docs: item store on @ControllerConfiguration

csviri:


metacosm (changes requested):
The documentation is incorrect. This is a simple copy and paste from the related ResourceConfiguration method. While the two methods are similar, they do not have the exact same behavior and therefore deserve appropriate documentation. Also, the links here and in ResourceConfiguration regarding the CaffeineBoundedCache are wrong.

metacosm (changes requested):
Still incorrect.

csviri:
What exactly? @metacosm

metacosm (approved):

PR #1946, opened 2023-06-16
fix: change links after migration to operator framework

csviri:


metacosm (approved):
I think GitHub has a permanent redirect from the old org to the new one so this shouldn't be needed but good to be done anyway.

PR #1951, opened 2023-06-19
highlights the default issue

shawkins:
There are a couple of SSA concerns raised on https://github.com/kubernetes/kubernetes/issues/118725 

The application of server side apply for ports and topologyspreadcontraints may lead to either invalid states, or unexpectedly duplicated entries - these will ideally be corner cases.  At the very least everywhere that in that has multiple listMapKey's, but there should be a doc somewhere containing other issues / problematic behavior.  This may not relate to change detection if there is an independent notion of desired state - but it definitely highlights there are some outcomes of using SSA that are unexpected.

Something else that may be confusing for users (and us) who are used to PUT / replace is the handling of default values - https://github.com/kubernetes/kubernetes/issues/118725#issuecomment-1597574423 This test just highlights that modifications like this will be seen as unmanaged - but that probably does not match the users expectation.  Other than requiring users to make the defaults explicit in the desired state I don't see a great way to approach this.


csviri:
@shawkins  pls compile the code and push it again, that will format it.  
Will take a look on PR tomorrow, thank you!!

shawkins:
> @shawkins pls compile the code and push it again, that will format it. Will take a look on PR tomorrow, thank you!!

To avoid confusion this is not a real pr, it's just to highlight what's happening with what's referenced in the issue.

I needed to correct a couple of things.  It wasn't clear that the desired state should only the managed state, from my converstation with @metacosm I thought it would be the "full state" (the current state from the server / cache with the desired changes overlayed).  It's been changed now so that the desired state is just the managed.

csviri:
> Something else that may be confusing for users (and us) who are used to PUT / replace is the handling of default values - https://github.com/kubernetes/kubernetes/issues/118725#issuecomment-1597574423 This test just highlights that modifications like this will be seen as unmanaged - but that probably does not match the users expectation. Other than requiring users to make the defaults explicit in the desired state I don't see a great way to approach this.

Yes, I think this is a conceptual issue, but if kubernetes defines it this way, what can we do? (other than maybe document it, but even that seems to be more like a k8s docs not JOSDK)

csviri:
> I needed to correct a couple of things. It wasn't clear that the desired state should only the managed state, from my converstation with @metacosm I thought it would be the "full state" (the current state from the server / cache with the desired changes overlayed). It's been changed now so that the desired state is just the managed.

From upcoming release yes, only the desired is managed. Until now we did not use SSA in DR, it just how it happens that k8s creates/updates resource that managedFields are added.

csviri:
> > Something else that may be confusing for users (and us) who are used to PUT / replace is the handling of default values - [kubernetes/kubernetes#118725 (comment)](https://github.com/kubernetes/kubernetes/issues/118725#issuecomment-1597574423) This test just highlights that modifications like this will be seen as unmanaged - but that probably does not match the users expectation. Other than requiring users to make the defaults explicit in the desired state I don't see a great way to approach this.
> 
> Yes, I think this is a conceptual issue, but if kubernetes defines it this way, what can we do? (other than maybe document it, but even that seems to be more like a k8s docs not JOSDK)

but basically if it is not in the desire we can understand it in a way that the user has no opinion on this, so it is an open variable, either way good so can be accepted / matched.

shawkins:
> Yes, I think this is a conceptual issue, but if kubernetes defines it this way, what can we do? (other than maybe document it, but even that seems to be more like a k8s docs not JOSDK)

Yes I want it to be documented both in fabric8 and for the JOSDK - if you are used to replace (PUT) semantics wrt to defaults SSA won't work that way.  For the keycloak pr I had started filling in some of what seemed to be important defaults, but decided it's a subtle distinction that needs more awareness.  

The other issues (again with SSA, not necessarily JOSDK) are:
- ports (Service or Container) - If the key changes the patch will be invalid (either a name will be required, or duplicate entries will be disallowed)
- topologySpreadConstraints - if the key changes you’ll end up with duplicate / contradictory entries
- exlusive situtations - such as a Ingress port name or number - which is represented by separate fields instead of IntOrString, you must try to remove the other field in the apply with an empty string.  If you do not, the SSA may be invalid.  I say try to remove because an empty string isn't always allowed and won't validate

And this is just from the conversion of keycloak.  There will likely be quite a few other situations as we go.

So the question becomes:
- how much responsibility should users take over defaults - or is it simply a problem for whatever may change them in an undersirable way.
- what to do when the SSA of a resource is invalid - is that simply a problem for whatever made the change that's breaking the operator functionality, or is it on the operator to still be able to assert the desired state.  For the keycloak pr I took a half measure of trying a strategic merge patch if the SSA failed - that address the port situtuation (although that handling is questioned by other kube issues).

csviri:
ok so thinking how to manage this, we will just maintain a page in JOSDK docs under the migration section?

shawkins:
> ok so thinking how to manage this, we will just maintain a page in JOSDK docs under the migration section?

The migration guide is the biggest part, so that expectations are set appropriately.  In addition to the situations failing validitation or handling patching differently than apply / strategic merge I was highlighting, they are tracking related situations in https://github.com/kubernetes-sigs/structured-merge-diff/issues/234 

The migration should be clear, or allow the user to choose, what happens should the server side apply be invalid - do nothing and let something else resolve the situation, do a strategic merge, or do an update / replace.

PR #1954, opened 2023-06-20
fix: we need the constructor that takes a ConfigurationService for QOSDK

metacosm:


metacosm:
This commit was somehow dropped from the previous PR: merging it quickly to get a snapshot run for QOSDK.

Issue #1976, opened 2023-07-24
Support pagination on Informer list operation


Bijnagte:
When an Informer gets started if a limit value is supplied instead of trying to fetch all items in a single call the fabric8 client will paginate to get all of the items. This has an impact on large informer caches where timeouts may be experienced or where it may cause memory pressure on the api server when all resources are in memory prior to completing the request.

The default should continue to be to use a non-paginated approach because pagination puts more pressure on etcd than non paginated listing.

We experienced issues with our ability to start informers when the size was large (> 100k resources) but are unsure of the tipping point.

This is more of a feature request than a bug and I will open a PR for this enhancement

PR #1977, opened 2023-07-24
feat: Allow configuration of informer pagination through list limit (…

Bijnagte:
…#1976).

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
@@ -145,4 +145,12 @@ public interface ResourceConfiguration<R extends HasMetadata> {
   default Optional<ItemStore<R>> getItemStore() {
     return Optional.empty();
   }
+
+  /**
+   * The maximum amount of items to return for a single list call when starting an informer. If this
+   * is a not null it will result in paginating for the initial load of the informer cache.
+   */
+  default Optional<Long> getInformerListLimit() {
+    return Optional.empty();
+  }
 }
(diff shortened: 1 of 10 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
```diff
@@ -145,4 +145,12 @@ default Set<String> getEffectiveNamespaces(ConfigurationService configurationSer
   default Optional<ItemStore<R>> getItemStore() {
     return Optional.empty();
   }
+
+  /**
+   * The maximum amount of items to return for a single list call when starting an informer. If this
+   * is a not null it will result in paginating for the initial load of the informer cache.
+   */
+  default Long getInformerLimit() {
```
csviri, line 153: Thinking if we could not find a more specific name. Like `getInformerListLimit` ?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ResourceConfiguration.java
csviri: Otherwise LGTM

csviri, line 153: Can we make this Optional? Also take it into account if not set?

Bijnagte: I updated with your naming selection and added a link to the underlying method that ends up getting since the javadoc  there has more info.

csviri (approved):
LGTM. Thank you @Bijnagte

PR #1980, opened 2023-07-25
feat: add a mapper that is based on owner reference kind

Bijnagte:
We use code similar to this so I figured I would contribute it back upstream. There is a mapper already that will look at owner references but it only handles a single reference and doesn't compare the kind.


csviri:
Thank you @Bijnagte !

PR #1988, opened 2023-07-31
Improvement/handle namespace deletion on finalizer gracefully

Donnerbart:
Fixes https://github.com/operator-framework/java-operator-sdk/issues/1987


csviri:
Hi @Donnerbart , thx for the PR, pls can you run  `mvn clean install` on the code and commit the changes, it will format the code.

csviri:
What does not adds up to me is, if the update failed, how the resources was deleted. In other words if removal of finalizer failed, how is the resource null, after getting it, should not be deleted. But other than that this look reasonable.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -295,7 +294,12 @@ private PostExecutionControl<P> handleCleanup(P resource,
       // cleanup is finished, nothing left to done
       final var finalizerName = configuration().getFinalizerName();
       if (deleteControl.isRemoveFinalizer() && resource.hasFinalizer(finalizerName)) {
-        P customResource = conflictRetryingUpdate(resource, r -> r.removeFinalizer(finalizerName));
+        P customResource = conflictRetryingUpdate(resource, r -> {
+          if (r == null) {
```
csviri, line 298: We should probably add a comment here explaining the situation when this can be null

csviri (changes requested):

csviri:
> Yeah, I tried to reproduce this several times, but couldn't trigger this exception again.
> 
> I think it might be a missing permission that's being removed by deleting the namespace, so the operator fails to remove the finalizer and then cannot retrieve the custom resource again.

Makes, sense but in that case we should rather not retry

csviri:
Probably would be good do a warning log there

Donnerbart:
I agree that this is probably a very rare corner case.

I've added a warning for the `null` case.

csviri (approved):
LGTM, thank you!

PR #1991, opened 2023-08-01
feat: update ClusterRoleBinding with default namespace ( WebPage sample )

yarinkos:
Although operator.yaml is mostly used by the E2E test, it would be nice to have it available for manual tests as well

Manually applying would currently result in the following error:

```
kubectl apply -f k8s/operator.yaml
serviceaccount/webpage-operator unchanged
..
...
The ClusterRoleBinding "operator-admin" is invalid: subjects[0].namespace: Required value

```


csviri (approved):
LGTM

csviri:
thank you @yarinkos

PR #2006, opened 2023-08-08
fix: use update instead of replace in DR

csviri:


csviri:
cc @shawkins

shawkins (commented):
In the SSA case you are explicitly setting the resourceVersion.  Are you expecting the user to have set the resourceVersion?  If not, then this will be unlocked if the resourceVersion is null.

csviri:
> In the SSA case you are explicitly setting the resourceVersion. Are you expecting the user to have set the resourceVersion? If not, then this will be unlocked if the resourceVersion is null.

I don't think the resource version can be null, since we here update a cloned `actual` version (which already has resource version).

shawkins:
> I don't think the resource version can be null, since we here update a cloned actual version (which already has resource version).

Are saying that the user should know to leave the resourceVersion populated on the target they create from the desired method?  And that if they set it as null, that it will be an unlocked update?  Shouldn't SSA work the same way?

My guess was that you are trying to force locking, so I'm proposing https://github.com/operator-framework/java-operator-sdk/pull/2005/files#diff-aa20588ab4b1ff4f171a897d4042d1b055c02737b067d09aaeb9cfd770adf3a0R131

csviri:
https://github.com/java-operator-sdk/java-operator-sdk/blob/b99a8b7c32001c3aaeca740709472241d3918605/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java#L42-L47

In generic case, here the logic works that we clone the actual resource (from cache), replacing the spec and annotations ans labels. So the resource version is always present in the result.

csviri:
> Shouldn't SSA work the same way?

no since that is not aware of the current state. Normally there should not be optimistick locking for SSA, just do it because of event recording, althoug will check but we might able to live without that too.

shawkins:
> In generic case, here the logic works that we clone the actual resource (from cache), replacing the spec and annotations ans labels. So the resource version is always present in the result.

Sorry I hadn't realized there was that additional complexity.  So basically every resource that lacks a spec, or ones that you want to manipulate a subresource, you have create a matcher for.  And if new mutable fields are added, they must also be added to the matcher.

> no since that is not aware of the current state. Normally there should not be optimistick locking for SSA, just do it because of event recording, althoug will check but we might able to live without that too.

That's not what I was thinking, my comment was based upon the possiblity that the resourceVersion in the non-SSA case could be null.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -154,7 +154,7 @@ public R update(R actual, R target, P primary, Context<P> context) {
           .forceConflicts().serverSideApply();
     } else {
       var updatedActual = updaterMatcher.updateResource(actual, target, context);
-      updatedResource = prepare(updatedActual, primary, "Updating").replace();
+      updatedResource = prepare(updatedActual, primary, "Updating").update();
```
shawkins, line 157: This may need to be an explicitly locked replace, or a patch.  One subtle difference between update and replace is that replace does some modifications to the resource (Services, Jobs, and OpenShift RoleBindings) based upon the present state - see HasMetadataOperation.modifyItemForReplaceOrPatch.  The intention is to remove that once replace is gone.

csviri: This does not makes up to me, we such changes are done for the resources:

https://github.com/fabric8io/kubernetes-client/blob/0c7d5150702387c1aeca66facb98508d590934f2/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/batch/v1/JobOperationsImpl.java#L163-L175

Shoudn't be this the responsibility of the user to fill those values? 

I don't see why should be this patch or replace instead of update because of this.

metacosm: Yes, it's not clear why these resources have a special treatment to me either…

shawkins: The root issue is that PUT has side effects.  Service is the poster child for this - if you attempt a PUT and the clusterIP is not populated, then it will be allocated, which will then conflict with the existing one.  If you use an empty string it will complain that the field is immutable - people have complained about this for years https://github.com/kubernetes/kubernetes/issues/91459  So I guess that in the past they wanted to smooth this behavior out in the fabric8 client.

In the last couple of years when users complain of new situations like this that don't work with createOrReplace we have been telling to manually do something like the proposed createOr, or more recently to use serverSideApply.

metacosm: Yeah, things can get messy fast when you get into discussion of HTTP verbs semantics :)
I do agree with one of the commenters that PUT should be idempotent so regardless of what controllers do, if they accepted one resource as valid at one point in time, they should accept that same resource again if re-PUT (and possibly return the existing one), which doesn't appear to be the case here…

csviri: > The root issue is that PUT has side effects. Service is the poster child for this - if you attempt a PUT and the clusterIP is not populated, then it will be allocated, which will then conflict with the existing one. If you use an empty string it will complain that the field is immutable - people have complained about this for years https://github.com/kubernetes/kubernetes/issues/91459 So I guess that in the past they wanted to smooth this behavior out in the fabric8 client.
In the last couple of years when users complain of new situations like this that don't work with createOrReplace we have been telling them to manually do something like the proposed createOr, or more recently to use serverSideApply.


Yeah, I think this workaround for example can nice used:
`First load the existing service that contains the current clusterIP. Set the old clusterIp to the updated V1Service.`

For these special cases would rather prepare some default implementations in dependent resources, rather than solving it on client level here. So would anyways stick with the `update`.

metacosm (approved):

PR #2008, opened 2023-08-09
improve: remove optimistic locking from update for SSA in dependent resources

csviri:


shawkins:
With #2012 optimistic locking is not required for either SSA or non-SSA - but it's still a good idea with non-SSA.

The rationale here is that nearly all SSA reconstructive operations don't care about what others have done.  Only if you want to take over management fo those changes do you need the absolute freshest state.

There are other corner cases with SSA, such as needing to change a immutable field on a deployment, or trying to overwrite a non-managed change that confuses SSA, such as something modifying port protocol - but those should be addressed with workflow mechanisms and shouldn't depend on locking.  At worst a SSA may temporarily fail.

csviri:
Yes, completely agree.

PR #2012, opened 2023-08-10
changes to simplify the recording mechanism

shawkins:
Retargeting of #2005


shawkins:
To move this out of draft there are a couple of things to cover:

- should the feature of adding the annotation be enabled/disabled via flag? If it's turned off, that just means we won't skip the event if it comes earlier than the update response. As mentioned above that may not be a big deal - operators that are working off of caches and that have effective matchers will just use a few extra cpu cycles.
- Is there a better way to do filtering of the annotation rather than adding special case logic in the matcher - and does the annotation need to be accountted for in the non-SSA case as well?
- what tests to include - integration? or will some unit tests suffice

metacosm:
Can we rebase this so that only the relevant changes show up in the diffs, please?

csviri:
> should the feature of adding the annotation be enabled/disabled via flag? If it's turned off, that just means we won't skip the event if it comes earlier than the update response. As mentioned above that may not be a big deal - operators that are working off of caches and that have effective matchers will just use a few extra cpu cycles.

Good question, I think that could be there but we can add it a separate pull request, would we add separate one for caching and one for event propagation?

csviri:
> what tests to include - integration? or will some unit tests suffice

I think there should be related IT which we might to update to check if there are annotations present, and unit tests yes.

shawkins:
> Good question, I think that could be there but we can add it a separate pull request, would we add separate one for caching and one for event propagation?

We have a couple of inter-related features here.

* skipping events if they come in too quickly via the annotation
* Not currently implemented, but mentioned elsewhere, maintaining a bounded cache of resource versions the dependent resource was responsible and skipping those events if they arrive late (after you have already operated on the temporary cache item).
* working off of a more updated cache (which without resource version parsing works in a limited fashion)

The relevant options could be:
- a flag for using the annotation
- a flag to parse resource versions
- a flag to disable all cache / event optimizations

shawkins:
Ok, this should be considered out of draft.  The integration test should work again based upon using a dependent resource instead - although it may look too hackish.

csviri (approved):
LGTM

metacosm (changes requested):
I have more changes that I think could clean some things up but I'd need access to push to your branch. Also, I think we need some kind of documentation on this change, in particular regarding the logic used to detect whether a change has already been processed.
More generally speaking, it seems like this behavior is only limited to Kubernetes dependents. Is that really true? What happens for informer-based events which are not coming from dependents?

shawkins:
> I have more changes that I think could clean some things up but I'd need access to push to your branch.

Maintainer access was checked.

> Also, I think we need some kind of documentation on this change, in particular regarding the logic used to detect whether a change has already been processed.

In rough terms the annotation is being used to detect when you recieve an event ahead of when your blocking create/update call returns.  If you have already moved ahead to a temporarily cached resource, or anything about your local state does not still match what is in the annotation value, we'll assume the event is needed.  @csviri has suggested using another issue / pr to use a flag for this optimization.  If disabled and the matching logic isn't being tripped up, then it's not a big deal to initiate another reconciliation cycle.  I'm not sure what a good place is to document this, or if it should just wait for the issue that makes the feature conditional.

> More generally speaking, it seems like this behavior is only limited to Kubernetes dependents. Is that really true? What happens for informer-based events which are not coming from dependents?

There are two things working together here, the KubernetesDependentResource and the InformerEventSource.  That is really no different than before - only the KubernetesDependentResource had the calls to event filtering logic.  Are you asking about the case of someone creating their own dependent like resource using the methods like InformerEventSource.prepareForCreateOrUpdateEventFiltering?  You could still expose something like that here, you would just need different signatures as they would need to be responsible for adding the annotation and so need access to the previous revision.

> Is there a better way to do filtering of the annotation rather than adding special case logic in the matcher - and does the annotation need to be accountted for in the non-SSA case as well?

Answering my own previous question - if someone were using a custom matcher and expecting labelAndAnnotation equality, then yes this new annotation would be a problem.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -72,17 +73,16 @@
...
   private static final Logger log = LoggerFactory.getLogger(InformerEventSource.class);
 
   private final InformerConfiguration<R> configuration;
-  // always called from a synchronized method
-  private final EventRecorder<R> eventRecorder = new EventRecorder<>();
   // we need direct control for the indexer to propagate the just update resource also to the index
   private final PrimaryToSecondaryIndex<R> primaryToSecondaryIndex;
   private final PrimaryToSecondaryMapper<P> primaryToSecondaryMapper;
   private Map<String, Function<R, List<String>>> indexerBuffer = new HashMap<>();
+  private String id = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
shawkins: I think my initial thought was to allow setting of the id for tider unit tests, but since that was never added it seems fine to make it final.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -193,11 +180,27 @@ public Result<R> match(R actualResource, R desired, P primary, Context<P> contex
     }
   }
 
+  private void copySDKAnnotations(R actualResource, final R desired) {
+    String actual = actualResource.getMetadata().getAnnotations()
+        .get(InformerEventSource.PREVIOUS_ANNOTATION_KEY);
+    Map<String, String> annotations = desired.getMetadata().getAnnotations();
+    if (actual != null) {
+      if (annotations == null) {
```
metacosm, line 188: I don't think that we need that part as the client always make sure the annotations exist, if I'm not mistaken. Unless, of course, we want to safeguard here, regardless of what the client does.

shawkins: The client will initialize the annotation, but the desired is coming from the user logic and could have been set to null.  I was just trying to match a similar guard that is used in addDefaultSecondaryToPrimaryMapperAnnotations - but I'd be fine if the assumption was that the user must leave the annotations as non-null and remove the initialization from both of those places.

Also it may not need to be part of these changes, but it's making things more complicated than needed to diffentiate the ssa and non-ssa cases wrt methods like addReferenceHandlingMetadata - I think this should be refactored so that the update / match logic can read more like create so that we'll take the same pre-steps in both cases.

shawkins: @metacosm see what you think of latest commit

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -217,13 +212,7 @@ protected Resource<R> prepare(R desired, P primary, String actionName) {
         desired.getClass(),
         ResourceID.fromResource(desired));
 
-    addReferenceHandlingMetadata(desired, primary);
-
-    if (desired instanceof Namespaced) {
-      return client.resource(desired).inNamespace(desired.getMetadata().getNamespace());
```
shawkins, line 220: @csviri @metacosm this is only needed if a NamespacedKubernetesClient that has had inNamespace called is used as the client.  That shouldn't be the case correct?

shawkins: As long as that namespace is already populated on the resource, then that is what the client will use.  The namespacing cleanup in the client didn't happen until 6.0, so you may have had an issue before that.

csviri: yep, ok great!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -152,51 +129,72 @@ public R update(R actual, R target, P primary, Context<P> context) {
...
     R updatedResource;
+    addMetadata(false, actual, target, primary);
     if (useSSA(context)) {
       updatedResource = prepare(target, primary, "Updating")
           .fieldManager(context.getControllerConfiguration().fieldManager())
           .forceConflicts().serverSideApply();
     } else {
       var updatedActual = updaterMatcher.updateResource(actual, target, context);
-      updatedResource = prepare(updatedActual, primary, "Updating").update();
+      updatedResource = prepare(updatedActual, primary, "Updating").replace();
```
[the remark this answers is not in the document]
shawkins: Actually I think it was just confusing for the rebase because my changes included toggling this from replace to update, then back to replace.  Once you rebased next your change from replace to update logically came before my changing it back to replace. 

I'll start squashing my changes before rebasing to prevent that sort of thing.

csviri (approved):
LGTM

csviri:
can we merge this?  @shawkins @metacosm

shawkins:
> can we merge this?

I thinks so.  There will be a follow-up task to allow disabling it.

PR #2014, opened 2023-08-11
generalization of the update matcher

shawkins:
This generalizes the update and match logic for fields other than spec and metadata.  This should make the usage of specific subclasses unnecessary.


shawkins:
I should mention that both the current logic and this logic have problems with matching stuff in the desired state that ends up being pruned.  This pr does make that problem worse - if the desired state for example includes the status subresource that will be seen as a difference is it doesn't match exactly with the actual.  But I don't think users should generally be doing this.

shawkins:
@metacosm @csviri there are two parts here

- the first is to generalize the update logic.  I did that initially with a custom mapper, but that complexity isn't needed we can work just over the regular map forms.  We'll basically replace any top-level field in the actual with what is in the desired with special handling for the metadata.  This is similar to the Jackson updateValue method, which we could consider exposing.

- the second is to generalize the matching.  Based upon all of the existing subclasses it makes the assumption that you want equality for anything on the desired that is non-spec field.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
```diff
@@ -24,6 +25,9 @@ public class GenericKubernetesResourceMatcher<R extends HasMetadata, P extends H
   private static final String OP = "op";
   public static final String METADATA_LABELS = "/metadata/labels";
   public static final String METADATA_ANNOTATIONS = "/metadata/annotations";
+  // without knowing the CRD we cannot ignore status as it may not be a subresource, so if it's
```
csviri, line 28: can you give an example when we cannot ignore status?

csviri: (I mean by definition does not have to be a sub-resource but I guess that usually is just a bug, or are there any real life examples?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
```diff
@@ -2,48 +2,40 @@
...
   }
 
+  @SuppressWarnings("unchecked")
+  @Override
   public R updateResource(R actual, R desired, Context<?> context) {
-    var clonedActual = context.getControllerConfiguration().getConfigurationService()
-        .getResourceCloner().clone(actual);
+    KubernetesSerialization kubernetesSerialization =
+        context.getClient().getKubernetesSerialization();
+    Map<String, Object> actualMap = kubernetesSerialization.convertValue(actual, Map.class);
```
csviri, line 30: Could we add unit test for this please?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
```diff
@@ -208,25 +212,38 @@ private static <R extends HasMetadata> boolean matchSpec(R actual, R desired, bo
     final List<String> ignoreList =
         ignoredPaths != null && ignoredPaths.length > 0 ? Arrays.asList(ignoredPaths)
             : Collections.emptyList();
-    // reflection will be replaced by this:
-    // https://github.com/fabric8io/kubernetes-client/issues/3816
-    var specDiffJsonPatch = getDiffsImpactingPathsWithPrefixes(wholeDiffJsonPatch, SPEC);
+    boolean specMatch = match(equality, wholeDiffJsonPatch, ignoreList, SPEC);
+    if (!specMatch) {
```
csviri, line 216: Please add unit tests at least for some of the special cases we were handling before?

csviri (changes requested):
Looks, good just some comments, and missing unit tests :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
shawkins: The only reference I have for this was when there was a change in fabric8 a while ago that made the assumption we could treat status as a subresource, it was brought up then that some users had legacy crs with non-subresource status fields.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
shawkins: All lines are already covered by GenericResourceUpdaterMatcherTest - are there some specific cases you are concerned about?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
csviri: I would threat those as exceptions. We can tell them, to make their custom marcher for that case.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
csviri: I meant also for the speacial cases, tole RoleBinging, ServiceAccount etc. At least for some of those which were deleted.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
shawkins: Same applies here GenericKubernetesResourceMatcherTest covers all of the modifications.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
shawkins: > At least for some of those which were deleted.

Secret is covered by the test.  I'll add something for ServiceAccount.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
```diff
@@ -24,6 +23,9 @@ public class GenericKubernetesResourceMatcher<R extends HasMetadata, P extends H
   private static final String OP = "op";
   public static final String METADATA_LABELS = "/metadata/labels";
   public static final String METADATA_ANNOTATIONS = "/metadata/annotations";
+  // without knowing the CRD we cannot ignore status as it may not be a subresource, so if it's
+  // included we expect it to match
+  private static Set<String> NOT_OTHER_FIELDS = Set.of(SPEC, "/metadata", "/apiVersion", "/kind");
```
csviri, line 28: Why SPEC is here? we want to match SPEC, or? I'm reading it wrong

csviri: > spec and metadata have their own handling, so that's not covered by the "other fields" matching.

sorry totally missed that.

shawkins: Yet another tweak would be to change the logic to just walk through the diffs once and do prefix matching to determine if the change is allowable.

csviri: > You would only want to do that if you suspect that users are including non-matching status that is not meaningful on their dependent resources. My guess would be that they are not including status, and if they are it would be because it's coming from the actual state (and would match), or that it's not a subresource and needs to be considered in matching. But if you feel strongly about it, it can be excluded.

So these are resources which usually created, but not managed by the controller, so in normal case we apply the desired state using SSA, that is usually the spec. We don't care about the status at all at this phase, we might case in a postCondition, or when we are setting the status for the primary resource.

shawkins: >  We don't care about the status at all at this phase,

If you assume it's a subresource.  If not, then you do.

csviri: yep, true, but again that should not be a feature flag? we are comparing the desired with the actual, the actual would contain the status even if it is a subresource, but the desired won't, wouldn't this always fail to match in that case?

shawkins: > we are comparing the desired with the actual, the actual would contain the status even if it is a subresource, but the desired won't, wouldn't this always fail to match in that case?

Ah, the issue you are raising is with the true parameter with https://github.com/operator-framework/java-operator-sdk/pull/2014/files#diff-3425845ceca0cbb05292b87cdceec9960b4342cefac82c18ab929a5ef22e5288R211 - if someone leaves the status unpopulated in the desired (which is expected) the exact match here will cause a problem.  I must have been implicitly assuming that all are add ops would apply here, but that would obviously require a separate case.  Let me make another revision and add a test.

shawkins: Further refactored to clarify the matching algorithm and added status as excluded.

csviri:
> One last thought is whether you want something for the loss of the protected method updateClonedActual - a release note? Bring back a similar method, or just let people override updateResource after calling the super method?

I think overriding `updateResource` is good enough for now, will see about feedback, and will make it easier if there will be a lot of buzz around it.

csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
```diff
@@ -2,48 +2,40 @@
...
+    return (ResourceUpdaterMatcher<R>) INSTANCE;
   }
 
+  @SuppressWarnings("unchecked")
+  @Override
   public R updateResource(R actual, R desired, Context<?> context) {
-    var clonedActual = context.getControllerConfiguration().getConfigurationService()
-        .getResourceCloner().clone(actual);
+    KubernetesSerialization kubernetesSerialization =
+        context.getClient().getKubernetesSerialization();
```
metacosm, line 29: I'm wondering if this is opening potential issues where the cloning occurs in a different manner than here if, for some reason, the configured resource cloner has a different serialization set up… Should we remove the cloner altogether?

metacosm (approved):
LGTM apart from one comment.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/updatermatcher/GenericResourceUpdaterMatcher.java
shawkins: If this was the only instance of the cloner being used, it could be removed or deprecated.  There is also a clone method on KubernetesSerialization.

Issue #1934, opened 2023-06-07
`stopOnInformerErrorDuringStartup` should not block application startup


dpozinen:
#### What did you do?

set `stopOnInformerErrorDuringStartup` to `false`

the doc states that 

> if false, the startup will ignore recoverable errors, caused for example by RBAC issues, and will try to reconnect periodically in the background.

#### What did you expect to see?

The Spring Boot application would continue with startup, but attempt to start the `Operator` in the background

#### What did you see instead? Under which circumstances?

The Spring Boot application is prevented from starting while attempting to start the `Operator`.

#### Possible Solution

Likely some executor is not configured to respect this property.

csviri:
So this seems to be not an issue, rather a feature, if you take a look on this PR:

https://github.com/operator-framework/java-operator-sdk/pull/2024

What happens is that the operator eventually starts, but there is by default a quite hight timeout.

see this part of code:

https://github.com/java-operator-sdk/java-operator-sdk/blob/1691cf0b3e5df80749467aaa1a904221b2211ac6/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java#L81-L90

Basically it is hard to differentiate during the startup if it is because access right or just syncing is taking long. You can configure this timeout as seen in the PR.

Will discuss this with fabric8 team if we can improve on this.

PR #2024, opened 2023-08-22
improve: add test for startup with no right and no stop config

csviri:


csviri:
closes #1934

metacosm (approved):
Adding some documentation on the purpose of this test would be useful.

csviri:
> Adding some documentation on the purpose of this test would be useful.

true, will add.

PR #2026, opened 2023-08-23
improve: dependent resource can be re-initialized

csviri:
see linked issue

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -365,7 +365,7 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
     super.setConfigurationService(configurationService);
 
     cache.addIndexers(indexerBuffer);
-    indexerBuffer = null;
+    indexerBuffer = new HashMap<>(0);
   }
 
   public void addIndexers(Map<String, Function<R, List<String>>> indexers) {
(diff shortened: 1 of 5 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -365,7 +365,7 @@ public void setConfigurationService(ConfigurationService configurationService) {
     super.setConfigurationService(configurationService);
 
     cache.addIndexers(indexerBuffer);
-    indexerBuffer = null;
+    indexerBuffer = new HashMap<>(0);
```
metacosm, line 368: I don't think that initializing the buffer to a zero size is a good idea since it will immediately need to get resized the first time something is added to it. Can't we initialize it to some default, reasonable size if we want to avoid overprovisioning for nothing? I'm not sure what's the typical use pattern here, like how many indexers might we expect to have?

shawkins: We mostly shouldn't worry about initializing to specific sizes.  The value is used lazily - so using zero or one is effectively the same.  If not specified it will default to 16, but you're only holding that memory ephemerally based upon the usage pattern here.

metacosm: I guess 0 leads to the internal table being initialized to size 1… which is reasonable if we expect a low number of indexer. No idea on the usage pattern.

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: Usually there is index just for one case (at least this is what we see now) when there is many-to-one / many-to-many  relationship between resource, see: https://javaoperatorsdk.io/docs/features#managing-relation-between-primary-and-secondary-resources. But yeah, maybe not necessary to optimizations like this, can be even confusing to read. will remove this number if not objections and have the default (as when initialized).

metacosm (approved):

PR #2028, opened 2023-08-23
Refinements to SSA matching

shawkins:
This highlights the state observed with https://github.com/kubernetes/kubernetes/issues/118519 

Note that the desired Secret uses stringData and the actual state after SSA references stringData as managed, but the resource is actually using data instead.

The matcher will currently throw an NPE when this occurs, so there's a minor change to allow it to continue - but it still of course sees this as a difference.

We know this occurs with Ingress as well - an empty path was being used in Keycloak because that is recommended in the openshift documentation when using the ImplementationSpecific pathType.  If it's left in the spec, it gets pruned and that causes the matching logic to think there has been a change.  However it does not seem necessary to use an empty path and so we removed that.

Some thoughts on how to handle this:
- It is likely an api server bug, but it doesn't look like it will be resolved any time soon.
- This situation, or anything involving pruning / modification of the apiServer state vs what's in the managed fields is problematic, and it won't be understood by the generic matcher either.
- You could consider some type specific canonicalization logic (Secret conversion of stringData to data for example) - but automatically handling empty strings will likely be more problematic.
- recommend or automatically install something like the MetadataAwareOnUpdateFilter when SSA is being used.  Even if the update is issued this will filter out the update event because there will be no real changes in the new vs the old cache state.  Of course this logic is extremely close to what's in the generic update matcher (just that status would be meaningful also) - so there is potential for reuse.


csviri:
> The matcher will currently throw an NPE when this occurs, so there's a minor change to allow it to continue - but it still of course sees this as a difference.

Not sure if there is a good solution for this, except to handle this as a specific, case, and provide an implementation accordingly.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates and other contributors as indicated by
```
csviri, line 2: I guess this copyright part can be removed

shawkins: Yes, sorry just copied this from what was in keycloak.

csviri:
I think this needs a rebase too :)

shawkins:
@csviri In double checking the statefulset and ingress issues that were highlighted previous in kubernetes issues, they do seem to be resolved with newer kubernetes versions wrt creating additional resource versions (they will still have match issues though).  

I've refined this pr to do a couple of things:
- localizes sanitation to ssa matcher.  If possible it doesn't seem like the josdk should be explicitly setting fields in the spec that aren't part of the users desired state.
- further ensures values won't be logged
- will detect if a non-meaningful revision has been created

csviri (changes requested):
Looks good to me, thank you!  
except the missing tests. Could you add at least some unit test for the filter. 

(We can think also some additional integration tests and feature flags, but that can be a separate PR

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
```diff
@@ -0,0 +1,85 @@
...
+          + "If the operator sdk was responsible for this update, this is a situation not handled well by SSA.",
+          newResource.getKind(), Cache.metaNamespaceKeyFunc(newResource));
+      return false;
+    }
+    return true;
+  }
+
+  private ObjectMeta minimizeObjectMeta(ObjectMeta newMetadata) {
+    return new ObjectMetaBuilder(newMetadata).withManagedFields().withResourceVersion(null)
+        .withUid(null).build();
```
csviri, line 82: Why UID is null?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -92,6 +97,35 @@ public boolean matches(R actual, R desired, Context<?> context) {
     return prunedActual.equals(desiredMap);
   }
 
+  /**
+   * Correct for known issue with SSA
+   */
+  @SuppressWarnings("unchecked")
+  private void sanitizeState(R actual, R desired, Map<String, Object> actualMap) {
```
csviri, line 104: Actually I'm not sure this is better here, so for this case we always need the filter.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
```diff
@@ -0,0 +1,85 @@
...
+      log.warn("Ignoring revision with no actual changes for resource type {}, {}. "
+          + "If the operator sdk was responsible for this update, this is a situation not handled well by SSA.",
+          newResource.getKind(), Cache.metaNamespaceKeyFunc(newResource));
+      return false;
+    }
+    return true;
+  }
+
+  private ObjectMeta minimizeObjectMeta(ObjectMeta newMetadata) {
+    return new ObjectMetaBuilder(newMetadata).withManagedFields().withResourceVersion(null)
```
csviri, line 81: Do I understand this correctly that the resourceVersion is same in this case?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
```diff
@@ -0,0 +1,85 @@
...
+    // since the hierarchy of model may not implement hashCode/equals, we'll convert to a generic
+    // form
+    // that should be less expensive than full serialization
+    var newNode = kubernetesSerialization.convertValue(newResource, ObjectNode.class);
+    newNode.remove("metadata");
+    var oldNode = kubernetesSerialization.convertValue(oldResource, ObjectNode.class);
+    oldNode.remove("metadata");
+    if (newNode.equals(oldNode)) {
+      // could check the previous annotation to see if it's our update
+      log.warn("Ignoring revision with no actual changes for resource type {}, {}. "
```
metacosm, line 72: How often do we anticipate this situation to occur? Shouldn't this be logged at the debug level rather?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * addresses the additional events noted on
+ * https://github.com/fabric8io/kubernetes-client/issues/5215
+ * <p>
+ * Several of these circumstances seem to have been addressed by ~1.25+
+ */
+public class MetadataAwareOnUpdateFilter<T extends HasMetadata> implements OnUpdateFilter<T> {
```
csviri, line 36: Actually would be nice to have an IT for this. That shows the problem, and it is solved by this.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
shawkins: For anything other than a resync event new and old should have different resourceVersions.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
shawkins: I've retested the issue seen with Secrets, Ingress, and StatefulSets on 1.26.3 - they no longer exhibit the behavior of creating an additional version.  The upstream statefulset issue mentions a fix in 1.25.  So this is really a fix to ensure there's no regressions or to support older Kube versions with SSA.  Alternatively we could just document this rather than introducing this class.

Secrets and Ingress (with empty strings) though will still have matching issues, but this is a lesser problem as we're already excluding secrets from SSA and without the additional revision there won't be the loop behavior.

csviri: Ahh, in that case, yes, it might be just enough to document it IMO.

csviri: But I guess the move of the sanitization still makes sense.

shawkins: Ok, narrowed this PR to not include the detection of meaningless additional revisions.

csviri (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/MetadataAwareOnUpdateFilter.java
csviri: thx, fine to merge this from my point

metacosm (changes requested):

metacosm:
@shawkins you need to sign your commits off with -s (or equivalent) now
A quick way to do so for all your commits is to do a `git rebase --signoff HEAD~N` and then force push.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -192,11 +189,11 @@ protected void addMetadata(boolean forMatch, R actualResource, final R target, P
     addReferenceHandlingMetadata(target, primary);
   }
 
-  protected void sanitizeDesired(R desired, R actual, P primary, Context<P> context) {
-    DesiredResourceSanitizer.sanitizeDesired(desired, actual, primary, context, useSSA(context));
-  }
-
   protected boolean useSSA(Context<P> context) {
+    // SSA by-passes the updateMatcher, since the use has customized don't use SSA
```
metacosm, line 193:
```suggestion
    // SSA bypasses the `ResourceUpdaterMatcher` so don't use SSA when the user is using a custom one
```

shawkins:
> @shawkins you need to sign your commits off with -s (or equivalent) now

Yes, I've seen that.  For some reason that box won't stay checked by default in eclipse.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
shawkins: switched instead to an explicit field

csviri (approved):

metacosm (approved):

shawkins:
@metacosm @csviri rebased

csviri (approved):

csviri:
after rebase, we can merge also this imo

shawkins:
> after rebase, we can merge also this imo

Again?  Is next being rebased against main frequently?

PR #2034, opened 2023-08-25
eventing refinements mentioned on #2012

shawkins:
provides two options
- to control if the annotation is used (to omit events that come too quickly)
- to parse the resource version (to keep the cache up-to-date and omit events if they come too slowly)

Looking for feedback on if these are the right places to introduce the configuration.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -365,20 +365,20 @@ public interface ConfigurationService {
   }
 
   /**
-   * If an annotation should be used so that the operator sdk can detect events from its own updates
-   * of dependent resources and then filter them.
+   * If a javaoperatorsdk.io/previous annotation should be used so that the operator sdk can detect
+   * events from its own updates of dependent resources and then filter them.
    * <p>
    * Disable this if you want to react to your own dependent resource updates
    *
    * @since 4.5.0
    */
-  default boolean previousAnnotationForDependentResources() {
+  default boolean previousAnnotationForDependentResourcesEventFiltering() {
     return true;
   }
 
   /**
-   * If the event logic should parse the resourceVersion to determine the ordering of events. This
-   * is typically not needed.
+   * If the event logic should parse the resourceVersion to determine the ordering of dependent
+   * resource events. This is typically not needed.
    * <p>
    * Disabled by default as Kubernetes does not support, and discourages, this interpretation of
    * resourceVersions. Enable only if your api server event processing seems to lag the operator
@@ -387,7 +387,7 @@ public interface ConfigurationService {
    *
    * @since 4.5.0
    */
-  default boolean parseResourceVersions() {
+  default boolean parseResourceVersionsForEventFilteringAndCaching() {
     return false;
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -364,4 +364,31 @@ public interface ConfigurationService {
     return Set.of(ConfigMap.class, Secret.class);
   }
 
+  /**
+   * If an annotation should be used so that the operator sdk can detect events from its own updates
+   * of dependent resources and then filter them.
+   * <p>
+   * Disable this if you want to react to your own dependent resource updates
+   *
+   * @since 4.5.0
+   */
+  default boolean previousAnnotationForDependentResources() {
+    return true;
+  }
+
+  /**
+   * If the event logic should parse the resourceVersion to determine the ordering of events. This
+   * is typically not needed.
+   * <p>
+   * Disabled by default as Kubernetes does not support, and discourages, this interpretation of
+   * resourceVersions. Enable only if your api server event processing seems to lag the operator
+   * logic and you want to further minimize the the amount of work done / updates issued by the
+   * operator.
+   *
+   * @since 4.5.0
+   */
+  default boolean parseResourceVersions() {
+    return false;
+  }
+
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -97,10 +97,10 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
     super(client.resources(configuration.getResourceClass()), configuration, parseResourceVersions);
     if (parseResourceVersions) {
       knownResourceVersions = Collections.newSetFromMap(new LinkedHashMap<>() {
-          @Override
-          protected boolean removeEldestEntry(java.util.Map.Entry<String, Boolean> eldest) {
-              return size() >= MAX_RESOURCE_VERSIONS;
-          }
+        @Override
+        protected boolean removeEldestEntry(java.util.Map.Entry<String, Boolean> eldest) {
+          return size() >= MAX_RESOURCE_VERSIONS;
+        }
       });
     } else {
       knownResourceVersions = null;
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -96,11 +96,11 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
       boolean parseResourceVersions) {
     super(client.resources(configuration.getResourceClass()), configuration, parseResourceVersions);
     if (parseResourceVersions) {
-      knownResourceVersions = Collections.newSetFromMap(new LinkedHashMap<String, Boolean>() {
-        @Override
-        protected boolean removeEldestEntry(java.util.Map.Entry<String, Boolean> eldest) {
-          return size() >= MAX_RESOURCE_VERSIONS;
-        }
+      knownResourceVersions = Collections.newSetFromMap(new LinkedHashMap<>() {
+          @Override
+          protected boolean removeEldestEntry(java.util.Map.Entry<String, Boolean> eldest) {
+              return size() >= MAX_RESOURCE_VERSIONS;
+          }
       });
     } else {
       knownResourceVersions = null;
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -69,8 +69,6 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
     extends ManagedInformerEventSource<R, P, InformerConfiguration<R>>
     implements ResourceEventHandler<R> {
 
-  private static final int MAX_RESOURCE_VERSIONS = 256;
-
   public static String PREVIOUS_ANNOTATION_KEY = "javaoperatorsdk.io/previous";
 
   private static final Logger log = LoggerFactory.getLogger(InformerEventSource.class);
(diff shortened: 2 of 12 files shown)


csviri:
@shawkins could you pls rebase this PR?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -275,6 +295,9 @@ public boolean allowsNamespaceChanges() {
     return configuration().followControllerNamespaceChanges();
   }
 
+  public boolean parseResourceVersions() {
```
csviri, line 298: This is not used.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -41,8 +42,15 @@ public TemporaryResourceCache(ManagedInformerEventSource<T, ?, ?> managedInforme
...
 
-  public synchronized Optional<T> removeResourceFromCache(T resource) {
-    return Optional.ofNullable(cache.remove(ResourceID.fromResource(resource)));
+  public synchronized void onEvent(T resource, boolean unknownState) {
+    ResourceID resourceId = ResourceID.fromResource(resource);
+
+    T cachedResource = cache.get(resourceId);
+
+    if (unknownState || (cachedResource != null
+        && !isLaterResourceVersion(resourceId, cachedResource, resource))) {
```
csviri, line 51: Kinda makes this code hard to read that the feature flag is checked inside the `isLaterResourceVersion`, would extract on a level above or rename this function to denote that fact.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -61,18 +69,34 @@ public synchronized void putResource(T newResource, String previousResourceVersi
...
+      if (managedInformerEventSource.getInformerConfiguration() instanceof InformerConfiguration
+          && ((InformerConfiguration) managedInformerEventSource.getInformerConfiguration())
+              .parseResourceVersions()
+          && Long.compare(Long.parseLong(newResource.getMetadata().getResourceVersion()),
+              Long.parseLong(cachedResource.getMetadata().getResourceVersion())) > 1) {
+        return true;
       }
+    } catch (NumberFormatException e) {
+      log.debug(
+          "Could not compare resourceVersions {} and {} for {}",
```
csviri, line 95: Shouldn't be this a warning? even if it could pollute cache? (no strong opinion)

shawkins: It won't pollute the cache, it would exhibit the same behavior as it does now.  It can of course be at whatever level you want.

csviri: I mean pollute logs :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: Will add a javadoc

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
shawkins: Right, it had been in some early code, but is no longer needed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -61,18 +69,34 @@ public synchronized void putResource(T newResource, String previousResourceVersi
...
+    }
+  }
+
+  private boolean isLaterResourceVersion(ResourceID resourceId, T newResource, T cachedResource) {
+    try {
+      if (managedInformerEventSource.getInformerConfiguration() instanceof InformerConfiguration
+          && ((InformerConfiguration) managedInformerEventSource.getInformerConfiguration())
+              .parseResourceVersions()
+          && Long.compare(Long.parseLong(newResource.getMetadata().getResourceVersion()),
+              Long.parseLong(cachedResource.getMetadata().getResourceVersion())) > 1) {
```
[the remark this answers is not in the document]
shawkins: Thanks, corrected.

shawkins:
> Just some minor issues with current code. I'm aware that it is WIP

No problem.  Made some changes to address those concerns.  What about the methods for configuration hooks - are those in the right places and how far up would that need to be exposed?

csviri:
> > Just some minor issues with current code. I'm aware that it is WIP
> 
> No problem. Made some changes to address those concerns. What about the methods for configuration hooks - are those in the right places and how far up would that need to be exposed?

I think both should be in configuration service, some of the properties we want to configure also on DR level, like if it should use SSA (I will do a separate PR), but not there two IMO

shawkins:
@csviri rebased and further refined the configuration.  If this looks good I'll add some tests.

csviri:
> @csviri rebased and further refined the configuration. If this looks good I'll add some tests.

sure looks good imo

metacosm (changes requested):
More generally speaking, shouldn't the event skipping be handled at the ManagedInformerEventSource level instead of in the InformerEventSource? In particular, wouldn't it make sense to handle the resource versions in the temporary cache as well?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
...
 
+  /**
+   * If an annotation should be used so that the operator sdk can detect events from its own updates
+   * of dependent resources and then filter them.
+   * <p>
+   * Disable this if you want to react to your own dependent resource updates
+   *
+   * @since 4.5.0
+   */
+  default boolean previousAnnotationForDependentResources() {
```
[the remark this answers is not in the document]
shawkins: Since there have been multiple mechanism for this the thinking is to use specific rather than general terminology, so that if it changes again the user can be explicit about which mechanism(s) are in play.

If you strongly prefer the general feature name, that's fine too.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
...
+    return true;
+  }
+
+  /**
+   * If the event logic should parse the resourceVersion to determine the ordering of events. This
+   * is typically not needed.
+   * <p>
+   * Disabled by default as Kubernetes does not support, and discourages, this interpretation of
+   * resourceVersions. Enable only if your api server event processing seems to lag the operator
```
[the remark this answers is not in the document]
shawkins: It's basically for feature parity with the mutable cache that is in go client - I'm not aware of user request for the feature, so if you feel strongly about it, then it could be hidden or removed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
     return Set.of(ConfigMap.class, Secret.class);
   }
 
+  /**
+   * If an annotation should be used so that the operator sdk can detect events from its own updates
```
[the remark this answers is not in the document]
shawkins: Other than specifying what the annotation name is, there isn't much greater detail that is needed here - the user is not expected to do anything with these annotations.

shawkins:
> More generally speaking, shouldn't the event skipping be handled at the ManagedInformerEventSource level instead of in the InformerEventSource? 

Event skipping (filtering or event recording / previous annotation) was already being handled at the InformerEventSource level, this is really just another case to consider.

> In particular, wouldn't it make sense to handle the resource versions in the temporary cache as well?

I had considered that, but it didn't initially make sense given the other code structure.  Now that does make more sense because parseResourceVersions is passed into the TemporaryCache constructor.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
...
+   * is typically not needed.
+   * <p>
+   * Disabled by default as Kubernetes does not support, and discourages, this interpretation of
+   * resourceVersions. Enable only if your api server event processing seems to lag the operator
+   * logic and you want to further minimize the the amount of work done / updates issued by the
+   * operator.
+   *
+   * @since 4.5.0
+   */
+  default boolean parseResourceVersions() {
```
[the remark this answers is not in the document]
shawkins: Let's decide if you even want this before making a name change.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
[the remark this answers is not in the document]
shawkins: @metacosm @csviri what do you think - would it be better to pull out or hide the resourceVersion parsing logic?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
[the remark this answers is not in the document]
csviri: I think this would be too generic. 
`previousAnnotationForDependentResourcesEventFiltering` or somethink would make more sense to me.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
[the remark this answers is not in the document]
csviri: In general I'm not against this. Also would stick with the original name, thus `parseResourceVersions` since this can have later effect on other parts.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -78,14 +80,31 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
   private final PrimaryToSecondaryMapper<P> primaryToSecondaryMapper;
   private Map<String, Function<R, List<String>>> indexerBuffer = new HashMap<>();
   private final String id = UUID.randomUUID().toString();
+  private final Set<String> knownResourceVersions;
```
csviri, line 83: Please document what are the situations this is solving

csviri:
I like the way it is, just we need to document every change / feature in the PR since it might be hard to follow after time the corner case situatios. But otherwise looks good.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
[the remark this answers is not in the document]
metacosm: Yes, I think it would make sense to document the annotation name so that people can look it up in case people notice it in their resources and wonder what that is.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
     return Set.of(ConfigMap.class, Secret.class);
   }
 
+  /**
+   * If a javaoperatorsdk.io/previous annotation should be used so that the operator sdk can detect events from its own updates
```
metacosm, line 368:
```suggestion
   * If a {@code javaoperatorsdk.io/previous} annotation should be used so that the operator sdk can detect events from its own updates
```

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -364,4 +364,31 @@ default Set<Class<? extends HasMetadata>> defaultNonSSAResource() {
...
+   * is typically not needed.
+   * <p>
+   * Disabled by default as Kubernetes does not support, and discourages, this interpretation of
+   * resourceVersions. Enable only if your api server event processing seems to lag the operator
+   * logic and you want to further minimize the the amount of work done / updates issued by the
+   * operator.
+   *
+   * @since 4.5.0
+   */
+  default boolean parseResourceVersionsForDependentResourcesEventFilteringAndCaching() {
```
csviri, line 390: This kind of a parsing and caching could be done also for the primary resource. For that reason would like to omit that this is specific for the dependent resources, and just maybe add it to the description now.

metacosm:
The tests are also failing consistently so something is fishy there.

shawkins:
The remaining test failure is already called out as #2043

csviri (approved):

PR #2036, opened 2023-08-28
improve: logging if no primary found for resource id

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -146,7 +146,7 @@ private void submitReconciliationExecution(ResourceState state) {
             controllerUnderExecution,
             maybeLatest.isPresent());
         if (maybeLatest.isEmpty()) {
-          log.debug("no custom resource found in cache for resource id: {}", resourceID);
+          log.debug("no primary resource found in cache with resource id: {}", resourceID);
```
metacosm, line 149: Better but still doesn't address the fact that we might get into that state because we're triggering event processing even though we don't expect a resource to even exist in the cache (like if a user sends events for resources that are not tracked by the Reconciler), i.e. the cache is an implementation detail that might be confusing, i.e. a user might then think: ok, it's not in cache, then fetch it from the server instead! 
That said, it's indeed a debug level log so maybe it doesn't matter all that much.

csviri: > we don't expect a resource to even exist in the cache

Yeah, but this is not always the case, if the controller informer is lagging (reconnect issues), the resource might be actually expected, just not yet there.

> it's not in cache, then fetch it from the server instead!

this is for primary, not sure if there is a place to fetch it; since the reconciler never triggered.

metacosm (approved):

PR #2049, opened 2023-09-08
fix: processor issue with cleanup resources

csviri:
if leader election not running delete event was not cleaning up state


metacosm (approved):
LGTM

PR #2050, opened 2023-09-10
Re-use ignoreList in GenericKubernetesResourceMatcher

Donnerbart:
Just a small improvement to re-use the `ignoreList` from line 178 instead of creating the identical `ignoreList` again in line 208.

Also optimized two `if (!method()) return false;` to `return method();` (suggestion from IDEA).


csviri (approved):
LGTM. Thank you @Donnerbart

PR #2053, opened 2023-09-12
feat: make primary cache accessible from Context

metacosm:
Alternative to #2051

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
@@ -10,6 +10,7 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.managed.ManagedDependentResourceContext;
 import io.javaoperatorsdk.operator.processing.event.EventSourceRetriever;
+import io.javaoperatorsdk.operator.processing.event.source.IndexedResourceCache;
 import io.javaoperatorsdk.operator.processing.event.source.IndexerResourceCache;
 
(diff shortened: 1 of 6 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -43,4 +44,12 @@ <R> Optional<R> getSecondaryResource(Class<R> expectedType,
...
    */
   ExecutorService getWorkflowExecutorService();
+
+  /**
+   * Retrieves the primary resource cache.
+   *
+   * @return the {@link IndexerResourceCache} associated with the associated {@link Reconciler} for
+   *         this context
+   */
+  IndexerResourceCache<P> getPrimaryCache();
```
metacosm, line 54: I don't like the fact that the `IndexerResourceCache` interface allows access to creating new indices at any time when these would actually only take effect whenever the associated informers are initialized so I think we should split this interface into a read-only part and one to manage indexers. The alternative would be to only expose `ResourceCache` instead of `IndexerResourceCache`.

csviri:
This should rather target `next` imo, since it's a feature.

csviri:
But I think we should we can go this way, just exposing the cache, and see if anybody would need actually more.

csviri (approved):
LGTM, apart that should target next.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
metacosm: I've added an `IndexedResourceCache` interface that extends `ResourceCache` to add the indexed-based retrieval and moved both classes to the reconciler package since they'd now be part of the API… It shouldn't break anything since the only thing that was used externally was the `IndexerResourceCache` interface.

csviri (approved):

PR #2059, opened 2023-09-18
fix: leader election stop not called

csviri:


ashangit (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -67,7 +67,7 @@ private void init(LeaderElectionConfiguration config) {
                 config.getRenewDeadline(),
                 config.getRetryPeriod(),
                 leaderCallbacks(),
-                true,
+                false,
```
metacosm, line 70: Can you add a comment to explain why setting this to true prevented stop from being called, please?

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
csviri: added

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java
```diff
@@ -67,7 +67,7 @@ private void init(LeaderElectionConfiguration config) {
                 config.getRenewDeadline(),
                 config.getRetryPeriod(),
                 leaderCallbacks(),
-                true,
+                false, // this is required to receive stop event in all cases
```
metacosm, line 70: Why, though?

csviri: what do you mean? we want to get those all cases

metacosm: What I mean is that the comment doesn't explain why changing the value to `false` makes sure that `stop` is called…

csviri: That should be part of the fabric8 client docs rather, actually not there but it's in the go docs?
https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig
Should I link this there?

metacosm: Maybe but mostly explain why this "release on cancel" setting has an impact on the lifecycle of our manager.

csviri: extended the comment, but basically this should be very straghforward, that the onStop is called, also if the lock is released.

metacosm (approved):

csviri:
> https://github.com/operator-framework/java-operator-sdk/pull/2059/files#diff-e493dfcdf8473a2824e76c20c95f07b6dc94a1305f6816d029b7de8e73a46931R59 should be removed. sorry for seeing this late

ahh, yes: https://github.com/operator-framework/java-operator-sdk/pull/2066

PR #2060, opened 2023-09-18
fix: prevent logs pollution for recursive matching algorithm

csviri:


metacosm (commented):
I'm not sure what this change is supposed to fix. How is changing the logging level addressing any recursion?

csviri:
> I'm not sure what this change is supposed to fix. How is changing the logging level addressing any recursion?

So it is not polluting the logs on Debug level.

metacosm:
OK, so I think this needs a better wording on the commit message then :)

csviri:
> OK, so I think this needs a better wording on the commit message then :)

Tried my best, changed the title. But feel free to change it :)

metacosm (approved):

PR #2068, opened 2023-09-21
fix: snapshot version update

csviri:
There was a problem with the latest release, version update action did not run properly.


metacosm (approved):
Why wasn't this done via the release process, actually?

csviri:
There was a partial error as it seems on maven side, so restart did not help fully. But problem was only with the caffeine cache module, so not big deal. But the update version action was not able to run.

PR #2082, opened 2023-10-09
chore: upgrade to Fabric8 client 6.9.0

metacosm:


metacosm:
Will open a new PR with 6.9.1 instead, as it seems to address the type inference issue (pending a potential 6.9.2 release which might address other issues as well).

PR #2136, opened 2023-12-01
feat: generic kubernetes resource usage

csviri:
This features makes it possible to use `GenericKubernetesResource` in plain reconciliations also in dependent resources.


metacosm:
Wondering if we should also make generic dependents available via the `Context`. Otherwise, LGTM.

metacosm (approved):

csviri:
> Wondering if we should also make generic dependents available via the `Context`. Otherwise, LGTM.

It is available with `context.getSecondaryResources(GenericKubernetesResource.class)` It might be not ideal, this way, I mean a GVK as param would be nice, (or just that), but using this in a project an it is easy enough to handle this way.

PR #2174, opened 2023-12-20
improve: thread-safe idempotent dynamic informer event source start

csviri:
this will allow to multiple threads wait idempotently for the start of an informer event source

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java
@@ -229,22 +229,21 @@ public class EventSourceManager<P extends HasMetadata>
   }
 
   @Override
-  public synchronized EventSource dynamicallyRegisterEventSource(String name,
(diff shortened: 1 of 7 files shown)


metacosm:
Can you elaborate on what this changes, please?

csviri:
Converted it to draft (meant it that way originally), until not finished.

The reason for this is related to dynamic registration, if in a reconciliation multiple event sources are registered dynamically possible of same name from different thread (and also different type), it is hard to make that efficient ATM. Thus to not make all the threads possibly wait in the synchronized block of [here](https://github.com/java-operator-sdk/java-operator-sdk/blob/9d8f4547c8844235a8e50cdd544ce5a02537a750/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventSourceManager.java#L232-L232). One way to do this si to query if the ES is already registered. But if registered does not mean it is started. By this change, calling the start method should ensure that it is started (in a thread-safe way) and also simultaneously waiting for the other thread if that is starting the event source. We can look on this as a lock striping.

metacosm (approved):

PR #2185, opened 2024-01-08
feat: extending event source context with workflow support

csviri:
also change webpage sample use workflow builder


metacosm (approved):
LGTM

PR #2190, opened 2024-01-09
feat: throw exception on problematic owner ference adding in dependent resources

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
```diff
@@ -71,6 +74,30 @@ public static String getNameFor(Class<? extends Reconciler> reconcilerClass) {
     return getDefaultNameFor(reconcilerClass);
   }
 
+  public static void checkIfCanAddOwnerReference(HasMetadata owner, HasMetadata resource) {
+    if (owner instanceof GenericKubernetesResource
```
metacosm, line 78: Maybe we should move this code to the Fabric8 client?

csviri: Yes, so I would merge this, and will work on that to move it to fabri8 client.
If also @manusa @shawkins thinks it a good idea. Not sure if we should add it as part of an utility, or just simply to `.addOwnerReference` or both.

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtils.java
metacosm: I would add it to `addOwnerReference`, personally.

csviri: yes, me too

manusa: Should be part of HasMetadata#addOwnerReference.

The other suggested alternative would involve exposing the validation method and then calling that method within addOwnerReference too. I don't really see the value of exposing that method.

PR #2226, opened 2024-02-01
fix: multiple dependents of same type exceptions

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowResult.java
```diff
@@ -36,9 +37,22 @@ public boolean erroredDependentsExist() {
 
   public void throwAggregateExceptionIfErrorsPresent() {
     if (erroredDependentsExist()) {
+      Map<String, Exception> exceptionMap = new HashMap<>();
+      Map<String, Integer> numberOfClasses = new HashMap<>();
+
+      for (Entry<DependentResource, Exception> entry : erroredDependents.entrySet()) {
```
metacosm, line 43: I'm not sure that numbering the dependents help here since people might give meaning to the number and/or wonder where the numbers come from. Also, since we're processing an unordered collection, the numbers might "identify" different dependents over time making things confusing, imo. I think it would be better to find a stable way to discriminate between DependentResources of the same type and use that rather than an arbitrary number.

csviri: Yes, sure, so this is basically just a temp, solution created issue for a proper one:
https://github.com/operator-framework/java-operator-sdk/issues/2227

Note that this is quite rare that something like this would happen in normal circumstances, having the same type of DR in a workflow and at least 2 having exception in same time. 

(Maybe a randomized string would be also good there?)

metacosm (approved):

PR #2250, opened 2024-02-26
feat: move name is directly to dependent resource

csviri:
- use this name when throwing aggregate exception

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
@@ -69,6 +69,8 @@ public interface DependentResource<R, P extends HasMetadata> {
     return this instanceof Deleter;
   }
 
-  String name();
+  default String name() {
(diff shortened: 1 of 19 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/DependentResource.java
```diff
@@ -68,4 +68,8 @@ static String defaultNameFor(Class<? extends DependentResource> dependentResourc
   default boolean isDeletable() {
     return this instanceof Deleter;
   }
+
+  String getName();
```
csviri, line 72: Ad other approach would be just have 
`String name()`

And the setter in a separate Inteface, but we want to set it from spec.
What do you think @metacosm

csviri: Changed it to a setter interface. Both are a bit ugly. But this highest level interface (`DependentResource`) is more consistent and readable.

metacosm (approved):

PR #2252, opened 2024-02-27
feat: distinguish resources based on desired state

csviri:


csviri:
TODO: 
- [ ] docs,
- [ ] release notes
- [ ] Sample for external resources

docs/documentation/dependent-resources.md
```diff
@@ -403,6 +409,11 @@ public class MultipleManagedDependentResourceConfigMap1 {
 }
 ```
 
+While using the automated mechanism should work well in most cases, resource discriminators might make sense for
```
metacosm, line 412: Note that I changed the wording here: I think we should keep the discriminators also for optimization purposes and not deprecate them or remove them in future releases because they might still be useful in some cases.

metacosm (approved):

metacosm (approved):

PR #2255, opened 2024-02-28
refactor: add default implementation for name()

metacosm:
keep name on DR node,Signed-off-by: Chris Laprun <claprun@redhat.com>

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -30,16 +30,21 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
   private ResourceDiscriminator<R, P> resourceDiscriminator;
   private final DependentResourceReconciler<R, P> dependentResourceReconciler;
 
-  protected String name = DependentResource.defaultNameFor(this.getClass());
+  protected String name;
 
   @SuppressWarnings({"unchecked"})
   protected AbstractDependentResource() {
+    this(null);
+  }
+
+  protected AbstractDependentResource(String name) {
     creator = creatable ? (Creator<R, P>) this : null;
     updater = updatable ? (Updater<R, P>) this : null;
 
     dependentResourceReconciler = this instanceof BulkDependentResource
         ? new BulkDependentResourceReconciler<>((BulkDependentResource<R, P>) this)
         : new SingleDependentResourceReconciler<>(this);
+    this.name = name == null ? DependentResource.defaultNameFor(this.getClass()) : name;
   }
 
   /**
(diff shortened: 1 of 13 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -30,16 +30,21 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
   private ResourceDiscriminator<R, P> resourceDiscriminator;
   private final DependentResourceReconciler<R, P> dependentResourceReconciler;
 
-  protected String name = DependentResource.defaultNameFor(this.getClass());
+  protected String name;
 
   @SuppressWarnings({"unchecked"})
   protected AbstractDependentResource() {
+    this(null);
```
csviri, line 37: `DependentResource.defaultNameFor(this.getClass())`  could be passed here

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/DependentResourceNode.java
```diff
@@ -54,6 +54,9 @@ public Optional<Condition<R, P>> getReconcilePrecondition() {
     return Optional.ofNullable(reconcilePrecondition);
   }
 
+  public String getName() {
```
csviri, line 57: This is not very good practice in my opinion, it kinda implies that it's the name of the dependent node not the dependent resource. I would either remove this, and always access the name though DependentResource or rename this to `getDependentResourceName()`

csviri (changes requested):
Added to minor things to discuss otherwise looks good to me

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
metacosm: No, this actually doesn't work (I tried, iirc) because you cannot call `this.getClass()` until the chained constructor is called. This will be possible in Java 22, though, I think.

csviri: Ahh ok , wishful thiking... :) thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/DependentResourceNode.java
metacosm: `DependentResourceNode` is an internal class so I'd argue that it doesn't really matter and since Dependent Resources always have names now, the name of the node is the name of the associated dependent, which this method captures. This certainly could be reverted but, to me, it's pretty clear that the name of the DRN is the name of the associated DR. :)

csviri: pls do, when started looking it was kinda confusing now, what now we are talking about

csviri: This is really small thing, at least for me it is  much easier to jump to a conclusion with that pattern

csviri (approved):
LGTM

PR #2257, opened 2024-03-01
improve: remove EventSourceInitializer

csviri:
Move `EventSourceIntiailizer` directly to `Reconciler`. Based on the patter, if properly done, there will be always EventSource. Still can return empty map.


csviri:
We should consider (I'm leaning to this) not even provide a default implementation, although might be confusing for someone who starts first, that is that part not simple to understand first (like how Informers works and why essential), but at least moves users news to operators to good direction.

what do you tink @metacosm @jcechace  ?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Reconciler.java
```diff
@@ -14,6 +20,78 @@ public interface Reconciler<R extends HasMetadata> {
...
+  }
+
+  /**
+   * Utility method to easily create map with generated name for event sources. This is for the use
+   * case when the event sources are not access explicitly by name in the reconciler.
+   *
+   * @param eventSources to name
+   * @return even source with default names
+   */
+  static Map<String, EventSource> nameEventSources(EventSource... eventSources) {
```
metacosm, line 45: I would prefer that we move all these methods to a different utilty class. They have nothing to do with Reconciler itself and pollute its API, imo.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
```diff
@@ -8,9 +8,7 @@
 /**
  * Creates an event source to trigger your reconciler whenever something happens to a secondary or
  * external resource that would not normally trigger your reconciler (as the primary resources are
- * not changed). To register EventSources with so that your reconciler is triggered, please make
- * your reconciler implement
- * {@link io.javaoperatorsdk.operator.api.reconciler.EventSourceInitializer}.
+ * not changed). To register EventSources with so that your reconciler is triggered.
```
metacosm, line 11: This doesn't make sense anymore.

csviri:
> The default implementation should stay because, imo, most people will never need to implement this method. What would the implementation be for someone who only uses managed workflows, for example?

ok that is good argument, to keep the default impl.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Reconciler.java
csviri: Will move it for now, was thinking to also have have EventSources "Named" like now with the Dependent Resources. Will make a separate PR. That would make these methods basically obsolete.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/EventSource.java
csviri: changed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Reconciler.java
csviri: Added `EventSourceUtils`.

metacosm (changes requested):
The documentation needs updates.

docs/documentation/dependent-resources.md
```diff
@@ -307,16 +306,12 @@ public class WebPageStandaloneDependentsReconciler
 There are multiple things happening here:
 
 1. Dependent resources are explicitly created and can be access later by reference.
-2. Event sources are produced by the dependent resources, but needs to be explicitly registered in
-   this case by implementing
-   the [`EventSourceInitializer`](https://github.com/java-operator-sdk/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/EventSourceInitializer.java)
-   interface.
-3. The input html is validated, and error message is set in case it is invalid.
-4. Reconciliation of dependent resources is called explicitly, but here the workflow
+2. The input html is validated, and error message is set in case it is invalid.
```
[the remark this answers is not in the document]
csviri: I think now this is fixed by you :)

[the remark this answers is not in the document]
metacosm: No, the example and its documentation still needs to be updated.

metacosm (approved):

metacosm (approved):

PR #2262, opened 2024-03-05
fix: allow to burst threads by default

csviri:
The previous approach used a non-bound queue, therefore the thread were not added dynamically into the Threadpool.
As described in the docs of https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html
If the queue is not bounded, it will just accumulate the submitted tasks.


csviri:
This approach will lead to better startup performance. The ExecutorService can be replaced and fine-tuned. 

IMO the best approach would be to have a pooled executor that increases the size as now to an upper bound on demand and then queues the remaining elements, which would cover the initial burst on startup nicely.

csviri:
This could be a change for v5, we don't need to necessarily release it as a bugfix.

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/2265

PR #2265, opened 2024-03-06
improve: use fixed thread pool explicitly

csviri:
Note that it was a fixed thread pool also now, just it was confusing with min and max values,


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -191,13 +200,11 @@ default Metrics getMetrics() {
   }
 
   default ExecutorService getExecutorService() {
-    return newThreadPoolExecutor(minConcurrentReconciliationThreads(),
-        concurrentReconciliationThreads());
+    return Executors.newFixedThreadPool(DEFAULT_RECONCILIATION_THREADS_NUMBER);
```
ysymi, line 203: current approach is safe enough. If user want to get better startup performance, they can increase reconciliation thread number or override getExecutorService method

metacosm (approved):

PR #2274, opened 2024-03-08
feat: Workflow extracted to a separate annotation

csviri:


metacosm (approved):
Maybe we should make the ControllerConfiguration annotation optional as well.

csviri:
> Maybe we should make the ControllerConfiguration annotation optional as well.

yeah, would make sense, could you create an issue pls?

PR #2302, opened 2024-03-20
improve: replace current formatting plugins with spotless plugin 

csviri:


operator-framework-bom/pom.xml
```diff
@@ -1,142 +1,170 @@
...
+    <plugins>
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <version>${spotless.version}</version>
+        <configuration>
+          <pom>
+            <includes>
+              <include>pom.xml</include>
+              <inclide>./**/pom.xml</inclide>
```
metacosm, line 86:
```suggestion
              <include>./**/pom.xml</include>
```

sample-operators/leader-election/src/main/java/io/javaoperatorsdk/operator/sample/LeaderElectionTestReconciler.java
```diff
@@ -2,15 +2,14 @@
 
 import java.time.Duration;
 import java.util.ArrayList;
+import javaoperatorsdk.sample.v1.LeaderElection;
+import javaoperatorsdk.sample.v1.LeaderElectionStatus;
```
metacosm, line 6: These two classes should be put in the proper package but that can be done in a different PR.

metacosm (approved):

operator-framework-bom/pom.xml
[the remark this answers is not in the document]
csviri: wonder what parser is used in the background, that this worked :)

PR #2306, opened 2024-03-21
improve: naming reconcile pre-condition is just condition

csviri:


csviri:
- we might want to discuss still if we want this, if yes however, would be probably rename also:
 ReadyPostCondition -> ReadyCondition
 DeletePostCondition -> DeleteCondition

@metacosm @jcechace what do you think?

jcechace:
IDK .. personally I like the pre/post-codition naming as it makes it implicit at what stage the condition is evaluated.

csviri:
> IDK .. personally I like the pre/post-codition naming as it makes it implicit at what stage the condition is evaluated.

yes, that is why I'm hesitating. Just we had a few misunderstandings mainly with `reconcilePreCondition`, since it also deletes the resource if the condition does not hold.

csviri:
But more and more I think about it, it might not be worth it.

metacosm:
> > IDK .. personally I like the pre/post-codition naming as it makes it implicit at what stage the condition is evaluated.
> 
> yes, that is why I'm hesitating. Just we had a few misunderstandings mainly with `reconcilePreCondition`, since it also deletes the resource if the condition does not hold.

I don't think that renaming the condition to something else (unless that becomes really a mouthful like `reconcileOrDeletePreCondition`) would help. We already have javadoc for the meaning of these conditions. Maybe we could improve them but I'm not even sure we could be much clearer.

csviri:
kk, agree, will close this issue and the PR for now.

PR #2313, opened 2024-03-25
feat: handle clustered resource on secondary to primary mapper init

### LINKED ISSUE #2311: Handling Cluster Scoped Primary Resource in Dependent Resource `SecondaryToPrimaryMapper`
Currently by default Dependent Resource assumes that the primary resource is namespace scoped, see:

https://github.com/operator-framework/java-operator-sdk/blob/0b17878ca5497d97838faa0bb8754ca61b8bd414/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L90-L90

This is something users have to then handle explicitly, by overriding the `SecondaryToPrimaryMapper`, see:

https://github.com/operator-framework/java-operator-sdk/issues/2310#issuecomment-2017523701

In DRs we don't have access to the class of primary to check if it is namespace scoped. But we might create some workarounds, like passing it as a flag for managed resources automatically.

metacosm:
Fixes #2311


csviri:
@metacosm this is wrong, it is the type of the secondary resource not the primary what you check. It should be the primary

metacosm:
I've changed the approach, but I'm not sure if the solution I chose is agreeable or not… I would say that it would work without changes most of the time but there are cases (as shown in some of the tests) where a light user intervention would be needed so maybe we want to: 
1. keep the change as-is and decide that it's ok for main
2. keep the change as-is but schedule it for v5
3. decide for a completely different approach altogether

Personally, I think that having the primary type available from the dependent resource might be useful (but with the current change, this is only available to descendants of `KubernetesDependentResource`).

csviri:
IMO this should target next, we might want to do few iteration also reflecting on other issues.

> Personally, I think that having the primary type available from the dependent resource might be useful (but with the current change, this is only available to descendants of KubernetesDependentResource).

Basically, we need primary just for this case for now, so having it as a configuration option (boolean) for KubernetesDependentResource would be enough.

Also, I think ATM is enough to have for Kubernetes Dependents. Only those use owner references.

metacosm:
OK, retargering to `next`.

metacosm:
Closed in favor of #2321

Issue #2314, opened 2024-03-25
Suspected timing issue around temporary cache and garbage collection of resources


shawkins:
## Bug Report

Keycloak tests in the same class run in the same namespace and typically just reuse the same parent cr name. The cleanup between tests ensures that the dependent statefulset is gone, but doesn't check for other dependents such as secrets or services.

In a rare case (1% of runs?) there is a timing issue happening where the state of the temporary cache is out of sync with the informer cache in such a way that does not resolve on its own. In particular the temporary cache will show a resource as existing, while the informer cache / server has it as non-existent.

#### What did you do?

It's difficult to reproduce, turning up logging made it even less likely to occur - with either fabric8 informer or operator sdk debug logging turned on for the reflector / informer it failed to reproduce.

Instead I resorted to some logic that checked the state of getSecondarResource vs a listing of what was in cache when were'nt fully ready after the main resolving method. That confirmed there was a temporary cache entry, but nothing in the informer cache nor on the server.

#### What did you expect to see?

The temporary cache state in sync.

#### What did you see instead? Under which circumstances?

In an extremely rare case, and seemingly related to the garbage collection of resource with the same name, the state was out-of-sync.

Somehow the temporary cache is referencing the garbage collected entry - or one that it thought it created shortly after garbage collection, but that the api server mistakenly deletes.

A possible explanation: a create of the resource seems successful so a resource is being returned for that operation. The client then receives an added and delete event in rapid succession - prior to the operator setting the state of the temporary cache. Then once it sets the temporary cache, it is no longer valid nor will there be an event that causes it to correct.

#### Environment

v4.8.0, Minikube v1.32.0, Kubernetes v1.27.10

#### Possible Solution

I'm not sure yet. I'd like to capture the full sequence of events to better understand. It could possibly invole tombstoning the temporary cache so that it won't set an entry if it's "too soon" after a delete event.

shawkins:
Ok, I think the full sequence goes something like this:

Test run finishes and deletes the parent keycloak, but doesn't wait for the delete event nor all of the dependents to be deleted.
The next run of the test will start off by using the keycloak that has been marked for deletion as the operator hasn't received that event yet, even though the next test has created a new version of the keycloak resource.
A reconciliation of the old keycloak happens at a point in time such that the old secret does not exist, so a new one is created but it's owner reference is still set to the one to be deleted.
Then prior to the operator setting the temporary cache state, it receives the add and delete events for that secret.
[...]

PR #2317, opened 2024-03-25
fix: using tombstones to account for rapid deletion

shawkins:
closes: #2314

If during a creation the events for both the add and the deletion occur before the resource is being added to the temporary cache, the temporary cache moves backwards in a way that may not resolve itself.

The approach here is to use tombstones to track deletions. The benefit of this approach is that it keeps the logic encapsulated to the TemporaryResourceCache. The drawback is that we're only guessing about how long to retain the tombstones.

The other approach would look more similar to event recording, which is notify the cache that a create is happening prior to when it starts so that deletions / tombstones can gathered for that specific resource.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
@@ -1,10 +1,8 @@
 package io.javaoperatorsdk.operator.processing.event.source.informer;
 
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
@@ -18,8 +16,8 @@ import io.javaoperatorsdk.operator.processing.event.ResourceID;
 /**
  * <p>
  * Temporal cache is used to solve the problem for {@link KubernetesDependentResource} that is, when
- * a create or update is executed the subsequent getResource opeeration might not return the
- * up-to-date resource from informer cache, since it is not received yet by webhook.
+ * a create or update is executed the subsequent getResource operation might not return the
+ * up-to-date resource from informer cache, since it is not received yet.
  * </p>
  * <p>
  * The idea of the solution is, that since an update (for create is simpler) was done successfully,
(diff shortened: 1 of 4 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -60,7 +63,23 @@ protected boolean removeEldestEntry(java.util.Map.Entry<String, Boolean> eldest)
...
+  }
+
+  public synchronized void onAddOrUpdateEvent(T resource) {
+    onEvent(resource, false);
+  }
+
+  synchronized void onEvent(T resource, boolean unknownState) {
+    if (!tombstones.isEmpty()) {
+      var iter = tombstones.entrySet().iterator();
+      var entry = iter.next();
```
csviri, line 78: Not sure if understand this part, it always removes one random event from the tombstone if expired? Didn't you mean to iterate over all events?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -38,8 +38,11 @@ public class TemporaryResourceCache<T extends HasMetadata> {
 
   private static final Logger log = LoggerFactory.getLogger(TemporaryResourceCache.class);
   private static final int MAX_RESOURCE_VERSIONS = 256;
+  private static final int TOMBSTONE_RETENTION_TIME = 120000;
```
csviri, line 41: Wondering about this value, if high enough.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -84,14 +103,25 @@ public synchronized void putResource(T newResource, String previousResourceVersi
...
-    if ((previousResourceVersion == null && cachedResource == null)
+    boolean moveAhead = false;
+    if (previousResourceVersion == null && cachedResource == null) {
+      if (tombstones.get(newResource.getMetadata().getUid()) != null) {
+        log.debug(
+            "Won't resurrect uid {} for resource id: {}",
+            newResource.getMetadata().getUid(), resourceId);
+        return;
+      }
+      moveAhead = true;
```
csviri, line 114: THis seems to be correct, but would be good to document why is this here, isn't is covered already by the condition on line 120-121 ? or just an optimization?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: No, this is to amortorize the expiration logic - you could of course decide on a greater number than 1.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: It is certianly possible there could be an extreme scenario where the client is configured for a bunch of exponential retries - I'm not sure there is a hard cap on the amount of time in that case. We can keep things for longer, or you can go back the other direction of telling the cache prior to the create that you are intending to create the given resource.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: I'll add a comment, it's just breaking apart the checks that were being performed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: Maybe desribe in there as a comment too, that is intentional, since I will at least naivly stop on this every time I open it :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
metacosm: I would like to see this as a comment, indeed, as well.

metacosm (changes requested):

metacosm (changes requested):
LGTM apart from missing comment explaining the amortization of cache emptying logic

shawkins:
@metacosm @metacosm updated the code with further comments and consolidated the secondary caching concerns to a single class. From what I can tell it will take around 170 MB to track up to 1000000 deletions. Also updated the base retention time to 10 minutes, rather than 2. Based upon seeing what the retained heap size of this could be, I also updated the knownResourceVersions cache to be able to store more entries.

metacosm (approved):

PR #2319, opened 2024-03-26
fix: using tombstones to account for rapid deletion

shawkins:
closes: #2314

uses pre/post calls, rather than time, to track tombstones. If you want to go this direction instead that seems fine, it will just push more up into the calling layers.


csviri:
I see thx, tbh I would prefer the former one: https://github.com/operator-framework/java-operator-sdk/pull/2317

PR #2320, opened 2024-03-26
improve: additional logging to mysql schema e2e test

csviri:


metacosm (approved):

csviri:
The issues was that event when the schema in test is instructed to be created to be utf-8, it is not, but having the name to be the ID here is good enough IMO, we can take a look later why it like this in mysql

metacosm (approved):

PR #2321, opened 2024-03-26
feat: handle clustered resource on secondary to primary mapper init

### LINKED ISSUE #2311: Handling Cluster Scoped Primary Resource in Dependent Resource `SecondaryToPrimaryMapper`

metacosm:
Fixes #2311

Overriding getPrimaryResourceType should allow to make things work even
in deeper hierarchies.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
@@ -103,21 +103,22 @@ public class Utils {
       return defaultValue;
     } else {
       property = property.trim().toLowerCase();
-      switch (property) {
-        case "true":
-          return true;
-        case "false":
-          return false;
-        default:
-          return defaultValue;
-      }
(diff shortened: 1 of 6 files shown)


csviri (approved):
LGTM, 

as I mentioned, this is an interesting problem, we can decide later if this solution is good enough or not, but definitely a good improvement.

PR #2328, opened 2024-04-03
fix: processor concurrency issue on startup

csviri:
This should fix the related issue when on startup theoretically an event can be received when the processor is starting. Considering:
- On startup event received while EventProcess is not started yet
- EventProcessor starts and submits all the events (distinguished by resource id) to reconcile
- But mean while a new event is received for a certain resource that was also getting submitted by the start operation


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -400,7 +400,8 @@ public synchronized void stop() {
   }
 
   @Override
-  public void start() throws OperatorException {
+  public synchronized void start() throws OperatorException {
+    log.debug("Stating event processor: {}", this);
```
metacosm, line 404:
```suggestion
    log.debug("Starting event processor: {}", this);
```

metacosm (approved):

PR #2344, opened 2024-04-12
fix: change namespace starts processor on namespace change even if not leader

csviri:


csviri:
Would be a bit nicer to check if the instance is the leader, but this should solve the problem for now. Will add a test for this use case.

metacosm (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -368,17 +368,25 @@ private void validateCRDWithLocalModelIfRequired(Class<P> resClass, String contr
...
+    // mainly start the processor on namespace change.
+    boolean eventProcessorWasRunning = eventProcessor.isRunning();
+    if (eventProcessorWasRunning) {
+      eventProcessor.stop();
+    }
     eventSourceManager.changeNamespaces(namespaces);
-    eventProcessor.start();
+    if (eventProcessorWasRunning) {
+      eventProcessor.start();
+    }
```
gyfora, line 389: could it happen that at this point we are are already the leader (and accidentally not start processing) or the `synchronised` will take care of that

csviri: exactly, `synchronized` takes care of that

gyfora (approved):

PR #2363, opened 2024-04-24
feat: silent exception handling in managed workflows

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Workflow.java
```diff
@@ -19,4 +19,13 @@
...
   boolean explicitInvocation() default false;
 
+  /**
+   * if true and an exception(s) is thrown in the managed workflow, when the workflow reconciliation
+   * finished the controller won't throw an exception, instead will call the reconcile method where
+   * the reconcile result is accessible through
+   * {@link Context#managedWorkflowAndDependentResourceContext()}. If false, the exception is thrown
+   * by the controller before executing the reconcile method.
+   */
+  boolean silentExceptionHandling() default false;
```
metacosm, line 29: This doc doesn't really help, in my mind. It would be better to explain why someone would use this rather than what it does. The method name might not be explicit enough either.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Workflow.java
csviri: yeah, was thinking on a better name, but did not come up with anything better.
Any suggestion? :)

maybe `explicitExceptionHandling` ?

csviri: or `explicitExceptionChecking` ?

metacosm (approved):

PR #2364, opened 2024-04-24
feat: feature flag to not clone secondary resource on access

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -404,4 +404,15 @@ default boolean useSSAToPatchPrimaryResource() {
...
 
+  /**
+   * By default, secondary resources are cloned when returned from the cache.
+   * ({@link io.javaoperatorsdk.operator.api.reconciler.Context#getSecondaryResource(Class)}) This
+   * might have a performance impact, and if reconciler relies on SSA it is not needed to do. Since
+   * the goal of the cloning is to prevent modification on the object in the cache, that would no
+   * longer reflect the state on the server in that case.
+   */
+  default boolean cloneSecondaryResourcesWhenGettingFromCache() {
+    return true;
```
metacosm, line 415: Shouldn't that be `false` by default since we're using SSA by default now, no?

csviri: Was, thinking about that, and there are pros/cons to that. It quite risky when somebody is migrating to the new version. On the other hand yes, that would be more optimal to have by default.

metacosm: The new version will have significant changes… actually, I'm thinking that the change to SSA will require significant rewrite of operators if you need to only send what changes. Right now, most operators send the whole changed resource, not simply what has changed. 
For that matter, that's actually kind of what we're pushing towards with the `desired` state: people compute the whole resource and that's what gets used. It might not be a big deal because these resources' fields are probably managed by the JOSDK reconciler already but I'm wondering how prevalent that actually is.

metacosm: One thing to consider, though, is that people might not upgrade if they perceive the upgrade requires too many changes without bringing much benefit (i.e. if someone's operator already works correctly, they probably won't upgrade, especially if they need to change things a lot to do so).

csviri: yes, however, that is why we add feature flags.

metacosm (approved):

PR #2372, opened 2024-04-30
feat: read-only bulk dependent

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/BulkDependentResourceReconciler.java
```diff
@@ -24,18 +24,24 @@ class BulkDependentResourceReconciler<R, P extends HasMetadata>
...
+        && !(bulkDependentResource instanceof Updater<?, ?>)) {
+      return ReconcileResult
+          .aggregatedResult(actualResources.values().stream().map(ReconcileResult::noOperation)
+              .toList());
+    }
+
+    final var desiredResources = bulkDependentResource.desiredResources(primary, context);
 
     // remove existing resources that are not needed anymore according to the primary state
     deleteExtraResources(desiredResources.keySet(), actualResources, primary, context);
```
csviri, line 40: TBH this is interesting we should probably check if we are deleter? But what if not ? :)

metacosm: Not sure I understand…

metacosm (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/BulkDependentResourceReconciler.java
csviri: I updated it, before deleting extra resources was called all the time now we check if this DR is a `deleter`.

PR #2428, opened 2024-06-06
chore: adjust debug logs for updating a dependent resource

matteriben:
Moved existing debug log message and added new debug log message.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -71,10 +71,17 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
             logForOperation("Updating", primary, desired);
             var updatedResource = handleUpdate(actualResource, desired, primary, context);
             return ReconcileResult.resourceUpdated(updatedResource);
+          } else {
+            log.debug("Update skipped for dependent {} as it matched the existing one",
+                actualResource instanceof HasMetadata
+                    ? ResourceID.fromResource((HasMetadata) actualResource)
+                    : getClass().getSimpleName());
           }
         } else {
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -71,9 +71,12 @@ protected ReconcileResult<R> reconcile(P primary, R actualResource, Context<P> c
             logForOperation("Updating", primary, desired);
             var updatedResource = handleUpdate(actualResource, desired, primary, context);
             return ReconcileResult.resourceUpdated(updatedResource);
+          } else {
+            log.debug("Update skipped for dependent {} as it matched the existing one",
+                actualResource);
```
csviri, line 76: Let's not log the whole resources, this is a security concern, to not log secrets, thus sensitive data.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
```diff
@@ -71,10 +71,13 @@ protected ReconcileResult<R> reconcile(P primary, R actualResource, Context<P> c
...
             return ReconcileResult.resourceUpdated(updatedResource);
+          } else {
+            log.debug("Update skipped for dependent {} as it matched the existing one",
+                getClass().getSimpleName());
           }
         } else {
-          log.debug("Update skipped for dependent {} as it matched the existing one",
-              actualResource);
+          log.debug("Update skipped for dependent {} implement Updater interface to modify it",
+              getClass().getSimpleName());
```
csviri, line 80: Probably would be better to put here also the resource ID.

csviri (approved):

metacosm (approved):
We should use the dependent name in `next` as well.

PR #2436, opened 2024-06-14
fix: ignore docsy changes on build

csviri:


csviri:
added a PR but just for `next`, so we can avoid some conflicts,will close this one.

see: https://github.com/operator-framework/java-operator-sdk/pull/2443/files

PR #2451, opened 2024-06-20
refactor: move getResourceClassResolver to BaseConfigurationService

metacosm:
The class resolving mechanism is only needed in that class right now


csviri (approved):
LGTM, simpler is better :)

PR #2456, opened 2024-06-26
Update lease access check to account for lease name and multiple rules

MikeEdgar:
Allow the verbs required for lease access to be split between multiple rules and exclude rules that name a non-matching lease name. This allows rules such as those below to work as expected.

```yaml
rules:
  - apiGroups:
      - coordination.k8s.io
    resources:
      # The "create" verb cannot be used with "resourceNames"
      - leases
    verbs:
      - create
  - apiGroups:
      - coordination.k8s.io
    resources:
      - leases
    resourceNames:
      - my-lease
    verbs:
      - get
      - list
      - watch
      - delete
      - patch
      - update
```


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/LeaderElectionManagerTest.java
```diff
@@ -3,28 +3,37 @@
...
+import static io.javaoperatorsdk.operator.LeaderElectionManager.LEASES_RESOURCE;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 class LeaderElectionManagerTest {
 
-  private LeaderElectionManager leaderElectionManager() {
+  private LeaderElectionManager leaderElectionManager(Optional<Object> selfSubjectReview) {
```
metacosm, line 32: It would be better to not use Optional here but pass the object directly and use `Optional.ofNullable` when resolving the kubernetes client instead.

metacosm (changes requested):

csviri:
Hi @MikeEdgar thx for the PR,

pls run `mvn clean install` to format the code.

csviri:
The integration test: `LeaderElectionChangeNamespaceIT` is also failing, could you pls take a look why? later also I can take a look

MikeEdgar:
> The integration test: `LeaderElectionChangeNamespaceIT` is also failing, could you pls take a look why? later also I can take a look

Thanks, I'll take a look today.

MikeEdgar:
> > The integration test: `LeaderElectionChangeNamespaceIT` is also failing, could you pls take a look why? later also I can take a look
> 
> Thanks, I'll take a look today.

The rule filter predicate on `main` has incorrect grouping that I didn't notice before, that was fixed with the changes. The filter is picking any rule that has the universal/wildcard verb. I'll fix the test in this PR to deploy/remove a role/binding.

https://github.com/operator-framework/java-operator-sdk/blob/8f77f686218172533d968def8017cfd439f955c5/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/LeaderElectionManager.java#L134-L138

MikeEdgar:
The permission check now tests for wildcard `*` for the API group and resource kinds.

csviri (approved):
LGMT!

Thank you!

metacosm (approved):

Issue #2457, opened 2024-06-26
Apply the CRD using the SSA method in LocallyRunOperationExtension.


10000-ki:
https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java#L185

![스크린샷 2024-06-26 오후 2 23 04](https://github.com/operator-framework/java-operator-sdk/assets/76486121/3a662dd1-3650-488d-b419-671286eea999)

Previously, we used the createOrReplace() API, but it is now deprecated. Since the operator itself uses the SSA method as the default, it is better to apply the CRD using the SSA method.

Especially with the CSA method, if users apply the CRD locally and also apply it in tests without a merge patch occurring between them, there can occasionally be issues with lost changes or the need for a force apply.

PR #2458, opened 2024-06-26
Apply the CRD using the SSA method in LocallyRunOperationExtension

10000-ki:
resolved: #2457


operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -182,7 +182,7 @@ public static void applyCrd(String resourceTypeName, KubernetesClient client) {
       var crdString = new String(is.readAllBytes(), StandardCharsets.UTF_8);
       LOGGER.debug("Applying CRD: {}", crdString);
       final var crd = client.load(new ByteArrayInputStream(crdString.getBytes()));
-      crd.createOrReplace();
```
10000-ki, line 185: I have documented the related content in the issue. Is there any specific reason we need to use the CSA method here?

csviri: No, It think this is fine.

csviri (approved):

metacosm (approved):

PR #2470, opened 2024-07-12
refactor: clean up informer configuration classes

metacosm:
- **refactor: move @InformerConfig to more appropriate package**
- **refactor: move InformerConfigHolder to appropriate package**
- **chore: remove unneeded code & dependencies**
- **refactor: InformerConfiguration to InformerEventSourceConfiguration**
- **refactor: rename inner InformerConfigurationBuilder to simply Builder**
- **refactor: rename InformerConfig and associated field to Informer**
- **refactor: rename InformerConfigHolder to InformerConfiguration**


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfigurationOverrider.java
```diff
@@ -30,12 +30,12 @@ public class ControllerConfigurationOverrider<R extends HasMetadata> {
...
   private Duration reconciliationMaxInterval;
   private Map<DependentResourceSpec, Object> configurations;
-  private final InformerConfigHolder<R>.Builder config;
+  private final InformerConfiguration<R>.Builder config;
 
   private ControllerConfigurationOverrider(ControllerConfiguration<R> original) {
     this.finalizer = original.getFinalizerName();
     this.generationAware = original.isGenerationAware();
-    this.config = InformerConfigHolder.builder(original.getResourceClass())
+    this.config = InformerConfiguration.builder(original.getResourceClass())
```
csviri, line 38: Wouldn't be it confusing that there is now an `InformerEventSourceConfiguration` renamed from `InformerConfiguration`, but there is now a new `InformerConfiguration` for something else?
Also an `@Informer` that is basically the new `InformerConfiguration` just represented by annotation?

Not what would be better, just might be confusing.

metacosm: We would definitely need to document it and add it to the javadoc as well so that the information is available contextually. I think this cleans things nicely this way, though, because having multiple things names InformerConfig-something was getting quite confusing, especially when some of them were not even related to informers at all…

csviri (approved):

PR #2497, opened 2024-08-14
improve: replace jenvtest with kube-api-test from fabric8 client

csviri:


operator-framework/pom.xml
```diff
@@ -85,9 +85,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.javaoperatorsdk</groupId>
-      <artifactId>jenvtest</artifactId>
-      <scope>test</scope>
```
metacosm, line 90: Shouldn't it still use the test scope?

csviri: I guess that is inherited from dependency management

metacosm: https://stackoverflow.com/questions/15221299/dependencymanagement-and-scope seems to indicate that putting scopes in `dependencyManagement` is a bad practice…

csviri: hmm, ok np, will move it

metacosm (approved):

metacosm (approved):

PR #2514, opened 2024-08-22
chore: update example of registering an event source

matteriben:
Updated the example of registering an event source and also the link.

Previous example:
- Had broken link
- Returned List instead of Map
- Used name `configMapEventSource `, but was apparently for `Deployment`
- Had apparent copy/paste error around lines 508-509

Updated example:
- Omits some details of the actual linked code
- Increased indentation


metacosm (approved):
Thank you!

PR #2515, opened 2024-08-23
feat: add getPlural method on GroupVersionKind

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
@@ -1,6 +1,7 @@
 package io.javaoperatorsdk.operator.processing;
 
 import java.util.Objects;
+import java.util.Optional;
 
 import io.fabric8.kubernetes.api.Pluralize;
 import io.fabric8.kubernetes.api.model.HasMetadata;
@@ -22,7 +23,7 @@ public class GroupVersionKind {
       this.group = groupAndVersion[0];
       this.version = groupAndVersion[1];
     }
-    this.plural = Pluralize.toPlural(kind);
+    this.plural = null;
     this.apiVersion = apiVersion;
   }
 
@@ -34,7 +35,7 @@ public class GroupVersionKind {
     this.group = group;
     this.version = version;
     this.kind = kind;
-    this.plural = plural != null ? plural : Pluralize.toPlural(kind);
+    this.plural = plural;
     this.apiVersion = (group == null || group.isBlank()) ? version : group + "/" + version;
   }
 
@@ -56,8 +57,16 @@ public class GroupVersionKind {
     return kind;
   }
 
-  public String getPlural() {
-    return plural;
+  /**
+   * Returns the plural form associated with the kind if it has been provided explicitly (either
+   * manually by the user, or determined from the associated resource class definition)
+   *
+   * @return {@link Optional#empty()} if the plural form was not provided explicitly (in which case,
+   *         it could be approximated by using {@link Pluralize#toPlural(String)} on the kind), or
+   *         the plural form if it was provided explicitly
+   */
+  public Optional<String> getPlural() {
+    return Optional.ofNullable(plural);
   }
 
   public String apiVersion() {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
@@ -1,18 +1,20 @@
 package io.javaoperatorsdk.operator.processing;
 
 import java.util.Objects;
-import java.util.Optional;
 
-import io.fabric8.kubernetes.api.Pluralize;
 import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.dependent.kubernetes.GroupVersionKindPlural;
 
 public class GroupVersionKind {
   private final String group;
   private final String version;
   private final String kind;
-  private final String plural;
   private final String apiVersion;
 
+  /**
+   * @deprecated Use {@link GroupVersionKindPlural#gvkFor(String, String)} instead
+   */
+  @Deprecated(forRemoval = true)
   public GroupVersionKind(String apiVersion, String kind) {
     this.kind = kind;
     String[] groupAndVersion = apiVersion.split("/");
@@ -23,26 +25,22 @@ public class GroupVersionKind {
       this.group = groupAndVersion[0];
       this.version = groupAndVersion[1];
     }
-    this.plural = null;
     this.apiVersion = apiVersion;
   }
 
-  public GroupVersionKind(String group, String version, String kind) {
-    this(group, version, kind, null);
+  public static GroupVersionKind gvkFor(Class<? extends HasMetadata> resourceClass) {
+    return GroupVersionKindPlural.gvkFor(resourceClass);
   }
 
-  public GroupVersionKind(String group, String version, String kind, String plural) {
+  /**
+   * @deprecated Use {@link GroupVersionKindPlural#gvkFor(String, String, String)} instead
+   */
+  @Deprecated(forRemoval = true)
+  public GroupVersionKind(String group, String version, String kind) {
     this.group = group;
     this.version = version;
     this.kind = kind;
-    this.plural = plural;
     this.apiVersion = (group == null || group.isBlank()) ? version : group + "/" + version;
-  }
-
-  public static GroupVersionKind gvkFor(Class<? extends HasMetadata> resourceClass) {
-    return new GroupVersionKind(HasMetadata.getGroup(resourceClass),
-        HasMetadata.getVersion(resourceClass), HasMetadata.getKind(resourceClass),
-        HasMetadata.getPlural(resourceClass));
   }
 
   public String getGroup() {
@@ -55,18 +53,6 @@ public class GroupVersionKind {
 
   public String getKind() {
     return kind;
-  }
-
-  /**
-   * Returns the plural form associated with the kind if it has been provided explicitly (either
-   * manually by the user, or determined from the associated resource class definition)
-   *
-   * @return {@link Optional#empty()} if the plural form was not provided explicitly (in which case,
-   *         it could be approximated by using {@link Pluralize#toPlural(String)} on the kind), or
-   *         the plural form if it was provided explicitly
-   */
-  public Optional<String> getPlural() {
-    return Optional.ofNullable(plural);
   }
 
   public String apiVersion() {
@@ -91,8 +77,7 @@ public class GroupVersionKind {
   @Override
   public String toString() {
     return "GroupVersionKind{" +
-        "group='" + group + '\'' +
-        ", version='" + version + '\'' +
+        "apiVersion='" + apiVersion + '\'' +
         ", kind='" + kind + '\'' +
         '}';
   }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
@@ -1,20 +1,19 @@
 package io.javaoperatorsdk.operator.processing;
 
+import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.javaoperatorsdk.operator.processing.dependent.kubernetes.GroupVersionKindPlural;
 
 public class GroupVersionKind {
   private final String group;
   private final String version;
   private final String kind;
   private final String apiVersion;
+  protected final static Map<Class<? extends HasMetadata>, GroupVersionKind> CACHE =
+      new ConcurrentHashMap<>();
 
-  /**
-   * @deprecated Use {@link GroupVersionKindPlural#gvkFor(String, String)} instead
-   */
-  @Deprecated(forRemoval = true)
   public GroupVersionKind(String apiVersion, String kind) {
     this.kind = kind;
     String[] groupAndVersion = apiVersion.split("/");
@@ -29,13 +28,14 @@ public class GroupVersionKind {
   }
 
   public static GroupVersionKind gvkFor(Class<? extends HasMetadata> resourceClass) {
-    return GroupVersionKindPlural.gvkFor(resourceClass);
+    return CACHE.computeIfAbsent(resourceClass, GroupVersionKind::computeGVK);
   }
 
-  /**
-   * @deprecated Use {@link GroupVersionKindPlural#gvkFor(String, String, String)} instead
-   */
-  @Deprecated(forRemoval = true)
+  private static GroupVersionKind computeGVK(Class<? extends HasMetadata> rc) {
+    return new GroupVersionKind(HasMetadata.getGroup(rc),
+        HasMetadata.getVersion(rc), HasMetadata.getKind(rc));
+  }
+
   public GroupVersionKind(String group, String version, String kind) {
     this.group = group;
     this.version = version;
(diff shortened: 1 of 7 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
```diff
@@ -2,12 +2,15 @@
...
 
+import io.fabric8.kubernetes.api.Pluralize;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 
 public class GroupVersionKind {
   private final String group;
   private final String version;
   private final String kind;
+  private final String plural;
+  private final String apiVersion;
```
csviri, line 13: We don't need explicit API Version, it is "group/version".

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
```diff
@@ -2,12 +2,15 @@
...
 import java.util.Objects;
 
+import io.fabric8.kubernetes.api.Pluralize;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 
 public class GroupVersionKind {
   private final String group;
   private final String version;
   private final String kind;
+  private final String plural;
```
csviri, line 12: What is the use case of plural?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
csviri: it is not present on go side either afaik.

xstefank (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
metacosm: This is to avoid re-computing it all the time.

csviri: I don't think we should do that. It you check this is only used when creating the informer, we should not do such optimizations. It would, worth to check but from java 9 there are quite clever optimizations on String concatenation, probably making such optimizations unnecessary.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
metacosm: While the plural can most of the time be inferred from the kind, it could be useful to know the precise plural version associated with resources identified by this GVK, if it is known. In particular, this would be useful for the linked PR (see: https://github.com/quarkiverse/quarkus-operator-sdk/pull/931/files#diff-b780cb662fc5c0a824655ac29ced7dca9efd35a33f771cd3f3a84a46f21b7cebR97 in particular).
The plural form is used in many spots, this is actually what is used to identify the REST endpoint to access when dealing with resources so knowing its precise form (if known) could be useful for pretty much anything that deals with RBACs.
To be clear, this is a convenience method which should be transparent to people not using it.

csviri: Sorry did not see the link. My only concern is, that in JOSDK scop this has no usage whatsoever, that might lead to a confusion. Currently it is an internal technical class with a very well defined scope.  So in that regard basically we have leaking requirements from QOSDK to JOSDK, shouldn't be rather a `GroupVersionKindPlural` class (extending GVK) created in QOSDK, so we have nice between the projects?!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
csviri: Let's take a look on this on opposite direction, this could also be seen also as premature optimization. It brings complexity to the code. (Althoug not significant for sure). If you can show this significantly increases any performance, it's fine. But I really don't think we should do such micro optimizations, especially in case of string concatenations, that are not repeadetly used. 

I mean we could argue that this also has negative side on memory usage. 

At the end it is more about complexity of the code in this case.

csviri: But I might be wrong :)  What do you think?  maybe @xstefank @andreaTP @manusa

metacosm: > Let's take a look on this on opposite direction, this could also be seen also as premature optimization. It brings complexity to the code. (Althoug not significant for sure). If you can show this significantly increases any performance, it's fine. But I really don't think we should do such micro optimizations, especially in case of string concatenations (that is optimized out on JVM nicely) , that are not repeadetly used.

It doesn't bring any complexity: the code that used to be in the getter is used in the constructor instead and, to the contrary, it simplifies both equals and hashCode (because you can just check apiVersion instead of needing to check both group and version), also "optimizing" their execution. So the complexity argument is just wrong, imo.

> We could argue that this also has negative side on memory usage.

That, I agree with but, in this case, I'm sure there are many other spots where we could save more memory than just saving an extra string here.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
metacosm: This isn't a requirement but this is something (what is the plural form?) that someone might want to know when given a GVK. It makes sense on its own, imo. We could rename the class but I think it makes sense as-is, not to mention that keeping the same name would avoid breaking the API and thus requiring waiting for a major version for this feature, while I would like to get that released in a patch release quickly if possible.

csviri: I don't mean renaming the class, but creating a class in QOSDK that extends this. 
It is not used anywhere in JOSDK - I mean plural, it is not in controller runtime gvk either. It might be needed for some completely different purpose in QOSDK to have this information, but then it should be a representation for it in QOSDK not here.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
csviri: > That, I agree with but, in this case, I'm sure there are many other spots where we could save more memory than just saving an extra string here. 

We could do performance optimizations on other spots too, this is not a valid argument.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
metacosm: > Sorry did not see the link. My only concern is, that in JOSDK scope this has no usage whatsoever, that might lead to confusion. 

This wouldn't be the first thing like this, even for stuff that isn't used in QOSDK. That's a property of framework to provide things that are not used within the framework.

csviri: Look this isn't nitpicking, this is a central technical class that I want to keep as clean as possible. GVK plays a important role in dynamic APIs, check the glue operator: https://github.com/csviri/kubernetes-glue-operator
basically all the keys in caches and stuff are built on this. This is reasonable to assume from anyone that builds dynamic stuff. So we really want to have this as clean as possible.  

It is very simple to create a `GroupVersionKindPlural` in QOSDK that simply extends `GroupVersionKind` and the boundaries will be very clean.

> This wouldn't be the first thing like this, even for stuff that isn't used in QOSDK. 

We usually make stuff open for extension, I think we should not leak back notions until is not necessary.

> That's a property of framework to provide things that are not used within the framework.

We tend to have at least Sample / Integrations Test for every feature. Plural is really not needed anywhere in JOSDK atm.

metacosm: > Look this isn't nitpicking, this is a central technical class that I want to keep as clean as possible. GVK plays a important role in dynamic APIs, check the glue operator: https://github.com/csviri/kubernetes-glue-operator basically all the keys in caches and stuff are built on this. This is reasonable to assume from anyone that builds dynamic stuff. So we really want to have this as clean as possible.

You make an argument against an extension to a class that could benefit other projects based on the fact that another project uses that class. That seems rather self-serving. There are no more reason to "help" the Glue operator by keeping GVK "clean" than helping other projects by adding the plural version here.
Moreover, how exactly is adding the plural form (which is central to how resources are handled by kubernetes) "dirtying" the GVK class? It's not like it's adding something that's completely unrelated: plural is derived from kind so it makes sense to have it there, especially when the derivation is not trivial and can't only be inferred from the kind. Classes that use GVK lose nothing by having GVK provide a plural form as well.

> 
> It is very simple to create a `GroupVersionKindPlural` in QOSDK that simply extends `GroupVersionKind` and the boundaries will be very clean.

That is not true because some information might be lost by not providing the plural form. It might actually make sense to make `getPlural` return `Optional<String>` instead to convey that information.

If anything, you could make a point that GVK shouldn't be exposed at all outside of JOSDK. You could also make a point that JOSDK should have used the GVK class from the Fabric8 client instead of creating its own. JOSDK created its own version and exposed it outside of it, that's OK and it's used by other projects, that's also OK. I'm arguing that this class could be made even more useful to outside projects by making the plural form available which JOSDK is often able to compute properly, which might not be the case for external projects that might lack the information JOSDK has.

The goal of a framework is to be useful to external projects, not adhere to some weird concept of "purity". Having plural on GVK increases integration opportunities by exposing information that JOSDK is uniquely positioned to compute correctly.

csviri: What I wanted to point is that the scope of this class now is to identify a Kubernetes object type. That is done by using Group, Version, Kind in Kubernetes. Plural plays a role on completely different places. 

It is completely fine to use GVK for this purpose outside of the project. But by adding the `plural` has nothing to do with this function (thus identifying a type).  That is what I mean clean.  Now you want to add plural, because it fits for some  functionality related to RBAC generation in QOSDK. That completely breaks the purpose of this class.

csviri: I'm fine with having `GroupVersionKindPlural` that extends `GroupVersionKind`  in JOSDK, saying that this is a class that helps in other places. But not by adding plural into `GroupVersionKind` directly.

csviri: see in Kubernetes: **https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupVersionKind**

csviri: note also the other types in that package: https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#pkg-types
with different purpose.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/GroupVersionKind.java
xstefank: IMHO, it could be in the field; this won't take that much space. While I understand Attila's arguments, I more often see this kind of computation cache like Chris is proposing.

csviri: oki, I'm fine with that. Might be good to comment on it, that is for performance reasons.

csviri (changes requested):
Was thinking more about this, to my understanding (correct me if wrong):

- We need `GroupVersionKindPlural` only to generate RBAC for `GenericKubernetesDependentResource` (GKDR) in Quarkus. 
- For example for `InformerEventSource` if we don't use DR we never need the GVKP only GVK. 
- For GKDR we need only (user will specify) if the Plural is special non computable way. 

Therefore thinking this PR is good, but:
-  we should not deprecate GVK, it should be still used for Informer, also for GKDR if no special plural is needed.
- What would be nice is to add an additional constructor to GKDR that takes GVKP. So we make it explicit in the code that is used there.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/generickubernetesresource/generickubernetesdependentresourcemanaged/ConfigMapGenericKubernetesDependent.java
```diff
@@ -24,7 +25,7 @@ public class ConfigMapGenericKubernetesDependent extends
   public static final String KEY = "key";
 
   public ConfigMapGenericKubernetesDependent() {
-    super(new GroupVersionKind("", VERSION, KIND));
```
csviri, line 27: The point here was to not use the class, but demonstrate how to do it classless or "dynamically". Would be nicer with a dedicated custom resource for sure. I just used config map. Pls revert this.

operator-framework/src/test/java/io/javaoperatorsdk/operator/sample/generickubernetesresource/generickubernetesdependentstandalone/ConfigMapGenericKubernetesDependent.java
```diff
@@ -24,7 +25,7 @@ public class ConfigMapGenericKubernetesDependent extends
   public static final String KEY = "key";
 
   public ConfigMapGenericKubernetesDependent() {
-    super(new GroupVersionKind("", VERSION, KIND));
+    super(GroupVersionKind.gvkFor(ConfigMap.class));
```
csviri, line 28: Same as above.

metacosm:
> Therefore thinking this PR is good, but:
> 
>     * we should not deprecate GVK, it should be still used for Informer, also for GKDR if no special plural is needed.

It is deprecated so that it is not used outside of JOSDK anymore. The idea is to expose only GVKP to external users, not GVK since it is only used internally so users will only deal with GVKP, which should only be needed outside of JOSDK via GKDR.

> 
>     * What would be nice is to add an additional constructor to GKDR that takes GVKP. So we make it explicit in the code that is used there.

There is one already.

csviri:
> There is one already.

Sorry missed that. Great.

csviri:
> It is deprecated so that it is not used outside of JOSDK anymore. The idea is to expose only GVKP to external users, not GVK since it is only used internally so users will only deal with GVKP, which should only be needed outside of JOSDK via GKDR.

My point it is that is should not be deprecated, should not be only internal. Either we hide both on APIs, or neither. Since for example for InformerEventSource initialization we should not force GVKP.

csviri:
> > It is deprecated so that it is not used outside of JOSDK anymore. The idea is to expose only GVKP to external users, not GVK since it is only used internally so users will only deal with GVKP, which should only be needed outside of JOSDK via GKDR.
> 
> My point it is that is should not be deprecated, should not be only internal. Either we hide both on APIs, or neither. Since for example for InformerEventSource initialization we should not force GVKP. For that should be GVK available.

In both cases this could be hidden by just accepting. 3 respectively 4 strings instead of GVK / GVKP in constructors, builder methods.

metacosm:
> > It is deprecated so that it is not used outside of JOSDK anymore. The idea is to expose only GVKP to external users, not GVK since it is only used internally so users will only deal with GVKP, which should only be needed outside of JOSDK via GKDR.
> 
> My point it is that is should not be deprecated, should not be only internal. Either we hide both on APIs, or neither. Since for example for InformerEventSource initialization we should not force GVKP. For that should be GVK available.

My conclusion now, looking at the code, is that it doesn't make sense to have both. We should only have one, with the plural available from there. Any other solution doesn't make sense to me.

metacosm:
I believe that all the comments should now be addressed.

csviri (approved):
LGTM, great, thank you!!

PR #2531, opened 2024-09-17
fix: integration test GC timeout on kube 1.31.0

csviri:


metacosm:
It'd be interesting to know why this happens, though, because it feels we're simply addressing the symptoms, not the cause.

metacosm (approved):

csviri:
> It'd be interesting to know why this happens, though, because it feels we're simply addressing the symptoms, not the cause.

Will check the k8s release notes (did not do yet for v1.31 of k8s) but this is not on our side.

PR #2538, opened 2024-09-24
refactor: make shouldUseSSA work with types instead of instances + tests

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -24,6 +24,7 @@ import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResourceFactory;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependent;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResource;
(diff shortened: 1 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -362,29 +361,45 @@ default boolean ssaBasedCreateUpdateMatchForDependentResources() {
...
   /**
-   * This is mostly useful as an integration point for downstream projects to be able to reuse the
-   * logic used to determine whether a given {@link KubernetesDependentResource} should use SSA or
-   * not.
-   *
-   * @param dependentResource the {@link KubernetesDependentResource} under consideration
-   * @return {@code true} if SSA should be used for
-   * @param <R> the dependent resource type
-   * @param <P> the primary resource type
-   * @since 4.9.4
```
xstefank, line 373: I personally don't think that deprecating should mean removing of old JavaDoc but I don't know what's the custom in this project.

xstefank (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
metacosm: I don't think we have one but removing the javadoc makes it even more obvious that people should really be using the other one instead, imo… 😅

metacosm: I've removed the deprecation in the end.

csviri (approved):

Issue #2492, opened 2024-08-13
[Suggestion] Resource Diff Printer


10000-ki:
## Idea

I'd like to propose an idea. 
How about providing a utility function that outputs the diff information to show which fields have changed between the actual resource and the desired resource?

Currently, it seems that such a feature doesn't exist, so when unintended events are generated, we have to log both the actual and desired resources as text and manually compare them field by field to find the differences. This process takes too much time.

## Example

```kotlin
    override fun match(
        actualResource: StatefulSet,
        desired: StatefulSet,
        primary: OpenSearch,
        context: Context<OpenSearch>,
    ): Matcher.Result<StatefulSet> {
        println(ResourceDiffPrinter.printDiff(actualResource, desired))
    }
```

---

```
"""
Diff : 
(actual) -> (desired)
annotations.last-modification-timestamp: "2024-08-13T11:07:04+09:00" -> annotations.last-modification-timestamp: "2024-08-13T12:07:04+09:00" 
spec.replicas: "3" -> spec.replicas: "4"
"""
```

csviri:
Would make sense IMO, I can imagine that as part of `SSABasedGenericKubernetesResourceMatcher`, that could also log this information. 

Do you plan to also create a PR?

10000-ki:
> Do you plan to also create a PR?

@csviri 
Of course! In that case, would it be okay if I work on it and then submit a PR?

10000-ki:
@csviri 
the PR should be created based on the `next` branch right??

metacosm:
> @csviri the PR should be created based on the `next` branch right??

Yes, please.

10000-ki:
done

PR #2542, opened 2024-09-29
feat: print diff resource as yaml

10000-ki:
resolved: #2492 

## AS-IS

Output diff data in text format

```
{data={key1=val1}, metadata={ownerReferences=[{name=kube-root-ca.crt, uid=1ef74cb4-dbbd-45ef-9caf-aa76186594ea, apiVersion=v1, kind=ConfigMap}]}}

{metadata={ownerReferences=[{apiVersion=v1, kind=ConfigMap, name=kube-root-ca.crt, uid=1ef74cb4-dbbd-45ef-9caf-aa76186594ea}]}, data={key1=different value}}
```

- problem : It's difficult to compare

## TO-BE

Output diff data in yml format

```yml
data:
  key1: "val1"
metadata:
  ownerReferences:
  - name: "kube-root-ca.crt"
    uid: "1ef74cb4-dbbd-45ef-9caf-aa76186594ea"
    apiVersion: "v1"
    kind: "ConfigMap"
 
 ---
metadata:
  ownerReferences:
  - apiVersion: "v1"
    kind: "ConfigMap"
    name: "kube-root-ca.crt"
    uid: "1ef74cb4-dbbd-45ef-9caf-aa76186594ea"
data:
  key1: "different value"
```

If the user sorts fields through Jackson settings `.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)`
it will look much better because the keys will be displayed in sorted order.

---

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -104,9 +104,9 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
     removeIrrelevantValues(desiredMap);
 
     if (LoggingUtils.isNotSensitiveResource(desired)) {
-      var actualYml = objectMapper.asYaml(prunedActual);
-      var desiredYml = objectMapper.asYaml(desiredMap);
-      log.debug("Pruned actual: \n {} \n desired: \n {} ", actualYml, desiredYml);
+      var actualYaml = objectMapper.asYaml(prunedActual);
+      var desiredYaml = objectMapper.asYaml(desiredMap);
+      log.debug("Pruned actual: \n {} \n desired: \n {} ", actualYaml, desiredYaml);
     }
 
     return prunedActual.equals(desiredMap);
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -114,9 +114,9 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
     if (log.isDebugEnabled()) {
       var actualYaml = serialization.asYaml(prunedActualMap);
       var desiredYaml = serialization.asYaml(desiredMap);
-      log.debug("Pruned actual: \n {} \n desired: \n {} ", actualYaml, desiredYaml);
+      log.debug("Pruned actual yaml: \n {} \n desired yaml: \n {} ", actualYaml, desiredYaml);
     } else {
-      log.debug("Pruned actual: \n {} \n desired: \n {} ", prunedActualMap, desiredMap);
+      log.debug("Pruned actual map: \n {} \n desired map: \n {} ", prunedActualMap, desiredMap);
     }
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -110,7 +110,8 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
     return prunedActual.equals(desiredMap);
   }
 
-  private void logDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap, KubernetesSerialization serialization) {
+  private void logDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap,
+      KubernetesSerialization serialization) {
     if (log.isDebugEnabled()) {
       var actualYaml = serialization.asYaml(prunedActualMap);
       var desiredYaml = serialization.asYaml(desiredMap);
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -104,7 +104,9 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
     removeIrrelevantValues(desiredMap);
 
     if (LoggingUtils.isNotSensitiveResource(desired)) {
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -104,7 +104,9 @@ public boolean matches(R actual, R desired, Context<?> context) {
     removeIrrelevantValues(desiredMap);
 
     if (LoggingUtils.isNotSensitiveResource(desired)) {
-      log.debug("Pruned actual: \n {} \n desired: \n {} ", prunedActual, desiredMap);
+      var actualYml = objectMapper.asYaml(prunedActual);
+      var desiredYml = objectMapper.asYaml(desiredMap);
+      log.debug("Pruned actual: \n {} \n desired: \n {} ", actualYml, desiredYml);
```
10000-ki, line 109: It is possible to use a library or write code to compare YAML files. However, as the size of the YAML file increases, the cost of processing grows. Therefore, I have not added this functionality here and simply focused on outputting the data.

If deemed necessary, I can consider adding it in the future.

10000-ki:
As mentioned above, users can configure Jackson to sort field keys when logging YAML files by enabling the `SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS setting`  
This ensures that the YAML fields are ordered alphabetically, making the log output more readable and consistent

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -104,7 +104,9 @@ public boolean matches(R actual, R desired, Context<?> context) {
     removeIrrelevantValues(desiredMap);
 
     if (LoggingUtils.isNotSensitiveResource(desired)) {
-      log.debug("Pruned actual: \n {} \n desired: \n {} ", prunedActual, desiredMap);
+      var actualYaml = objectMapper.asYaml(prunedActual);
```
metacosm, line 107: I would protect these calls with a check to whether debug is enabled so as to not inferred the YAML serialization cost in that situation.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
metacosm: I'm not sure that the sorting would make that much of a difference compared to serializing to YAML in the first place…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
10000-ki: > I would protect these calls with a check to whether debug is enabled so as to not inferred the YAML serialization cost in that situation.

metacosm (approved):
Thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
10000-ki: @metacosm 
https://github.com/operator-framework/java-operator-sdk/pull/2542/commits/72911d651efe4268dade07d05e5a5377d1c28ab3

fixed it in this commit

thank you!!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -104,12 +104,23 @@ public boolean matches(R actual, R desired, Context<?> context) {
...
   }
 
+  private void logDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap,
+      KubernetesSerialization serialization) {
+    if (log.isDebugEnabled()) {
+      var actualYaml = serialization.asYaml(prunedActualMap);
+      var desiredYaml = serialization.asYaml(desiredMap);
+      log.debug("Pruned actual yaml: \n {} \n desired yaml: \n {} ", actualYaml, desiredYaml);
+    } else {
+      log.debug("Pruned actual map: \n {} \n desired map: \n {} ", prunedActualMap, desiredMap);
```
csviri, line 120: Not sure if I follow this. "If debug is not enabled we log something on debug" ?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
metacosm: Ah yes, missed that, though there should actually be no output since `debug` is checked internally (iirc) but it does make sense to remove the else branch altogether, indeed.

csviri: What I mean is that the else block never logs any message, since only called if no debug is enabled but logs on debug level.

10000-ki: @csviri 

Ah understand

https://github.com/operator-framework/java-operator-sdk/pull/2542/commits/96920abbe7322c0fa20f36c22065ebec8c067b87

I thought it would be better to just remove this part, so I modified it as shown above. What do you think?

csviri (approved):
LGTM

metacosm (approved):

csviri:
Thank you @10000-ki !

bachmanity1:
Hi @10000-ki. Thanks for the PR! 

I think printing the whole actual and desired resource is a bit inconvenient since we still need to copy and paste it into other tools to get the real diff. How about improving the logging logic to directly print the actual diff, like shown below?

```
2024-10-02 14:30:29,489 1 i.j.o.p.d.k.SSABasedGenericKubernetesResourceMatcher [DEBUG] Diff between actual and desired state for resource: Deployment with name: test in namespace: default is: 
--- 
+++ 
@@ -15,1 +15,1 @@
-      - image: "nginx:1.17.0"
+      - image: "nginx:1.17.1"

```

metacosm:
Hi @bachmanity1,

I was kind of thinking the same thing but seeing the resource in YAML format is already helpful. We'd happily consider a PR to output the actual diffs! Keep in mind, though, that we want to keep this output as efficient as possible so it should probably be only activated via a flag.

PR #2544, opened 2024-10-02
feat: improve diff logging

bachmanity1:
Based on [this](https://github.com/operator-framework/java-operator-sdk/pull/2542#issuecomment-2387669854) comment.

Diff is calculated and printed only if there is a mismatch between actual and desired state and logger is enabled for the debug level. Here is the sample output:
 
```
2024-10-02 16:38:42,233 1 i.j.o.p.d.k.SSABasedGenericKubernetesResourceMatcher [DEBUG] Diff between actual and desired state for resource: ConfigMap with name: test1 in namespace: default is: 
--- 
+++ 
@@ -2,3 +2,3 @@
 data:
-  key1: "val1"
+  key1: "different value"
 metadata:

```

### DIFF
--- a/operator-framework-core/pom.xml
+++ b/operator-framework-core/pom.xml
@@ -58,7 +58,6 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.fabric8</groupId>
--- a/operator-framework-core/pom.xml
+++ b/operator-framework-core/pom.xml
@@ -15,6 +15,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>io.github.java-diff-utils</groupId>
+      <artifactId>java-diff-utils</artifactId>
+    </dependency>
+    <dependency>
       <groupId>io.fabric8</groupId>
       <artifactId>kubernetes-client</artifactId>
       <exclusions>
@@ -58,6 +62,7 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.fabric8</groupId>
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -120,11 +120,15 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
 
   private String getDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap,
       KubernetesSerialization serialization) {
-    var actualLines = serialization.asYaml(sortMap(prunedActualMap)).lines().toList();
-    var desiredLines = serialization.asYaml(sortMap(desiredMap)).lines().toList();
+    var actualYaml = serialization.asYaml(sortMap(prunedActualMap));
+    var desiredYaml = serialization.asYaml(sortMap(desiredMap));
+    if (log.isTraceEnabled()) {
+      log.trace("Pruned actual resource: \n{} \ndesired resource: \n{} ", actualYaml, desiredYaml);
+    }
 
-    var patch = DiffUtils.diff(actualLines, desiredLines);
-    List<String> unifiedDiff = DiffUtils.generateUnifiedDiff("", "", actualLines, patch, 1);
+    var patch = DiffUtils.diff(actualYaml.lines().toList(), desiredYaml.lines().toList());
+    List<String> unifiedDiff =
+        DiffUtils.generateUnifiedDiff("", "", actualYaml.lines().toList(), patch, 1);
     return String.join("\n", unifiedDiff);
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -124,7 +124,7 @@ public class SSABasedGenericKubernetesResourceMatcher<R extends HasMetadata> {
     var desiredLines = serialization.asYaml(sortMap(desiredMap)).lines().toList();
 
     var patch = DiffUtils.diff(actualLines, desiredLines);
-    List<String> unifiedDiff = DiffUtils.generateUnifiedDiff("", "", actualLines, patch, 0);
+    List<String> unifiedDiff = DiffUtils.generateUnifiedDiff("", "", actualLines, patch, 1);
     return String.join("\n", unifiedDiff);
   }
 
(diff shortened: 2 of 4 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -103,20 +105,65 @@ public boolean matches(R actual, R desired, Context<?> context) {
 
     removeIrrelevantValues(desiredMap);
 
-    if (LoggingUtils.isNotSensitiveResource(desired)) {
-      logDiff(prunedActual, desiredMap, objectMapper);
+    var matches = prunedActual.equals(desiredMap);
+
+    if (!matches && LoggingUtils.isNotSensitiveResource(desired)) {
+      var diff = getDiff(prunedActual, desiredMap, objectMapper);
```
csviri, line 111: The diff should be calculated only if debug enabled.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -103,20 +105,65 @@ public boolean matches(R actual, R desired, Context<?> context) {
...
+    return matches;
   }
 
-  private void logDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap,
+  private String getDiff(Map<String, Object> prunedActualMap, Map<String, Object> desiredMap,
       KubernetesSerialization serialization) {
-    if (log.isDebugEnabled()) {
-      var actualYaml = serialization.asYaml(prunedActualMap);
-      var desiredYaml = serialization.asYaml(desiredMap);
-      log.debug("Pruned actual yaml: \n {} \n desired yaml: \n {} ", actualYaml, desiredYaml);
```
csviri, line 118: We still might want to log the actual resources.

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
bachmanity1: when should we log actual resource?

csviri: I mean it is good that  we log the diff, but some might be interested in the actual resources (I mean both "actual" and desired yaml). It might be useful to log that too?

bachmanity1: > What do you think?

how about logging actual resource in case when  logger is enabled for the trace level? or should we introduce new configuration parameter?

csviri: Trace level is fine by me, we could have there both the resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
10000-ki: > The diff should be calculated only if debug enabled.

That's right, this costs a lot.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
bachmanity1: ok, I enabled printing actual and desired resources when trace is enabled.

operator-framework-core/pom.xml
```diff
@@ -58,7 +58,6 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
```
10000-ki, line 61: I am not sure if it is right to include `assertj` as a dependency when build time rather than in a test environment.

csviri: Yes, if this could be done nicer, would be great.

bachmanity1: possible alternative is to use https://github.com/java-diff-utils/java-diff-utils

bachmanity1: Let me know if I should remove the last commit or not.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -103,20 +107,68 @@ public boolean matches(R actual, R desired, Context<?> context) {
...
+      if (item instanceof Map) {
+        Map<String, Object> mapItem = (Map<String, Object>) item;
+        sortedList.add(sortMap(mapItem));
+      } else if (item instanceof List) {
+        sortedList.add(sortList((List<?>) item));
+      } else {
+        sortedList.add(item);
+      }
     }
+    return sortedList;
```
10000-ki, line 171: What if you could create a separate utility class like `DiffGenerator` and write unit tests for it?

operator-framework-core/pom.xml
10000-ki: > possible alternative is to use https://github.com/java-diff-utils/java-diff-utils

This looks good to me, the library itself doesn't seem heavy.
It looks like there may be another opportunity to use this library for other purposes in the future.

metacosm: Yes, I was going to comment on using this lib instead of assertj. Thanks!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -103,20 +107,68 @@ public boolean matches(R actual, R desired, Context<?> context) {
...
+    }
+
+    var patch = DiffUtils.diff(actualYaml.lines().toList(), desiredYaml.lines().toList());
+    List<String> unifiedDiff =
+        UnifiedDiffUtils.generateUnifiedDiff("", "", actualYaml.lines().toList(), patch, 1);
+    return String.join("\n", unifiedDiff);
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map<String, Object> sortMap(Map<String, Object> map) {
```
csviri, line 139: Yes, for this pls add unit tests.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
csviri: makes sense

csviri (approved):
LGTM thx!

metacosm (approved):
Thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
bachmanity1: I'm sorry, I was planning to separate this into util class but I got a little busy lately.

csviri: Np, feel free to create a separate PR. thx

PR #2546, opened 2024-10-08
improve: naming target resource selector method

csviri:
Also, add a sample usage.


csviri:
Managed term is already used for dependent resources, for me this was confusing when I was checking this again, IMO this is a bit better name.

metacosm (approved):

PR #2548, opened 2024-10-10
feat(ci): split integration tests execution by category

metacosm:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,37 @@
+name: Build
+
+env:
+  MAVEN_ARGS: -V -ntp -e
+
+on:
+  workflow_call:
+
+jobs:
+  integration_tests:
+    strategy:
+      matrix:
+        java: [ 17, 21 ]
+        kubernetes: [ 'v1.28.14', 'v1.29.9','1.30.5', '1.31.1' ]
+        it-category: [ 'baseapi', 'dependent', 'workflow' ]
+        httpclient: [ 'vertx', 'jdk', 'jetty' ]
+    uses: ./.github/workflows/integration-tests.yml
+    with:
+      java-version: ${{ matrix.java }}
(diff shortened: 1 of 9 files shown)


csviri:
@metacosm could youd tell what exactly we try to solve with this?
The main concern is the runtime length?

metacosm:
Runtime length but also easier diagnosing of issues with shorter logging.

csviri:
Well, now we have maybe shorter log files, but 81 runs, which is not nice on it's own. So the problem that this solved is the large log files? 

At least for me is not the problem to download it in raw format, and do a search on the unit test name.

csviri (approved):
I mean I don't have big objection against this, just let's have a rational behind it at least in comments.

metacosm:
> Well, now we have maybe shorter log files, but 81 runs, which is not nice on it's own. So the problem that this solved is the large log files?

Why do you think more runs are an issue? (I mean I have some opinions on this but I'd like to hear your rationale)
We now run more tests (especially now all the integration tests are run using all HTTP client implementations). We could, of course, decide that we don't need to do this to reduce the number of different runs.

> 
> At least for me is not the problem to download it in raw format, and do a search on the unit test name.

Shorter logs make it easier / faster to see what's going on from CI directly instead of having to manually check it out on your computer by downloading the logs.

csviri:
> Why do you think more runs are an issue? (I mean I have some opinions on this but I'd like to hear your rationale)

Not a big issues just simpler to see the structure if is shorter / simpler. 

> We now run more tests (especially now all the integration tests are run using all HTTP client implementations)

TBH this would omit in future if we see there are no more issues with this, since it's not our responsibility to test the client. But ok for now. We should probably test eventually just with vertex (AFAIK that is what fabric8 recommends)

csviri:
Maybe we could limit http client flavor tests to low level API.

PR #2549, opened 2024-10-10
fix: race condition in workflow reconciler

metacosm:


metacosm:
> Which one is here caused the issue? AFAIK concurrentHashMap should cover those usages here

I thought so as well but it seems that the results might not be updated correctly with the proper visibility, resulting in conditions tests not seeing the actual results (e.g. in the test that was failing, it seemed like the unmet condition on DR2 was not seen when checking whether DR4 should be reconciled)

csviri (approved):

PR #2555, opened 2024-10-18
wip: attempt to simplify workflows handling

metacosm:
- **refactor: remove now unneeded NameSetter interface**
- **refactor: simplify workflow handling**
- **refactor: initial work to combine ManagedWorkflow and regular one**


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/DefaultManagedWorkflow.java
```diff
@@ -10,37 +9,25 @@
...
+    final var alreadyResolved = new HashMap<String, DependentResourceNode>(orderedSpecs.size());
+    for (DependentResourceSpec spec : orderedSpecs) {
+      final var node = new UnresolvedDependentResourceNode(spec);
+      alreadyResolved.put(spec.getName(), node);
+      spec.getDependsOn()
+          .forEach(depend -> node.addDependsOnRelation(alreadyResolved.get(depend)));
     }
+
+    this.workflow = new DefaultWorkflow<>(alreadyResolved.values(), false, hasCleaner);
+    this.orderedSpecs = orderedSpecs; // todo: remove
```
xstefank, line 30: I don't understand these todos in this class. Do you still plan to work on them in this PR?

xstefank (commented):
Other than todos I don't see anything wrong.

csviri:
@metacosm I would prefer to review this together, I fail to see how this simplifies things.

csviri:
Might be also better to do such refactorings step by step with limited scoped PR- if possible.

metacosm:
This isn't done yet so, indeed, this doesn't really simplify things at the moment… also, I'm not sure the individual steps will make any difference since they will only make sense once the whole is completed (if it can actually be achieved as expected, which I'm not yet sure of as this is still rather exploratory).

metacosm:
Replaced by #2566

PR #2558, opened 2024-10-25
docs: release 5 blogpost

csviri:
The first version of the blogpost, will be probably extended with subsequent PR-s


metacosm:
It might be better to simply cherry-pick the changes in a new branch than trying to rebase since it seems rather complex for some reason. It also looks like some changes got into that branch that shouldn't have made it in there (adding the `controller-namespace-deletion` sample, for example).

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/2583

PR #2559, opened 2024-10-25
improve: workflow builder naming

csviri:


csviri:
@metacosm a bit nit on naming, let me know if sounds better also to you ears :)

csviri:
Was thining about more variation, I know how did you mean it but, just having the same wording as in annotations (where it makes sense) is a bit better IMO.

( https://github.com/operator-framework/java-operator-sdk/blob/1da6558a9f7804a2210a24e6d0400494e9911353/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/Dependent.java#L79 )

metacosm:
I prefer the fluent version but using the same wording as the annotation attribute makes sense.

metacosm (approved):

PR #2565, opened 2024-10-28
Fix infinite resource updates due to canonical format conversion of resource requirements

### LINKED ISSUE #2509: CPU resource unit comparison error
## Bug Report
In Kubernetes, some resource units are automatically converted to a more convenient format internally. For example:

```yml
resources:
    limits:
      cpu: 2000m
```

Even if the CPU is set to 2000m, when you retrieve this configuration using kubectl, you might see:

```yml
resources:
    limits:
      cpu: 2
```

This indicates that the unit has been converted.

The problem arises when an Operator sets the CPU unit as `2000m`, but the Kubernetes API server keeps converting it back to `2`. This leads to a situation where, during the `DependentResource.match` process, the Operator and the API server identify the resources as different, causing `match to return false`.

스크린샷 2024-08-21 오후 5 07 14

The reason for this issue is that when comparing the actual and desired states, the values are checked through simple string comparison.

As a result, the operator keeps sending patch requests, causing the `resourceVersion` to increase infinitely.

#### Environment

`$ java -version`

17

`$ kubectl version`

v1.24

#### Possible Solution

```kotlin
Quantity("2000m") == Quantity("2")
```

In fact, when using `Quantity` in fabric8 for comparison, it can be determined that these two values are the same. It might be better to use `Quantity` for resource comparisons. However, given the current structure, this doesn't seem straightforward. Do you think it might be possible?

Donnerbart:
Fixes #2509

### Code coverage

* Branch `next`
  image

* Branch `bugfix/2509-ssa-matcher-quantity/next`
  image

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/ResourceRequirementsSanitizer.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/ResourceRequirementsSanitizer.java
@@ -1,0 +1,100 @@
+package io.javaoperatorsdk.operator.processing.dependent.kubernetes;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.GenericKubernetesResource;
+import io.fabric8.kubernetes.api.model.PodTemplateSpec;
+import io.fabric8.kubernetes.api.model.Quantity;
+import io.fabric8.kubernetes.api.model.ResourceRequirements;
+
+/**
+ * Sanitizes the {@link ResourceRequirements} in the containers of a pair of {@link PodTemplateSpec}
+ * instances.
+ * <p>
+ * When the sanitizer finds a mismatch in the structure of the given templates, before it gets to
+ * the nested resource limits and requests, it returns early without fixing the actual map. This is
+ * an optimization because the given templates will anyway differ at this point. This means we do
+ * not have to attempt to sanitize the resources for these use cases, since there will anyway be an
+ * update of the K8s resource.
+ * <p>
+ * The algorithm traverses the whole template structure because we need the actual and desired
+ * {@link Quantity} instances to compare their numerical amount. Using the
+ * {@link GenericKubernetesResource#get(Map, Object...)} shortcut would need to create new instances
+ * just for the sanitization check.
+ */
+class ResourceRequirementsSanitizer {
+
+  static void sanitizeResourceRequirements(final Map<String, Object> actualMap,
+      final PodTemplateSpec actualTemplate, final PodTemplateSpec desiredTemplate) {
+    if (actualTemplate == null || desiredTemplate == null) {
+      return;
+    }
+    if (actualTemplate.getSpec() == null || desiredTemplate.getSpec() == null) {
+      return;
+    }
+    sanitizeResourceRequirements(actualMap, actualTemplate.getSpec().getInitContainers(),
+        desiredTemplate.getSpec().getInitContainers(), "initContainers");
+    sanitizeResourceRequirements(actualMap, actualTemplate.getSpec().getContainers(),
+        desiredTemplate.getSpec().getContainers(), "containers");
+  }
+
+  private static void sanitizeResourceRequirements(final Map<String, Object> actualMap,
+      final List<Container> actualContainers, final List<Container> desiredContainers,
+      final String containerPath) {
+    int containers = desiredContainers.size();
+    if (containers == actualContainers.size()) {
+      for (int containerIndex = 0; containerIndex < containers; containerIndex++) {
+        var desiredContainer = desiredContainers.get(containerIndex);
+        var actualContainer = actualContainers.get(containerIndex);
+        if (!desiredContainer.getName().equals(actualContainer.getName())) {
+          return;
+        }
+        sanitizeResourceRequirements(actualMap, actualContainer.getResources(),
+            desiredContainer.getResources(),
+            containerPath, containerIndex);
+      }
+    }
+  }
+
+  private static void sanitizeResourceRequirements(final Map<String, Object> actualMap,
+      final ResourceRequirements actualResource, final ResourceRequirements desiredResource,
+      final String containerPath, final int containerIndex) {
+    if (desiredResource == null || actualResource == null) {
+      return;
+    }
+    sanitizeQuantities(actualMap, actualResource.getRequests(), desiredResource.getRequests(),
+        containerPath, containerIndex, "requests");
+    sanitizeQuantities(actualMap, actualResource.getLimits(), desiredResource.getLimits(),
+        containerPath, containerIndex, "limits");
+  }
+
+  @SuppressWarnings("unchecked")
+  private static void sanitizeQuantities(final Map<String, Object> actualMap,
+      final Map<String, Quantity> actualResource, final Map<String, Quantity> desiredResource,
+      final String containerPath, final int containerIndex, final String quantityPath) {
+    Optional.ofNullable(
+        GenericKubernetesResource.get(actualMap, "spec", "template", "spec", containerPath,
+            containerIndex, "resources", quantityPath))
+        .map(Map.class::cast)
+        .filter(m -> m.size() == desiredResource.size())
+        .ifPresent(m -> actualResource.forEach((key, actualQuantity) -> {
+          var desiredQuantity = desiredResource.get(key);
+          if (desiredQuantity == null) {
+            return;
+          }
+          // check if the string representation of the Quantity instances is equal
+          if (actualQuantity.getAmount().equals(desiredQuantity.getAmount())
+              && actualQuantity.getFormat().equals(desiredQuantity.getFormat())) {
+            return;
+          }
+          // check if the numerical amount of the Quantity instances is equal
+          if (actualQuantity.equals(desiredQuantity)) {
+            // replace the actual Quantity with the desired Quantity to prevent a resource update
+            m.replace(key, desiredQuantity.toString());
+          }
+        }));
+  }
+}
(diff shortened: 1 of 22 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/ResourceRequirementsSanitizer.java
```diff
@@ -0,0 +1,100 @@
...
+ * The algorithm traverses the whole template structure because we need the actual and desired
+ * {@link Quantity} instances to compare their numerical amount. Using the
+ * {@link GenericKubernetesResource#get(Map, Object...)} shortcut would need to create new instances
+ * just for the sanitization check.
+ */
+class ResourceRequirementsSanitizer {
+
+  static void sanitizeResourceRequirements(final Map<String, Object> actualMap,
+      final PodTemplateSpec actualTemplate, final PodTemplateSpec desiredTemplate) {
+    if (actualTemplate == null || desiredTemplate == null) {
```
Donnerbart, line 32: I'm not 100% sure if all of these `null` checks are needed. I've added them for all fields that have no default value in their declaration.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -92,31 +95,26 @@ public boolean matches(R actual, R desired, Context<?> context) {
...
-    if (LoggingUtils.isNotSensitiveResource(desired)) {
-      log.trace("Original actual: \n {} \n original desired: \n {} ", actual, desiredMap);
-    }
-
     var prunedActual = new HashMap<String, Object>(actualMap.size());
     keepOnlyManagedFields(prunedActual, actualMap,
         managedFieldsEntry.getFieldsV1().getAdditionalProperties(),
         objectMapper);
 
+    var desiredMap = objectMapper.convertValue(desired, Map.class);
```
csviri, line 104: Why was this moved?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcherTest.java
```diff
@@ -119,52 +122,132 @@ void addedLabelInDesiredMakesMatchFail() {
...
     assertThat(sortedMap.keySet()).containsExactly("a", "b", "c");
 
-    Map<String, Object> sortedNestedMap = (Map<String, Object>) sortedMap.get("b");
+    var sortedNestedMap = (Map<String, Object>) sortedMap.get("b");
     assertThat(sortedNestedMap.keySet()).containsExactly("y", "z");
   }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"sample-sts-volumeclaimtemplates-desired.yaml",
+      "sample-sts-volumeclaimtemplates-desired-with-status.yaml",
```
csviri, line 169: I'm a little bit more fan of building such resources in java, would be more expressive a bit, but that is jut a nit :)

csviri (approved):
Overall this looks good to me, just made some really minor. comments.
In future maybe we could have sanitizers customizable, like have the ability to pass only few sanitizers if there will be more.

Thank you @Donnerbart !!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
Donnerbart: The original order of method invocations jumped between pruning/sanitizing the actual and the desired resource. Just from the method names it was partially ambiguous to me which resource is changed (actual/desired, actualMap/desiredMap/prunedActual).

So I re-arranged the code into these logical groups:
* get managed fields (or throw)
* get serialization
* prepare actual -> actualMap -> prunedActual
* desired -> desiredMap
* compare prunedActual with desiredMap

By removing some of the many newlines, these logical blocks are also visually grouped together.

Donnerbart: After going through the code there might be one undesired change: When `keepOnlyManagedFields()` throws an exception, we won't see the `log.trace("Original actual: \n {} \n original desired: \n {} ", actual, desiredMap)` anymore.

If that needs to be fixed, I would suggest to move the whole `desired -> desiredMap` block before the actual block:
```java
    var objectMapper = context.getClient().getKubernetesSerialization();

    var desiredMap = objectMapper.convertValue(desired, Map.class);
    if (LoggingUtils.isNotSensitiveResource(desired)) {
      log.trace("Original actual:\n {}\n original desired:\n {}", actual, desiredMap);
    }

    var actualMap = objectMapper.convertValue(actual, Map.class);
    sanitizeState(actual, desired, actualMap);
    var prunedActual = new HashMap<String, Object>(actualMap.size());
    keepOnlyManagedFields(prunedActual, actualMap,
        managedFieldsEntry.getFieldsV1().getAdditionalProperties(),
        objectMapper);

    var matches = prunedActual.equals(desiredMap);
```

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcherTest.java
Donnerbart: We're using the fabric8 builder classes in our operator tests a lot, but sometimes I think the nested builders are harder to read (heavily depending on the code formatter settings). For example, using `.withNewTemplate()` the nested `.withNewSpec()/endSpec()` are often intended on the same level as the outer `.withNewSpec()/endSpec()` from the STS, which I personally find much harder to read than a well formatted XML file. This can be mitigated by using `.withTemplate(new PodTemplateSpecBuilder()...build())` though.

I just had a look how e.g. `sample-sts-volumeclaimtemplates.yaml` would look like in Java code:
```java
    var actualStatefulSet = new StatefulSetBuilder()
        .withNewMetadata()
        .withName("test")
        .withUid("50913e35-e855-469f-bec6-3e8cd2607ab4")
        .withManagedFields(new ManagedFieldsEntryBuilder()
            .withManager("controller")
            .withOperation("Apply")
            .withApiVersion("apps/v1")
            .withTime("2024-10-24T19:15:25Z")
            .withNewFieldsV1()
             ???
            .endFieldsV1()
            .build())
        .endMetadata()
        .withNewSpec()
        .withReplicas(1)
        .withNewSelector().addToMatchLabels("app", "test-app").endSelector()
        .withServiceName("nginx-service")
        .withTemplate(new PodTemplateSpecBuilder()
            .withNewMetadata().addToLabels("app", "test-app").endMetadata()
            .withNewSpec()
            .addNewContainer()
            .withName("nginx")
            .withImage("nginx:1.17.0")
            .addNewPort().withContainerPort(80).endPort()
            .addNewVolumeMount()
            .withName("persistent-storage")
            .withMountPath("/usr/share/nginx/html")
            .endVolumeMount()
            .endContainer()
            .endSpec()
            .build())
        .addToVolumeClaimTemplates(new PersistentVolumeClaimBuilder()
            .withNewMetadata().withName("persistent-storage").endMetadata()
            .withNewSpec()
            .addToAccessModes("ReadWriteOnce")
            .withNewResources()
            .addToRequests("storage", new Quantity("1Gi"))
            .endResources()
            .withStorageClassName("standard")
            .withVolumeMode("Filesystem")
            .endSpec()
            .withNewStatus()
            .withPhase("Pending")
            .endStatus()
            .build())
        .endSpec()
        .build();
```
I gave up on the managed fields. Without looking up an example in the fabric8 client I have no idea how to use that builder. I already had to guess how to write that correctly in XML, by adapting this from another resource file :sweat_smile:

So even as a heavy user of those builder classes, I think for these complex resources XML files are easier to write and even to compare (IDEA -> mark two files -> diff).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
Donnerbart: Or we could also align what we print in that trace log:
```java
    var objectMapper = context.getClient().getKubernetesSerialization();
    var actualMap = objectMapper.convertValue(actual, Map.class);
    var desiredMap = objectMapper.convertValue(desired, Map.class);
    if (LoggingUtils.isNotSensitiveResource(desired)) {
      log.trace("Original actual:\n {}\n original desired:\n {}", actualMap, desiredMap);
    }

    sanitizeState(actual, desired, actualMap);
    var prunedActual = new HashMap<String, Object>(actualMap.size());
    keepOnlyManagedFields(prunedActual, actualMap,
        managedFieldsEntry.getFieldsV1().getAdditionalProperties(),
        objectMapper);

    var matches = prunedActual.equals(desiredMap);
```
I think this is even more readable.

csviri:
I guess this needs a rebase. Sorry I'm on out this and partially next week. @metacosm  from my side this PR can go in after rebase, pls review it also.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
Donnerbart: Since I anyway had to rebase, I changed the code to the last variant to restore the order of execution, but switched to `actualMap` for the trace logging (I think it'll be much easier to read if both resources are logged in the same format).

Donnerbart:
Rebase is done.

csviri (approved):

metacosm:
Sorry, I should have merged this before rebasing next. Apologies for the extra work that was bad scheduling on my part. 🙍

metacosm (approved):
Thank you!

PR #2569, opened 2024-11-06
feat: allow manually specifying CRDs in test extension

### LINKED ISSUE #2561: Confusing hard-coded v1 expectation for CRD Yamls
#### What did you do?

I wrote a reconciler test along the lines of the one generated by the quickstart. The one difference between the quickstart and what I'm trying to do is I want to use a CRD-first approach (rather than Java-first approach) for my operator. So I'm generating by CRD classes using the Fabric8 plugin. This seems to be working fine.

I expected to be able to use the `LocallyRunOperatorExtension`, maybe needing to tell it where my CRD YAML lives.

The test dies with this stacktrace:

```
java.lang.IllegalStateException: Cannot apply CRD yaml: /META-INF/fabric8/kafka-proxies.kroxylicious.io-v1.yml

	at io.javaoperatorsdk.operator.junit.LocallyRunOperatorExtension.applyCrd(LocallyRunOperatorExtension.java:192)
	at io.javaoperatorsdk.operator.junit.LocallyRunOperatorExtension.applyCrd(LocallyRunOperatorExtension.java:169)
	at io.javaoperatorsdk.operator.junit.LocallyRunOperatorExtension.before(LocallyRunOperatorExtension.java:157)
	at io.javaoperatorsdk.operator.junit.AbstractOperatorExtension.beforeEachImpl(AbstractOperatorExtension.java:148)
	at io.javaoperatorsdk.operator.junit.AbstractOperatorExtension.beforeEach(AbstractOperatorExtension.java:79)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: java.lang.IllegalStateException: Cannot find CRD at /META-INF/fabric8/kafka-proxies.kroxylicious.io-v1.yml
	at io.javaoperatorsdk.operator.junit.LocallyRunOperatorExtension.applyCrd(LocallyRunOperatorExtension.java:180)
	... 6 more
```

The expectation that my CRD lives in `/META-INF/fabric8/` is a bit annoying, but bearable. 
A bigger problem is this seems to assume that my CRD is supporting a `v1`, which is not the case (I'm starting with a `v1alpha1`).
There doesn't seem to be any way to override this behaviour. So it looks like I'm forced to declare my CRD in a file with a really confusing name.

#### Environment

N/A

#### Possible Solution

Allow passing the CRD resource name in the call to `LocallyRunOperatorExtension.withReconciler()` would appear to be a reasonable API for allowing people to use the extension without hardcoding naming assumptions about CRD resources.

metacosm:
This is useful when using a contract-first approach where the Java
classes are generated from the CRD instead of the reverse.

Fixes #2561

### DIFF
--- a/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
+++ b/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
@@ -45,7 +45,7 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
   private final List<LocalPortForward> localPortForwards;
   private final List<Class<? extends CustomResource>> additionalCustomResourceDefinitions;
   private final Map<Reconciler, RegisteredController> registeredControllers;
-  private final Map<Class<? extends CustomResource>, String> crdMappings;
+  private final Map<String, String> crdMappings;
 
   private LocallyRunOperatorExtension(
       List<ReconcilerSpec> reconcilers,
@@ -60,7 +60,7 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
       Consumer<ConfigurationServiceOverrider> configurationServiceOverrider,
       Function<ExtensionContext, String> namespaceNameSupplier,
       Function<ExtensionContext, String> perClassNamespaceNameSupplier,
-      Map<Class<? extends CustomResource>, String> crdMappings) {
+      Map<String, String> crdMappings) {
     super(
         infrastructure,
         infrastructureTimeout,
@@ -207,10 +207,16 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
         ref.controllerConfigurationOverrider.accept(oconfig);
       }
 
+      final var unapplied = new HashMap<>(crdMappings);
+      final var resourceTypeName = ReconcilerUtils.getResourceTypeName(resourceClass);
       // only try to apply a CRD for the reconciler if it is associated to a CR
       if (CustomResource.class.isAssignableFrom(resourceClass)) {
-        applyCrd(resourceClass);
+        applyCrd(resourceTypeName);
+        unapplied.remove(resourceTypeName);
       }
+
+      // apply yet unapplied CRDs
+      unapplied.keySet().forEach(this::applyCrd);
 
       var registeredController = this.operator.register(ref.reconciler, oconfig.build());
       registeredControllers.put(ref.reconciler, registeredController);
@@ -229,7 +235,11 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
    * @param crClass the custom resource class for which we want to apply the CRD
    */
   public void applyCrd(Class<? extends CustomResource> crClass) {
-    final var path = crdMappings.get(crClass);
+    applyCrd(ReconcilerUtils.getResourceTypeName(crClass));
+  }
+
+  public void applyCrd(String resourceTypeName) {
+    final var path = crdMappings.get(resourceTypeName);
     if (path != null) {
       try (InputStream inputStream = new FileInputStream(path)) {
         applyCrd(inputStream, path, getKubernetesClient());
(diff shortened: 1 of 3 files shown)


operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.awaitility.Awaitility.await;
+
+public class CRDMappingInTestExtensionIT {
+  private final KubernetesClient client = new KubernetesClientBuilder().build();
+
+  @RegisterExtension
+  LocallyRunOperatorExtension operator =
+      LocallyRunOperatorExtension.builder()
+          .withReconciler(new TestReconciler())
+          .withCRDMapping(TestCR.class, "src/test/crd/test.crd")
```
csviri, line 31: Regarding this API, not sure if we need to specify the class here, that is rather limiting, if someone for example uses dynamic API does not have the class. It would be enough to have just the path to cover even that possibility.

csviri (changes requested):

operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
metacosm: This feature is meant for people who use the contract-first approach so they would have the class handy and most likely wouldn't use generic resources but I guess we could use the CRD name instead of the CR class.

csviri: I mean we can make it even cover even that case (So dynamic resources handling with `GenericKubernetesResource`)

Why we need a name? cannot we have an API just with a path?

metacosm: Because that would imply loading the file to check the name, which I would rather avoid. It also allows to potentially override a generated CRD with a manual one if needed (though, this probably is more of a corner case).

csviri: I'm sorry, but I don't get this. Why does it need a name? We want to simply apply a CRD from a file here, like `kubectl apply` would.  

Also, why would someone want to override a generated one? (Even if do, for that would rather create a separate method to cancel apply of a CRD, but only if a user comes with a real example)

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -152,11 +207,17 @@ protected void before(ExtensionContext context) {
...
       }
 
+      final var unapplied = new HashMap<>(crdMappings);
+      final var resourceTypeName = ReconcilerUtils.getResourceTypeName(resourceClass);
       // only try to apply a CRD for the reconciler if it is associated to a CR
-      if (CustomResource.class.isAssignableFrom(config.getResourceClass())) {
-        applyCrd(config.getResourceTypeName());
+      if (CustomResource.class.isAssignableFrom(resourceClass)) {
+        applyCrd(resourceTypeName);
+        unapplied.remove(resourceTypeName);
```
csviri, line 215: Could you ellaborate what is the problem this is solving?

operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
csviri: Hmm, but should we rather parse the CRD and index it based on API version. And check if that is the same as resources API version mentioned in reconciler. (But always apply it even if there is no reconciler corresponding to it). This also delegates responsibility to the user to name it correctly, what we could definitely simplify in the background seamlessly.

csviri: So basically I think we should parse the crd to have nicer API here.

metacosm: Yes, please open a PR with that change.

csviri (approved):

PR #2571, opened 2024-11-12
wip: attempt to simplify workflows handling

metacosm:
- **refactor: remove now unneeded NameSetter interface**
- **refactor: simplify workflow handling**
- **refactor: initial work to combine ManagedWorkflow and regular one**


metacosm:
While it could be interesting to share more code between managed and manual workflows, there also isn't a need for it from a user's point of view at this point, so shutting this effort down for now.

PR #2572, opened 2024-11-12
fix: concurrent modification when getting event sources (v5)

csviri:


csviri:
This is a fix for v5, will backport it also for main

metacosm (approved):

PR #2574, opened 2024-11-14
improve: simpler api for adding additional CRD file

csviri:


operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
```diff
@@ -28,7 +28,7 @@ public class CRDMappingInTestExtensionIT {
   LocallyRunOperatorExtension operator =
       LocallyRunOperatorExtension.builder()
           .withReconciler(new TestReconciler())
-          .withCRDMapping("tests.crd.example", "src/test/crd/test.crd")
+          .withAdditionalCRD("src/test/resources/crd/test.crd")
```
metacosm, line 31: This was explicitly not put in `resources` to check that the file could be accessed from a "random" part of the file system, not just from the classpath.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -138,6 +144,17 @@ private static void applyCrd(InputStream is, String path, KubernetesClient clien
     }
   }
 
+  public static CustomResourceDefinition parseCrd(String path, KubernetesClient client) {
+    try (InputStream is = new FileInputStream(path)) {
+      return (CustomResourceDefinition) client.load(new ByteArrayInputStream(is.readAllBytes()))
```
metacosm, line 149: This should handle better weird cases: if there are more than just one resource in the file, if the CRD is not a v1 spec, if the resource in the file is not a CRD are the most obvious cases where this might fail.

metacosm (changes requested):

operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
csviri: I see, isn't it trivial?  I would rather stick with this, but no strong opinion, if you insist can change it back.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
csviri: Update it to handle more CRD in file. I don't think we should support other CRD version that v1, note that this is not the resource definition version inside, CRD itself is stable for very long time.

csviri: Yes, agree, we should fix that, thus not generate those anymore.

PR #2578, opened 2024-11-15
refactor: remove now unneeded NameSetter interface

metacosm:


metacosm:
This change actually cannot work without requiring a constructor taking the name as parameter for all `DependentResource` implementations.

PR #2587, opened 2024-11-18
Method signature mismatched with javadoc and implementation

robobario:
The javadoc  of `ManagedDependentResourceContext#put` describe it as returning `Optional`. The implementation returns an `Optional` but casts it to T, so it is currently a compile error to write:

```
Optional<String> result = new ManagedDependentResourceContext().put("key", "value");
```
Using the method as declared results in a ClassCastException:
```
DefaultManagedDependentResourceContext context = new DefaultManagedDependentResourceContext();
context.put("key", "value");
String actual = context.put("key", "valueB");
```
fails with:
```
java.lang.ClassCastException: class java.util.Optional cannot be cast to class java.lang.String (java.util.Optional and java.lang.String are in module java.base of loader 'bootstrap')
```

The user can cast to get at the Optional like `Optional<String> actual = (Optional<String>) (Object) context.put("key", "value");`

So this PR updated the javadoc and implementation to match the method declaration and return the object-or-null rather than an Optional.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
@@ -3,11 +3,16 @@ package io.javaoperatorsdk.operator.api.reconciler.dependent.managed;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.javaoperatorsdk.operator.processing.dependent.workflow.WorkflowCleanupResult;
 import io.javaoperatorsdk.operator.processing.dependent.workflow.WorkflowReconcileResult;
 
 @SuppressWarnings("rawtypes")
 public class DefaultManagedDependentResourceContext implements ManagedDependentResourceContext {
+  private static final Logger log =
+      LoggerFactory.getLogger(DefaultManagedDependentResourceContext.class);
   public static final Object RECONCILE_RESULT_KEY = new Object();
   public static final Object CLEANUP_RESULT_KEY = new Object();
   private final ConcurrentHashMap attributes = new ConcurrentHashMap();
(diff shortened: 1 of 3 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
```diff
@@ -21,11 +21,11 @@ public <T> Optional<T> get(Object key, Class<T> expectedType) {
 
   @Override
   @SuppressWarnings("unchecked")
-  public <T> T put(Object key, T value) {
+  public <T> Optional<T> put(Object key, T value) {
     if (value == null) {
-      return (T) Optional.ofNullable(attributes.remove(key));
+      return (Optional<T>) Optional.ofNullable(attributes.remove(key));
```
robobario, line 26: the cast and suppressions is ugly, if a mismatched type had been `put` in prior then the user would eventually get a classcast exception when the `Optional<String>` they were hoping for contains a different type.

I wonder if the type should become first class (part of the map key) so you can:

```
put("key", Long.class, 1L);
put("key", String.class, "hi");
get("key", String.class) -> Optional("hi");
get("key", Long.class) -> Optional(1L);
```
to sidestep type mismatch issues. But it might be a bit of a surprising API that it can have different values for a key.

Or we could return `Optional<?>` and put the typechecking work on the caller?

metacosm:
This should target the `next` branch because this would be a breaking API change. Alternatively, we would need to deprecate the existing methods and provide new implementations but this cannot go as-is in `main`.

robobario:
pushed up creating a new method and deprecating the old, a bit clunky since `put` is a nicer name than the new one.

metacosm:
Upon further review, looking at the history of the modifications, I believe that this API got broken at some point in the past and we should probably conform to the `Map`-like API and change the documentation accordingly instead of returning Optional. I have made the required changes but I cannot push to your branch, @robobario.

metacosm:
@robobario see #2595 for the changes that I think should be included as part of this work.

robobario:
great, thanks @metacosm . Have updated and cleaned up the git history (and renamed a test method)

metacosm (approved):

PR #2595, opened 2024-11-19
fix: put should return previous value instead of Optional

metacosm:
- **refactor: add unit tests for DefaultManagedDependentResourceContext**
- **refactor: add putOrRemove to DefaultManagedDependentResourceContext**
- **refactor: deprecate ManagedDependentResourceContext#put**
- **fix: implementation of put should return value instead of Optional**

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
@@ -21,11 +21,9 @@ public class DefaultManagedDependentResourceContext implements ManagedDependentR
 
   @Override
   @SuppressWarnings("unchecked")
+  @Deprecated(forRemoval = true)
   public <T> T put(Object key, T value) {
-    if (value == null) {
-      return (T) Optional.ofNullable(attributes.remove(key));
-    }
-    return (T) Optional.ofNullable(attributes.put(key, value));
+    return (T) putOrRemove(key, value);
   }
 
   @Override
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
@@ -29,6 +29,15 @@ public class DefaultManagedDependentResourceContext implements ManagedDependentR
   }
 
   @Override
+  @SuppressWarnings("unchecked")
+  public <T> Optional<T> putOrRemove(Object key, T value) {
+    if (value == null) {
+      return (Optional<T>) Optional.ofNullable(attributes.remove(key));
+    }
+    return (Optional<T>) Optional.ofNullable(attributes.put(key, value));
+  }
+
+  @Override
   @SuppressWarnings("unused")
   public <T> T getMandatory(Object key, Class<T> expectedType) {
     return get(key, expectedType).orElseThrow(() -> new IllegalStateException(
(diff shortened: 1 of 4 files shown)


metacosm:
Extension to #2587 to properly (according to me) fix the implementation to properly return the previous value as the implementation was done instead of returning `Optional` as the javadoc suggests (the implementation had been changed in a refactoring without the doc being properly updated).

metacosm:
Obsolete.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
```diff
@@ -22,10 +27,26 @@ public <T> Optional<T> get(Object key, Class<T> expectedType) {
...
+          ") for key (" + key +
+          ") was not of type " + value.getClass() +
+          ". This might indicate an issue in your code. If not, use put(" + key +
+          ", null) first to remove the previous value.");
     }
-    return (T) Optional.ofNullable(attributes.put(key, value));
+    return (T) previous;
+  }
+
+  // only for testing purposes
```
xstefank, line 47: Seems this is not only used in testing.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
```diff
@@ -22,10 +27,26 @@ public <T> Optional<T> get(Object key, Class<T> expectedType) {
   @Override
   @SuppressWarnings("unchecked")
   public <T> T put(Object key, T value) {
+    Object previous;
     if (value == null) {
-      return (T) Optional.ofNullable(attributes.remove(key));
+      return (T) attributes.remove(key);
```
xstefank, line 32: IMHO, there should be a reset method. This seems unintuitive to me.

csviri: agree, or simply a `remove`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
metacosm: The whole purpose of this method is to be overridden in the test so that we can check that the logging really happens without having to set up a complex logging fixture just for that purpose.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContext.java
metacosm: Was debating that, indeed. Not sure how many people would actually need this so decided against it (people would usually put stuff in the context and get it back instead of removing them so I think this is an edge case).

csviri: That is true, that the scope of this is the actual reconciliation. So maybe we don't even need remove at all. But if we do would be nice to have it explicit.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/DefaultManagedDependentResourceContextTest.java
```diff
@@ -0,0 +1,106 @@
...
+    final var prior = "value";
+    context.put("key", prior);
+    String actual = context.put("key", "valueB");
+    assertThat(actual).isEqualTo(prior);
+  }
+
+  @Test
+  void putNewValueThrowsExceptionIfTypesDiffer() {
+    // to check that we properly log things without setting up a complex fixture
+    final String[] messages = new String[1];
```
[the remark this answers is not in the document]
metacosm: Unless I am mistaken, it does. Simply using a final String doesn't work since there is no way for Java to check that it is always assigned and therefore complains that you're assigning a value to a final variable in the overridden method. Not using a final String won't work either because this is effectively a capture so the variable needs to be final so the only way that I know of to do this is to use a final array and change the element.

PR #2599, opened 2024-11-20
improve: mapping from annotation depends on type

csviri:
compiles


csviri:
If this is merged, I will do a separate PR that improved on the API, namely that `SecondaryToPrimary` mapper will have a setter for primary type, so it can be set under the hood, so user won't have to do it.

xstefank:
@csviri some rebase went wrong :)

csviri:
> @csviri some rebase went wrong :)

This happens when we reabase new PRs on next, but after the next is rebased again on main. Then only trivial solution is just (squash+) cherry-pick commits on a new branch. Maybe would be good to sync when we do rebases of next on top of main.  Ideally if there are no complex PRs targeting next. cc @metacosm

PR #2601, opened 2024-11-20
feat: return result when calling workflow explicitly

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowCleanupResult.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowCleanupResult.java
@@ -1,30 +1,22 @@
 package io.javaoperatorsdk.operator.processing.dependent.workflow;
 
 import java.util.List;
-import java.util.Map;
 
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
 @SuppressWarnings("rawtypes")
-public class WorkflowCleanupResult extends WorkflowResult {
-  private Boolean allPostConditionsMet;
+public interface WorkflowCleanupResult extends WorkflowResult {
+  WorkflowCleanupResult EMPTY = new WorkflowCleanupResult() {};
 
-  WorkflowCleanupResult(Map<DependentResource, Detail<?>> results) {
-    super(results);
+  default List<DependentResource> getDeleteCalledOnDependents() {
+    return List.of();
   }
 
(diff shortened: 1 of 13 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowCleanupResult.java
```diff
@@ -1,30 +1,22 @@
...
 
 import java.util.List;
-import java.util.Map;
 
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
 @SuppressWarnings("rawtypes")
-public class WorkflowCleanupResult extends WorkflowResult {
-  private Boolean allPostConditionsMet;
+public interface WorkflowCleanupResult extends WorkflowResult {
```
xstefank, line 8: Can this kind of change go in not a major release? This is a public class so it will break user code if someone extended it originally and now it's an interface.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileResult.java
```diff
@@ -1,32 +1,28 @@
...
 
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
 @SuppressWarnings("rawtypes")
-public class WorkflowReconcileResult extends WorkflowResult {
+public interface WorkflowReconcileResult extends WorkflowResult {
```
xstefank, line 9: Same here.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowCleanupResult.java
metacosm: Theoretically, yes. Practically, these classes are not meant to be extended by the user, they're objects that the SDK returns as part of its processing. The only case where I could potentially imagine a user extending these classes would be in tests but even so, that would be quite a stretch, in my opinion…

csviri: This targets next, what will be a major release.

metacosm: Also, yes! Forgot about that! 🤦😅

PR #2540, opened 2024-09-26
chore: update to Fabric8 7.0-SNAPSHOT for testing

metacosm:
Signed-off-by: Chris Laprun <claprun@redhat.com>

PR #2612, opened 2024-11-30
Fix CRD-Generator tests and examples

baloo42:
Fix CRD-Generator tests and examples for #2540

See also https://github.com/fabric8io/kubernetes-client/issues/6683

/cc @metacosm

### DIFF
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
@@ -43,7 +43,7 @@ public class CRDMappingInTestExtensionIT {
   @Group("crd.example")
   @Version("v1")
   @Kind("Test")
-  private static class TestCR extends CustomResource<Void, Void> implements Namespaced {
+  public static class TestCR extends CustomResource<Void, Void> implements Namespaced {
   }
 
   @ControllerConfiguration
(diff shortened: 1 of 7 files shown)


operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
```diff
@@ -43,7 +43,7 @@ void correctlyAppliesManuallySpecifiedCRD() {
   @Group("crd.example")
   @Version("v1")
   @Kind("Test")
-  private static class TestCR extends CustomResource<Void, Void> implements Namespaced {
+  public static class TestCR extends CustomResource<Void, Void> implements Namespaced {
```
metacosm, line 46: Is this change necessary? I think we used different visibility on purpose to check if this was possible to do…

operator-framework/src/test/java/io/javaoperatorsdk/operator/config/runtime/TestCustomResource.java
```diff
@@ -7,5 +7,5 @@
 
 @Group("sample.javaoperatorsdk")
 @Version("v1")
-class TestCustomResource extends CustomResource<Void, Void> implements Namespaced {
+public class TestCustomResource extends CustomResource<Void, Void> implements Namespaced {
```
metacosm, line 10: Same as above

operator-framework/src/test/java/io/javaoperatorsdk/operator/CRDMappingInTestExtensionIT.java
baloo42: Found the root cause here https://github.com/fabric8io/kubernetes-client/blob/main/crd-generator/api-v2/src/main/java/io/fabric8/crdv2/generator/CustomResourceInfo.java#L168

`getDeclaredConstructor()` does only work with public classes.

PR #2616, opened 2024-12-03
chore: fabri8 client to v7.0.0

csviri:


metacosm:
This needed more clean-up, imo, in particular to remove the snapshots repos as we talked about recently.

PR #2618, opened 2024-12-04
docs: migration guide as blogpost refactor

csviri:
The idea is that we document changes as a blogpost only


csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/2632

PR #2639, opened 2024-12-16
improve: optional workflow result

csviri:


csviri:
see discussion here: https://github.com/operator-framework/java-operator-sdk/issues/2625#issuecomment-2545356470

metacosm (commented):
That would be an API breaking change, though…

csviri:
Since we have just beta release that should be fine.

metacosm (approved):

PR #2640, opened 2024-12-16
bump: minikube version and add k8s version 1.32.0

csviri:


metacosm (commented):
All other versions should be updated as well, imo, and maybe remove 1.29?

csviri:
Yep, will take a look

csviri:
updates based on this: https://kubernetes.io/releases/
1.29 is still there.

metacosm (approved):

PR #2661, opened 2025-01-08
chore: bump spotless 2.44.2 and format code

csviri:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
@@ -97,19 +97,8 @@ public class Utils {
     return getBooleanFromSystemPropsOrDefault(DEBUG_THREAD_POOL_ENV_KEY, false);
   }
 
-  static boolean getBooleanFromSystemPropsOrDefault(String propertyName, boolean defaultValue) {
-    var property = System.getProperty(propertyName);
-    if (property == null) {
-      return defaultValue;
-    } else {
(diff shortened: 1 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
```diff
@@ -97,19 +97,8 @@ public static boolean debugThreadPool() {
...
-      property = property.trim().toLowerCase();
-      return switch (property) {
-        case "true" -> true;
-        case "false" -> false;
-        default -> defaultValue;
-      };
-    }
-  }
+  static boolean getBooleanFromSystemPropsOrDefault(String propertyName,
+      boolean defaultValue) {var property=System.getProperty(propertyName);if(property==null){return defaultValue;}else{property=property.trim().toLowerCase();return switch(property){case"true"->true;case"false"->false;default->defaultValue;};}}
```
metacosm, line 101: Hmm, that doesn't seem right

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/Utils.java
csviri: wow, interesting, not sure, will take a look a look later, this might be a bug in spotless

metacosm: Yes, there seems something wrong with 2.44

PR #2680, opened 2025-02-04
fix: explicit workflow invovation cleanup issue

csviri:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
@@ -194,7 +194,8 @@ public class Controller<P extends HasMetadata>
               WorkflowCleanupResult workflowCleanupResult = null;
 
               // The cleanup is called also when explicit invocation is true, but the cleaner is not
-              // implemented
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/Controller.java
```diff
@@ -195,7 +195,7 @@ public DeleteControl execute() throws Exception {
 
               // The cleanup is called also when explicit invocation is true, but the cleaner is not
               // implemented
-              if (managedWorkflow.hasCleaner() || !explicitWorkflowInvocation) {
+              if (managedWorkflow.hasCleaner() && !explicitWorkflowInvocation) {
```
xstefank, line 198: TBH, I miss the context here, but if I understand the comment above correctly it should be:
```suggestion
              if (managedWorkflow.hasCleaner() || explicitWorkflowInvocation) {
```

csviri: Updated the PR and also the comment. It's a bit more complicated than that.

xstefank (approved):

metacosm (approved):

PR #2683, opened 2025-02-10
refactor: move spotless to a profile active only in the root directory

xstefank:
This allows running maven in submodules, for instance to run only some tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,38 +314,6 @@
     </pluginManagement>
     <plugins>
       <plugin>
-        <groupId>com.diffplug.spotless</groupId>
-        <artifactId>spotless-maven-plugin</artifactId>
-        <configuration>
-          <pom>
-            <includes>
-              <include>pom.xml</include>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -362,6 +330,50 @@
     </plugins>
   </build>
   <profiles>
+    <profile>
+      <id>spotless</id>
```
csviri, line 334: Why is this under a profile?

pom.xml
```diff
@@ -362,6 +330,50 @@
...
+      <activation>
+        <file>
+          <exists>contributing</exists>
```
xstefank, line 337: @csviri because of this. Basically, only run spotless if the `contrinbuting` dir is correctly set. So I can do for instance: 

```
cd operator-framework
mvn clean install -Pintegration-tests -Dtest=InformerRemoteClusterIT

metacosm (approved):

PR #2687, opened 2025-02-17
fix: allow keeping deleted CRDs in test with configuration

xstefank:


operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -43,6 +43,8 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(LocallyRunOperatorExtension.class);
   private static final int CRD_DELETE_TIMEOUT = 1000;
+  private static final LinkedList<AppliedCRD> appliedCRDs = new LinkedList<>();
```
csviri, line 46: What is the reason using LinkedList here? Just since those are large objects?

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -43,6 +43,8 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
...
+  private static final boolean deleteCRDs = Boolean.parseBoolean(System.getProperty("testsuite.deleteCRDs", "true"));
```
csviri, line 47: Shouldn't be these rather here as other configs?
[Config](https://github.com/operator-framework/java-operator-sdk/blob/7519920c13f10764e355f88adf998c80ff056987/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java#L336)

csviri (changes requested):

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
xstefank: LinkedList is only needed to keep FIFO ordering of applied CRDs, but I don't think it's necessary. Why large? AppliedCRD is only two Strings - https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java#L333

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
xstefank: I don't think we want to configure this per Test, but rather for the whole testsuite. I think that per test it can be confusing.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
csviri: The order should not be relevant for CRDs, so it is rather confusing. I would lean more to just have a List.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
csviri: Hmm, good point,
I'm working this this now:
https://github.com/operator-framework/java-operator-sdk/issues/2688
And here the CRD's will have to be managed manually, but fur sure be at least one step where the CRD is updated. But that should be fine too, so yes, if we see that this canont cover such use cases when in particular tests user wants to manage use cases we can still add a property also to builder.

xstefank:
@csviri refactored it to ArrayDeque. I think we can merge this.

csviri:
@xstefank  This is just test, it has zero implications, there fore in these places I usually prefer using the well known data structures like ArrayList. This first time somebody going to read that part of code will ask the question why is there a Dequeu. (So might be in this case also reason to make a comment about it). But fine by me.

let's wait until @metacosm  approves it too, then we can merge. The agreement is both of us approve PRs.

xstefank:
just changed to LIFO as it makes more sense IMO.

csviri:
> just changed to LIFO as it makes more sense IMO.

If we would like to be absolutely correct, Set should be used, to indicate that the order does not matter. IMHO.

xstefank:
and it is Set :)

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -8,11 +8,16 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
+import java.util.ArrayDeque;
```
metacosm, line 11: Seems like imports need to be optimized…

metacosm (approved):

Issue #2694, opened 2025-02-21
failed invoking InformerEventSource event handler: null: java.lang.NullPointerException


matteriben:
I'm updating to version 5. I ran an existing test that creates many custom resources which need to be reconciled and noticed that I would occasionally see these errors in the logs.

```
2025-02-21 12:11:43,724 ERROR [io.fab.kub.cli.inf.imp.cac.SharedProcessor] (-1324526451-pool-1-thread-3) v1/configmaps failed invoking InformerEventSource{resourceClass: ConfigMap} event handler: null: java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at io.quarkus.vertx.mdc.provider.LateBoundMDCProvider.putLocal(LateBoundMDCProvider.java:147)
	at io.quarkus.vertx.mdc.provider.LateBoundMDCProvider.put(LateBoundMDCProvider.java:63)
	at org.jboss.logmanager.MDC.put(MDC.java:90)
	at org.slf4j.impl.Slf4jMDCAdapter.put(Slf4jMDCAdapter.java:32)
	at org.slf4j.MDC.put(MDC.java:123)
	at io.javaoperatorsdk.operator.processing.MDCUtils.addResourceInfo(MDCUtils.java:36)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.javaoperatorsdk.operator.processing.event.EventProcessor.submitReconciliationExecution(EventProcessor.java:128)
	at io.javaoperatorsdk.operator.processing.event.EventProcessor.handleMarkedEventForResource(EventProcessor.java:119)
	at io.javaoperatorsdk.operator.processing.event.EventProcessor.handleEvent(EventProcessor.java:109)
	at io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource.lambda$propagateEvent$4(InformerEventSource.java:230)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource.propagateEvent(InformerEventSource.java:220)
	at io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource.onAddOrUpdate(InformerEventSource.java:175)
	at io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource.onUpdate(InformerEventSource.java:140)
[...]

PR #2695, opened 2025-02-22
fix: prevent null pointer exception

matteriben:
Prevent null pointer exception described in issue #2694.


csviri:
the problem should be covered here:
https://github.com/operator-framework/java-operator-sdk/pull/2696

if no objections will close this PR, but thank you for the effort!

PR #2699, opened 2025-02-23
feat: enable deactivation of MDC logging

metacosm:


csviri (approved):
Could you pls ellaborate the reason / use case of this change?

LGTM

metacosm:
We've had, iirc, two issues recently that involved MDC logging in some way and I forgot we even had it. Not sure many people actually do and/or use this, and, as this has a potential (small) impact, I figured we should probably enable deactivating that feature and maybe, in a future release, disable it by default, especially if we add support for OpenTelemetry for instance.

csviri:
Not sure if openTelemtry will affect logging. But this information is very useful, and easy to use just by adding it to the logging pattern. IMO it should be enabled by default (as it is now).

PR #2704, opened 2025-02-25
fix: spotless plugin activation issue

### LINKED ISSUE #2697: Spotless plugin activation issue
Now the spotless plugin does not activates on main.
It was moved recently under a profile that should activate for the rood dir only:

```
 <profile>
      <id>spotless</id>
      <activation>
        <file>
          <exists>contributing</exists>
        </file>
      </activation>
      <build>
        <plugins>
          <plugin>
            <groupId>com.diffplug.spotless</groupId>
            <artifactId>spotless-maven-plugin</artifactId>
            <configuration>
              <pom>

 ...rest omitted  

```

But even if I run explicitly `mvn spotless:apply` it does not reformat code.  Check is not working either
that is why PRs are not catching this issue (` ./mvnw ${MAVEN_ARGS} spotless:check --file pom.xml ` )

xstefank:
Fixes #2697


contributing/eclipse-google-style.xml
```diff
@@ -65,7 +65,7 @@
 <setting id="org.eclipse.jdt.core.formatter.comment.format_line_comments" value="true"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_closing_angle_bracket_in_type_arguments" value="do not insert"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_enum_declarations" value="insert"/>
-<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="false"/>
```
xstefank, line 68: @csviri @metacosm this is the only change I had to do in order for spotless to not mess things up. But it unrolled a lot of lines.

xstefank:
I propose to move to this https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#google-java-format which would remove the need for a custom config file and should follow latest google styles - https://github.com/google/google-java-format/releases. If agreed, I can make a new PR with this move so we avoid back and forth formatting changes. I really don't like that we can't keep the same format options because spotless changed.

contributing/eclipse-google-style.xml
csviri: I would rather stay on an older version while is this fixed in spotless, or is it not expected to be fixed?

csviri:
> I propose to move to this https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#google-java-format which would remove the need for a custom config file and should follow latest google styles - https://github.com/google/google-java-format/releases. If agreed, I can make a new PR with this move so we avoid back and forth formatting changes. I really don't like that we can't keep the same format options because spotless changed.

could you pls do a separate PR to see how that would look like?

xstefank:
@csviri 

google-java-format - https://github.com/operator-framework/java-operator-sdk/pull/2706
palantir - https://github.com/operator-framework/java-operator-sdk/pull/2707

PR #2706, opened 2025-02-26
fix: spotless plugin (googleJavaFormat)

xstefank:


pom.xml
```diff
@@ -327,53 +327,43 @@
...
+              <include>./**/pom.xml</include>
+            </includes>
+            <sortPom>
+              <expandEmptyElements>false</expandEmptyElements>
+            </sortPom>
+          </pom>
+          <java>
+            <googleJavaFormat>
+              <reflowLongStrings>true</reflowLongStrings>
+              <formatJavadoc>false</formatJavadoc>
```
csviri, line 346: Why format javadoc is false?

csviri:
@metacosm @xstefank  I'm fine with basically both palantir and google format, not sure what are the major differences. If the palantir really has some major advantages over google.

pom.xml
xstefank: it was in the examples, but sure, why not

csviri:
> @metacosm @xstefank I'm fine with basically both palantir and google format, not sure what are the major differences. If the palantir really has some major advantages over google.

Given that google has bigger community we might want to choose that over palantir

csviri:
But if we could fix eclipse that would be the best :)

xstefank:
@csviri so are we going with this google-format then? :) Can I close the other PRs?

csviri (approved):
I'm fine with this one, we should check if the links in the documentation are still correct, probably not, but that can be a separate PR.

csviri:
let's wait for @metacosm , what he thinks

xstefank:
I updated links in this PR. Did I miss something?

xstefank:
@csviri I'll do a pass over docs in a follow up PR.

csviri:
@metacosm can we merge this? kinda blocks the work on other issues, if I want to avoid conflicts

metacosm:
Going to check.

metacosm:
Are there options we can tweak, like the line length?

xstefank:
@metacosm I don't think so - https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/java/GoogleJavaFormat.java

We are giving away flexibility for easier config.

metacosm:
OK, not too fond of the rather short line length but so be it…

metacosm (approved):

PR #2708, opened 2025-02-26
fix: spotless plugin (downgrade to 2.43.0)

xstefank:


csviri (approved):

csviri:
For some reason check format fails

PR #2711, opened 2025-02-27
fix: less minikube downloads and starts in CI

### LINKED ISSUE #2698: Investigate if we want/need to reduce the numbers of tests we run
What I would propose is also take a look how this would look like, if we would just remove the testing of various client flavors, and remove the integration test categories as a separate PR and maybe compare with this approach. (Can try later just quite sick atm)

_Originally posted by @csviri in https://github.com/operator-framework/java-operator-sdk/issues/2691#issuecomment-2670748515_

xstefank:
Fixes #2698

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,83 +11,11 @@ on:
         required: true
 
 jobs:
-  set_up_kubernetes:
-    name: Set up Kubernetes ${{ inputs.kube-version }}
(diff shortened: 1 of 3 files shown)


metacosm:
This removes the parallelism of the tests and make the PR checking way slower so I'm not in favor of this solution.

xstefank:
Ok, no strong opinions here.

PR #2722, opened 2025-03-09
improve: more strict delete condition

csviri:
This way it is easier to reason about the resource deletion.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/KubernetesResourceDeletedCondition.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/KubernetesResourceDeletedCondition.java
@@ -27,14 +27,24 @@ public class KubernetesResourceDeletedCondition implements Condition<HasMetadata
       Context<HasMetadata> context) {
     var optionalResource = dependentResource.getSecondaryResource(primary, context);
     if (optionalResource.isEmpty()) {
-      logger.debug(
-          "Resource not found in cache, considering it deleted. "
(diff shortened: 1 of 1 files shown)


xstefank (approved):
OK, but isn't this potentially breaking change?

csviri:
> OK, but isn't this potentially breaking change?

Not really, it is a bit change in the behavior, but rather the previous one was confusing.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/KubernetesResourceDeletedCondition.java
```diff
@@ -4,14 +4,6 @@
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResource;
 
-/**
```
metacosm, line 7: We should update this documentation accordingly, not just remove it.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/KubernetesResourceDeletedCondition.java
csviri: I removed since, from the code is very trivial, so self descriptive. Not sure if this needs to be explicitly documented.

csviri: kk will add back docs.

Issue #2723, opened 2025-03-09
Owner references are not correctly resolved when primary resource has empty group


mnk:
#### What did you do?

When implementing a workflow where the primary resource is a standard K8s kind (ConfigMap)
I was unable to get the secondary resources from the context.

#### What did you expect to see?

I expected to be able to fetch the secondary resources via `context.getSecondaryResource`

#### What did you see instead? Under which circumstances?

The secondary resources are always null

#### Environment

Running latest from main branch 5.0.4-SNAPSHOT 

#### Possible Solution

The problem seems to be caused by a bug in the apiVersion matching. I will post a fix proposal as PR.

mnk:
Temporary workaround merged to java-operator-sdk and fix for root cause also merged to kubernetes-client

PR #2724, opened 2025-03-09
Fix apiVersion match for standard K8s resources

mnk:
The HasMetadata.getApiVersion(primaryResourceType) method return "/v1" for the ConfigMap type. Since the K8s owner reference uses "v1" instead, the secondary resource of a primary ConfigMap will not be tracked correctly.

Proposed fix for #2723

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
@@ -83,9 +83,10 @@ public class Mappers {
 
   public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerReferences(
       String apiVersion, String kind, boolean clusterScope) {
+    String correctApiVersion = apiVersion.startsWith("/") ? apiVersion.substring(1) : apiVersion;
     return resource ->
         resource.getMetadata().getOwnerReferences().stream()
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
```diff
@@ -83,9 +83,10 @@ public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerRefer
 
   public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerReferences(
       String apiVersion, String kind, boolean clusterScope) {
+    String correctApiVersion = apiVersion.startsWith("/") ? apiVersion.substring(1) : apiVersion;
```
mnk, line 86: This is probably not the right way to fix the problem. Maybe `HasMetadata.getApiVersion` should just return "v1"?

csviri: Yes, this seems to be a bug in fabric8 client, pls create an issue also there.

mnk: I created a PR for fabric8: https://github.com/fabric8io/kubernetes-client/pull/6941
Feel free to propose another fix if it is not what you were thinking.

xstefank (approved):

csviri:
thank you @mnk !

PR #2726, opened 2025-03-09
improve: remove seamless type change

csviri:
The curret implementation changed the underlying type, what is not we expect when passing an instance
will create issues when some compares GVKs that are first passed then read.
On the other hand if the user passes an instance of GVKP just can cast back.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesDependentResource.java
@@ -9,7 +9,7 @@ import io.javaoperatorsdk.operator.processing.GroupVersionKind;
 public class GenericKubernetesDependentResource<P extends HasMetadata>
     extends KubernetesDependentResource<GenericKubernetesResource, P> {
 
-  private final GroupVersionKindPlural groupVersionKind;
+  private final GroupVersionKind groupVersionKind;
 
   public GenericKubernetesDependentResource(GroupVersionKind groupVersionKind) {
     this(GroupVersionKindPlural.from(groupVersionKind));
@@ -27,7 +27,7 @@ public class GenericKubernetesDependentResource<P extends HasMetadata>
   }
 
   @SuppressWarnings("unused")
-  public GroupVersionKindPlural getGroupVersionKind() {
+  public GroupVersionKind getGroupVersionKind() {
     return groupVersionKind;
   }
 }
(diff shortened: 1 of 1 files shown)


xstefank (commented):
Commenting, since I think that I miss some rules. Yes, user can fix this be downcating but this will be a breaking change. So it this is OK to push non-major release?

csviri:
> Commenting, since I think that I miss some rules. Yes, user can fix this be downcating but this will be a breaking change. So it this is OK to push non-major release?

Usually we discuss such breaking changes based on some possible impact, the Generic Kubernetes Resource APIs is already something quite limited, usually used to just cover resources for which there is no generated Java classes - what is rare - since there is a generator, or you doing some dynamic resource handling. This was asked also from some users, but  probably the major use case is the Glue operator, where this is causing problem during the update. (Also I see usage some places in QOSDK)

metacosm:
I don't understand why this change is needed. As you can pass a GVKP to create a `GenericKubernetesDependentResource` is perfectly normal to use the wider type. Also, as pointed out, this would be a breaking change.

csviri:
> I don't understand why this change is needed. As you can pass a GVKP to create a `GenericKubernetesDependentResource` is perfectly normal to use the wider type. Also, as pointed out, this would be a breaking change.

I literally was debugging 3 hours because of this, since you pass GVK and it under the hood changes it to GVKP, where the equal fails from that point. So we either do this this way or override GVK / GVKP equals method so they are aware of each other:

https://github.com/java-operator-sdk/kubernetes-glue-operator/blob/f1562a12f4254d30e6e80bcd7950779985d6c3df/src/main/java/io/javaoperatorsdk/operator/glue/reconciler/glue/GlueReconciler.java#L376-L383

But I prefer this since it can be very confusing for other users also.

csviri:
@metacosm where is this used now at all? I mean GVKP?

csviri:
Hmm, so the problem is that if I pass GVK it transforms it seamlessly to GVKP, so when I get it back that is not comparable to the input.

So we have a history with this also that GVKP is not used in JOSDK at all, we added this because of RBAC feature of in QOSDK, but it is not used atm neither in QOSDK. Is there still the intention to use it? Shouldn't we just move GVKP to QOSDK in that case?  Or remove it in general, at least in scope of JOSDK this does not make sense to keep.

metacosm:
I think the appropriate fix should be #2731. It needs to stay in JOSDK because downstream projects do not control how these objects are created. Generally speaking, plurals are pretty badly handled in Kubernetes despite the fact that they are quite central…

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/2731

PR #2737, opened 2025-03-19
docs: new structure of docs

csviri:
Provides a new structure - some pages were improved.


csviri:
@metacosm @xstefank I came up with this, for now basically separated docs to more granular an structured topics. Did not change the actual chapters itself which I will do eventually, thus make some improvements on every page.

xstefank (approved):
I like this.

metacosm (approved):

PR #2738, opened 2025-03-22
feat(ci): build with Java 21 and 24

metacosm:


csviri:
Shouldn't we support also 17 ? In practice probably used a lot.

metacosm:
Yeah, we could keep testing on 17, just wanted to check on 24 as well

xstefank (approved):
Agreed that 17 should be kept, otherwise looks good.

csviri (approved):

PR #2745, opened 2025-03-31
docs: explanation for max maxReconciliationInterval

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java
```diff
@@ -44,7 +44,9 @@
 
   /**
    * Optional configuration of the maximal interval the SDK will wait for a reconciliation request
-   * to happen before one will be automatically triggered.
+   * to happen before one will be automatically triggered. The intention behind this feature is to
+   * have a failsafe, not to have another way of repeated reconciliation. For that use {@link
```
metacosm, line 48:
```suggestion
   * have a failsafe, not to artificially force repeated reconciliations. For that use {@link
```

Issue #2249, opened 2024-02-26
Dependent Resource Event Filtration with Previous version as Annotation


csviri:
When a KubernetesDependentResource makes an update, it adds an annotation with the actual version of the resource:

https://github.com/operator-framework/java-operator-sdk/blob/6cab68b53829befa4755742fe0155878f5964bb9/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L187-L192

This is then used to if an event comes it could be skipped (won't trigger the reconciliation) since it is a result of the controller. This halves the reconciliation execution with the same result.
The actual setup that uses `matchers` works well. No functional issue.

The `matchers` (SSA based matcher by default) are however an optimization, that other controller might or might do, also there are some special cases that are hard to handle. So in case the matcher simply returns false, and the reconciliation is triggered, while the controller of the dependent resource (not our controller) made changes, the update done by the dependent resource (with the changed resources version in the annotation) will trigger an update (therefore an update event in the informer). While this will be filtered out, it would not create an event (thus there would not be a change (also no `resourceVersion` change) if the annotation with resource version is not changed.

This causes a real problem just in the case that the controller of the resource is reacting on every change, a typical (and not found more yet) is the `Deployment`, where there is somehow an exceptional functionality if the annotations are changed the `generation` is changed in the metadata. What triggers a reconciliation of the `Deployment`, that results in a change of `observedGeneration`. Which again triggers the reconciliation on our side. (If the matchers are in place this not causes an infinite loop).

So there are things here: 
- in some cases, we might produce additional changes to the resource using the annotation
- if the `matchers` are turned off in some very rare cases this might cause a loop in the reconciliation (in combination with the resources controller, like `Deployment`)

Note that there is a feature flag there to not use this annotation.

Should we do anything regarding this? Well would be nicer to have a simple algorithm without the annotation. The previous algorithm was more complex but did not use the annotation. So we might revisit this and see if we can do better.

csviri:
We added blacklist will close this issue for now

Issue #2742, opened 2025-03-27
SSABasedGenericKubernetesResourceMatcher Incorrectly Detects Mismatch for initialDelaySeconds=0 in Deployment ReadinessProbe


kamilchociej:
#### What did you do?

1. Define a Deployment CRUDKubernetesDependentResource in a Java Operator SDK controller with desired resource containing spec.template.spec.containers[0].readinessProbe.initialDelaySeconds=0.
2. Trigger reconciliation.
3. Observe that the controller repeatedly considers the resource as out-of-sync, even though there is no actual change needed.

The SSABasedGenericKubernetesResourceMatcher in Java Operator SDK incorrectly considers a Deployment resource as not matching the actual Kubernetes resource when spec.template.spec.containers[*].readinessProbe.initialDelaySeconds is set to 0.

Kubernetes omits fields with default values (0 for initialDelaySeconds) when storing resources, meaning the actual resource map retrieved from Kubernetes does not contain this field. However, the desired resource map includes initialDelaySeconds: 0, leading to an unnecessary update being triggered.

#### What did you expect to see?

If initialDelaySeconds: 0 is absent in the actual resource due to Kubernetes' defaulting behavior, the matcher should recognize this as equivalent to explicitly setting it to 0 in the desired resource.

#### What did you see instead? Under which circumstances?

The matcher detects a difference because the desired resource contains initialDelaySeconds: 0, while the actual resource lacks this field. This results in an unnecessary update being applied.

#### Environment

**Kubernetes cluster type:**

vanilla

`$ Mention java-operator-sdk version from pom.xml file`

4.9.7

`$ java -version`

21

`$ kubectl version`

v1.30.4

#### Possible Solution

Modify SSABasedGenericKubernetesResourceMatcher to handle cases where Kubernetes omits fields with default values, treating their absence as equivalent to explicitly setting the default value.

#### Additional context

This issue may affect other defaulted fields in Kubernetes objects, not just initialDelaySeconds. A general approach to handling default values might be necessary.

csviri:
Thx @kamilchociej , I think the current workaround is to fill the defaults in you current resources.

csviri:
Just to make sure I understand, this does not result in an infinite loop in your side, just an additional reconciliation right?

kamilchociej:
Whole process looks like:
1. Desired Deployment contains containing spec.template.spec.containers[0].readinessProbe.initialDelaySeconds=0.
2. Controller applies desired state.
[...]

PR #2749, opened 2025-04-01
fix: use SSA matcher to filter events

shawkins:
closes: #2249

Seeing the recent #2742, I was wondering about revisiting #2249. One thought shown roughly here is that if we use the SSA matcher to compare the new / old. If it matches, then the change may not be meaningful as it didn't touch an owned field. Any normalization that kubernetes does wouldn't cause false positives here.

It's currently wired in dubiously  - it probably would need to be done somehow via SSA enabled KubernetesDependentResources adding a filter to their informer event sources - which could make this independent of using the previous annotation.

When won't this work:
- if you need to react to changes in things you don't own
- if you miss handling an update that modifies your managed fields, after that point comparing the new / old like this doesn't help.

The only ways to workaround that I can think of would be to declare what non-owned state you care about and to track if we are behind on reconciling a resource - however these additional complexities may not be worth it.


shawkins:
Given our other discussions, this is not worth persuing.

Issue #2751, opened 2025-04-03
Error handling if status patching fails


shawkins:
**Is your feature request related to a problem? Please describe.**
If a controller has custom error handling that attempts a status patch, a failure will cause the original exception and retry handling to be lost.

**Describe the solution you'd like**
Catch the patching error.

**Describe alternatives you've considered**

**Additional context**

csviri:
see dicussion here: https://github.com/operator-framework/java-operator-sdk/pull/2752#issuecomment-2777901773

PR #2752, opened 2025-04-03
fix: catch exceptions from updating the status

shawkins:
There are two possible handlings here:
- catch and log the error from the status patch, so that the original retry handling is preserved
- log the original exception, and let the error from the status patch take over as it does now

I opted for the former, but as open to the latter.

closes: #2751

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
@@ -212,9 +212,18 @@ class ReconciliationDispatcher<P extends HasMetadata> {
 
     P updatedResource = null;
     if (errorStatusUpdateControl.getResource().isPresent()) {
-      updatedResource =
-          customResourceFacade.patchStatus(
-              errorStatusUpdateControl.getResource().orElseThrow(), originalResource);
+      try {
+        updatedResource =
+            customResourceFacade.patchStatus(
+                errorStatusUpdateControl.getResource().orElseThrow(), originalResource);
+      } catch (Exception ex) {
+        log.error(
+            "updateErrorStatus failed for resource: {} with version: {} for error {}",
+            getUID(resource),
+            getVersion(resource),
+            e.getMessage(),
+            ex);
+      }
     }
     if (errorStatusUpdateControl.isNoRetry()) {
       PostExecutionControl<P> postExecutionControl;


csviri:
> log the original exception, and let the error from the status patch take over as it does now

This makes more sense to me, considering the following situation having a patch from error handler that fails and not to retry raises some concerns of consistency. This way might be at least eventual consistent in general. But not retrying lead to an unexpected state: reconciliation failed and even status is not updated.
What do you think?

csviri:
> > log the original exception, and let the error from the status patch take over as it does now
> 
> This makes more sense to me, considering the following situation having a patch from error handler that fails and not to retry raises some concerns of consistency. This way might be at least eventual consistent in general. But not retrying lead to an unexpected state: reconciliation failed and even status is not updated. 

What do you think?

shawkins:
> This makes more sense to me, considering the following situation having a patch from error handler that fails and not to retry raises some concerns of consistency. 

Why would you not retry? Wouldn't that mean the core reconciliation action has no retry or retires left?

An independent retry for patching the status doesn't make as much sense for me. Since there is already retry logic built in the client, the most likely causes for the patching to fail would be:
- a 409 - in which case I'd expect another event to come in
- the api server is down for longer than the client retry, which I don't think requires special handling just for patching the status

csviri:
> > This makes more sense to me, considering the following situation having a patch from error handler that fails and not to retry raises some concerns of consistency.
> 
> Why would you not retry? Wouldn't that mean the core reconciliation action has no retry or retires left?
> 

Hmm, not sure if I'm following / we understand each other. My view is to retry is to issue reconciliation also in this case. Since if the user expects that the status should be patched, and in this case it won't be.  This might be sub-optimal, but better than a silently just ignoring this issue (in general).

Note that you can always just override this logic, and handle the patching explicitly and catch the error.

> An independent retry for patching the status doesn't make as much sense for me. Since there is already retry logic built in the client, the most likely causes for the patching to fail would be:
a 409 - in which case I'd expect another event to come in
the api server is down for longer than the client retry, which I don't think requires special handling just for patching the status

Or any temporal network error on the node.

csviri:
So if there is explicit no retry we could say, we don't retry even if patch fails. But again, I guess some might want to patch status for a reason, but you might be right so we might be good enough with logging in this case.

csviri (approved):

xstefank (approved):
It makes sense to me like this. If there are requests to patch the status, we could put this under config.

metacosm (approved):

metacosm:
Ideally, it would make sense to always try to patch the status but it might not be possible, especially if it already failed at least once…

csviri:
Thank you @shawkins !

PR #2757, opened 2025-04-07
Improve customizability of SSABasedGenericKubernetesResourceMacher

murillio4:
This PR addresses #2509 and makes it easier to customize the matching logic in SSABasedGenericKubernetesResourceMatcher without having to clone the entire class just to override its behavior.

The change is based on the solution suggested by @Donnerbart [here](https://github.com/operator-framework/java-operator-sdk/issues/2553#issuecomment-2493712071), but it avoids passing the primary resource directly in order to keep the existing matches method signature unchanged. If the primary resource is needed, it can instead be made available through the context.

The implementation introduces a new protected matches method that handles the actual comparison. The public matches method delegates to this new method, making it easy to override in subclasses.


xstefank (approved):

metacosm (approved):
Thank you!

metacosm:
We should probably make it easier to use an SSA matcher without having to override the `KubernetesDependentResource.match` method.

csviri (approved):

PR #2760, opened 2025-04-09
feat: allow easier configuration of matcher

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
@@ -37,9 +37,17 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
     implements ConfiguredDependentResource<KubernetesDependentResourceConfig<R>> {
 
   private static final Logger log = LoggerFactory.getLogger(KubernetesDependentResource.class);
+
+  @SuppressWarnings("rawtypes")
+  private static final SSABasedGenericKubernetesResourceMatcher defaultMatcher =
+      SSABasedGenericKubernetesResourceMatcher.getInstance();
+
   private final boolean garbageCollected = this instanceof GarbageCollected;
   private KubernetesDependentResourceConfig<R> kubernetesDependentResourceConfig;
   private volatile Boolean useSSA;
+
+  @SuppressWarnings("unchecked")
+  private SSABasedGenericKubernetesResourceMatcher<R> matcher = defaultMatcher;
 
   public KubernetesDependentResource(Class<R> resourceType) {
     this(resourceType, null);
@@ -52,6 +60,13 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
   @Override
   public void configureWith(KubernetesDependentResourceConfig<R> config) {
     this.kubernetesDependentResourceConfig = config;
+  }
+
+  public void setMatcher(SSABasedGenericKubernetesResourceMatcher<R> matcher) {
+    if (this.matcher != defaultMatcher) {
+      throw new IllegalStateException("Can only set a matcher once.");
+    }
+    this.matcher = matcher;
   }
 
   @SuppressWarnings("unused")
@@ -111,9 +126,7 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
     final boolean matches;
     addMetadata(true, actualResource, desired, primary, context);
     if (useSSA(context)) {
-      matches =
-          SSABasedGenericKubernetesResourceMatcher.getInstance()
-              .matches(actualResource, desired, context);
+      matches = matcher.matches(actualResource, desired, context);
     } else {
       matches =
           GenericKubernetesResourceMatcher.match(desired, actualResource, false, false, context)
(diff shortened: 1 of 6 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -37,10 +37,18 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
     implements ConfiguredDependentResource<KubernetesDependentResourceConfig<R>> {
 
   private static final Logger log = LoggerFactory.getLogger(KubernetesDependentResource.class);
+
+  @SuppressWarnings("rawtypes")
+  private static final SSABasedGenericKubernetesResourceMatcher defaultMatcher =
```
xstefank, line 42: maybe this could help with the warning:
```suggestion
  private static final SSABasedGenericKubernetesResourceMatcher<?> defaultMatcher =
```

xstefank: so it might be also `R` since it's set only on line 50.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -54,6 +62,13 @@ public void configureWith(KubernetesDependentResourceConfig<R> config) {
     this.kubernetesDependentResourceConfig = config;
   }
 
+  public void setMatcher(SSABasedGenericKubernetesResourceMatcher<R> matcher) {
+    if (this.matcher != defaultMatcher) {
```
xstefank, line 66: I'm not sure how big the matcher object can be, but this could be a boolean.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
metacosm: This won't work on static fields.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
metacosm: Not sure I'm following

xstefank: I mean you don't need to keep the `defaultMatcher`in the field if it's only used to verify that the `matcher` is set only once. It can be a boolean that is flipped when first set so it's only taking a byte (usually) instead of the size of the `defaultMatcher` object.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -49,11 +54,26 @@ public KubernetesDependentResource(Class<R> resourceType, String name) {
...
+    matcherSet = true;
+  }
+
   @Override
   public void configureWith(KubernetesDependentResourceConfig<R> config) {
     this.kubernetesDependentResourceConfig = config;
   }
 
+  public void setMatcher(SSABasedGenericKubernetesResourceMatcher<R> matcher) {
+    if (matcherSet) {
```
csviri, line 70: Why is this explicitly forbidden?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -38,16 +38,13 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
...
-  private static final SSABasedGenericKubernetesResourceMatcher defaultMatcher =
-      SSABasedGenericKubernetesResourceMatcher.getInstance();
-
   private final boolean garbageCollected = this instanceof GarbageCollected;
   private KubernetesDependentResourceConfig<R> kubernetesDependentResourceConfig;
   private volatile Boolean useSSA;
 
-  @SuppressWarnings("unchecked")
-  private SSABasedGenericKubernetesResourceMatcher<R> matcher = defaultMatcher;
+  private SSABasedGenericKubernetesResourceMatcher<R> matcher =
```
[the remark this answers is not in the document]
metacosm: Wanted to keep it simple first but it definitely could go on the config.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
metacosm: Because it doesn't make sense to change the matching strategy after the resource has been used already. If anything, it should probably be set when the dependent is created (which is an argument to put it on the config only).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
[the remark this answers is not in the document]
csviri: @metacosm do you to do this change?

csviri:
Might be nice to see this on some integration test also. (But since it is only a configuration value it is not necessary)

xstefank (approved):
LGTM

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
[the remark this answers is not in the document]
metacosm: Yes, I made it so it is now configured instead of settable manually.

csviri (approved):

PR #2761, opened 2025-04-11
feat: primary resource caching for followup reconciliation(s)

csviri:
It provides facilities to cache the primary resource for follow-up reconciliations, thus ensuring that the reconciler always handles the up-to-date primary.

Later, we might extend this to the resource part, not just status (in terms of utils, cache already supports that) - now the utility methods focus on status.


operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache/StatusPatchPrimaryCacheReconciler.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  public UpdateControl<StatusPatchPrimaryCacheCustomResource> reconcile(
+      StatusPatchPrimaryCacheCustomResource primary,
+      Context<StatusPatchPrimaryCacheCustomResource> context)
+      throws InterruptedException {
+
+    primary = cache.getFreshResource(primary);
+
+    if (primary.getStatus() != null && primary.getStatus().getValue() != latestValue) {
+      errorPresent = false;
```
xstefank, line 41: it feels like this should be true

csviri: fixed thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  public static <P extends HasMetadata> P patchAndCacheStatusWithLock(
+      P primary, Context<P> context, BiFunction<P, KubernetesClient, P> patch) {
+    checkResourceVersionPresent(primary);
+    var updatedResource = patch.apply(primary, context.getClient());
+    context
+        .eventSourceRetriever()
+        .getControllerEventSource()
+        .handleRecentResourceUpdate(ResourceID.fromResource(primary), updatedResource, primary);
+    return null;
```
afalhambra-hivemq, line 85:
```suggestion
    return updatedResource;
```

xstefank (approved):
LGTM

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache/StatusPatchPrimaryCacheIT.java
```diff
@@ -0,0 +1,48 @@
...
+
+    // the reconciliation id periodically triggered, the status values should be increasing
+    // monotonically
+    await()
+        .pollDelay(Duration.ofSeconds(1))
+        .pollInterval(Duration.ofMillis(30))
+        .untilAsserted(
+            () -> {
+              assertThat(reconciler.errorPresent).isFalse();
+              assertThat(reconciler.latestValue).isGreaterThan(10);
```
afalhambra-hivemq, line 38: Maybe to make sure and guarantee no status update is lost, we could check an array or a list of values from 1 to 10 are present (all of them in particular order). 
Otherwise, this assert `isGreaterThan(10)` will eventually be true regardless of the caching.

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache/StatusPatchPrimaryCacheIT.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
+import io.javaoperatorsdk.operator.junit.LocallyRunOperatorExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+public class StatusPatchPrimaryCacheIT {
```
afalhambra-hivemq, line 14:
```suggestion
class StatusPatchPrimaryCacheIT {
```

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/withlock/StatusPatchCacheWithLockIT.java
```diff
@@ -0,0 +1,48 @@
...
+
+    // the reconciliation id periodically triggered, the status values should be increasing
+    // monotonically
+    await()
+        .pollDelay(Duration.ofSeconds(1))
+        .pollInterval(Duration.ofMillis(30))
+        .untilAsserted(
+            () -> {
+              assertThat(reconciler.errorPresent).isFalse();
+              assertThat(reconciler.latestValue).isGreaterThan(10);
```
afalhambra-hivemq, line 38: Maybe to make sure and guarantee no status update is lost, we could check an array or a list of values from 1 to 10 are present (all of them in particular order).
Otherwise, this assert `isGreaterThan(10)` will eventually be true regardless of the caching.

afalhambra-hivemq (commented):
When using the `PatchAndCacheStatusWithLock` approach, we will need to explicitly use optimistic locking here, which in theory it should be fine. 
However, the default retry mechanism provided by the JOSDK when a `409` conflict happen doesn't pick the latest resource status, forcing us to implement a custom retry mechanism for this. Otherwise the default retry mechanism is not useful here.

Anyways, I can raise a separate issue for that default mechanism to see if it can be improved there.

Beside that, great work and thanks a lot for this PR.

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache/StatusPatchPrimaryCacheIT.java
csviri: Hi there is that is error check, what should induce that there was monotonic, but yeah definitelly can improve the azsserts

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/withlock/StatusPatchCacheWithLockIT.java
csviri: See comment above, let's discuss there.

csviri:
> When using the `PatchAndCacheStatusWithLock` approach, we will need to explicitly use optimistic locking here, which in theory it should be fine. However, the default retry mechanism provided by the JOSDK when a `409` conflict happen doesn't pick the latest resource status, forcing us to implement a custom retry mechanism for this. Otherwise the default retry mechanism is not useful here.
> 
> Anyways, I can raise a separate issue for that default mechanism to see if it can be improved there.
> 
> Beside that, great work and thanks a lot for this PR.

Probably the best would be to not use the approach with the lock, but the lockess alternative.

afalhambra-hivemq:
> Probably the best would be to not use the approach with the lock, but the lockless alternative.

The only drawback here with the lockless alternative is that it will require us to handle the cache in the operator itself

csviri:
> > Probably the best would be to not use the approach with the lock, but the lockless alternative.
> 
> The only drawback here with the lockless alternative is that it will require us to handle the cache in the operator itself

Yes, but that is as simple as in the integration test.  We might interate on this in the future, but in the short term I think that is best for you.

afalhambra-hivemq:
> resource(resource).replace(), that would basically retry locally the update 

Curious here, what do you mean by `retry locally`?


> Some kind of retry would be nice also for patch withouth the lock. So will spend some more time on this to improve. 

That would be awesome! Thanks

csviri:
@afalhambra-hivemq @metacosm @xstefank added docs and made some improvements.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
 From v5, by default, the finalizer is added using Server Side Apply. See also `UpdateControl` in docs.
+
+### Making sure primary is up to date for the next reconciliation
+
+When you implement a reconciler as a final step (but maybe also multiple times during one reconciliation), you 
+usually update the status subresource with the information that was available during the reconciliation.
+Sometimes this is referred to as the last observed state. 
+When the resource is updated, the framework does not cache the resource directly from the response of the update.
+Instead, the underlying informer eventually receives an event with the updated resource and caches the resource.
+Therefore, it can happen that on next reconciliation the primary resource is not up-to-date regarding your updated (note that other event sources
```
afalhambra-hivemq, line 180:
```suggestion
Therefore, it can happen that on next reconciliation the primary resource is not up-to-date regarding your updated status subresource (note that other event sources
```

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+can trigger the reconciliation meanwhile). This is not usually a problem, since the status is not used as an input,
+the reconciliation runs again, and the status us updated again. The caches are eventually consistent.
```
afalhambra-hivemq, line 182:
```suggestion
the reconciliation runs again, and the status is updated again. The caches are eventually consistent.
```

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+
+### Making sure primary is up to date for the next reconciliation
+
+When you implement a reconciler as a final step (but maybe also multiple times during one reconciliation), you 
+usually update the status subresource with the information that was available during the reconciliation.
+Sometimes this is referred to as the last observed state. 
+When the resource is updated, the framework does not cache the resource directly from the response of the update.
+Instead, the underlying informer eventually receives an event with the updated resource and caches the resource.
+Therefore, it can happen that on next reconciliation the primary resource is not up-to-date regarding your updated (note that other event sources
+can trigger the reconciliation meanwhile). This is not usually a problem, since the status is not used as an input,
```
afalhambra-hivemq, line 181: > since the status is not used as an input,

I guess this depends on the reconciler implementation. In our case for instance, we check the status of the resource and react based up on it.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+
+However, there are cases when you would like to store some state in the status, typically generated 
+IDs of external resources. 
+See related topic in Kubernetes docs: [Representing Allocated Values](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#representing-allocated-values).
+In this case, it is reasonable to expect to have the state always available for the next reconciliation,
+to avoid generating the resource again and other race conditions.
+
+Therefore,
+the framework provides facilities
+to cover these use cases withing [`PrimaryUpdateAndCacheUtils`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java#L16).
```
afalhambra-hivemq, line 192:
```suggestion
to cover these use cases with [`PrimaryUpdateAndCacheUtils`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java#L16).
```

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+    
+    return UpdateControl.noUpdate();
+  }
+```
+
+In the background `PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatusWith` puts the result of the update into an internal
+cache of the event source of primary resource, and will make sure that the next reconciliation will contain the most 
+recent version of the resource. Note that it is not necessarily the version from the update response, it can be newer
+since other parties can do additional updates meanwhile, but if not explicitly modified, it will contain the up-to-date
+status.
```
afalhambra-hivemq, line 221: > Note that it is not necessarily the version from the update response, it can be newer
since other parties can do additional updates meanwhile, but if not explicitly modified, it will contain the up-to-date
status.

Not so clear, what you mean here. Updated resource version?

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+  
+```
+
+[`PrimaryResourceCache`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/support/PrimaryResourceCache.java)
+is designed for this purpose. 
+As shown in the example above, it is up to you to provide a predicate to determine if the resource is more recent than the one available.
+In other words, when to evict the resource from the cache. Typically, as show in the [integration test](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache)
+you can have a counter in status to check on that. 
+
+Since all of this happens explicitly, you cannot use it for now with managed dependent resources and workflows.
```
afalhambra-hivemq, line 280: > Since all of this happens explicitly, you cannot use it for now with managed dependent resources and workflows.

Can you please elaborate? This `PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatus(primary, freshCopy, context, cache);` cannot be used under what conditions?

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+    
+    freshCopy.getStatus().setValue(statusWithState());
+    
+    var updatedResource = PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatus(resource, freshCopy, context);
+    
+    return UpdateControl.noUpdate();
+  }
+```
+
+In the background `PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatusWith` puts the result of the update into an internal
```
afalhambra-hivemq, line 217:
```suggestion
In the background `PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatus` puts the result of the update into an internal
```

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,121 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+As shown in the example above, it is up to you to provide a predicate to determine if the resource is more recent than the one available.
+In other words, when to evict the resource from the cache. Typically, as show in the [integration test](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache)
+you can have a counter in status to check on that. 
+
+Since all of this happens explicitly, you cannot use it for now with managed dependent resources and workflows.
+
+#### Additional remarks
+
+As shown in the integration tests, there is no optimistic locking used when updating the
+[resource](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/internal/StatusPatchCacheReconciler.java#L41)
```
afalhambra-hivemq, line 285: In this particular IT, optimistic locking is in place, or am completely I wrong here?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -0,0 +1,222 @@
...
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+public class PrimaryUpdateAndCacheUtils {
+
+  private PrimaryUpdateAndCacheUtils() {}
+
+  private static final Logger log = LoggerFactory.getLogger(PrimaryUpdateAndCacheUtils.class);
+
+  /**
+   * Makes sure that the up-to-date primary resource will be present during the next reconciliation.
```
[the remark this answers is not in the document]
csviri: Will clean the javadoc, neither requires optimistic locking anymore.

docs/content/en/docs/documentation/reconciler.md
csviri: There is no optimistic locing there, the fresh copy does not contain resource version

afalhambra-hivemq: in that test, in the fresh copy, the resource version is actually set as below there:

```java
    var freshCopy = createFreshCopy(resource);
    // setting the resource version
    freshCopy.getMetadata().setResourceVersion(resource.getMetadata().getResourceVersion());
```

docs/content/en/docs/documentation/reconciler.md
csviri: Improved workding, the point is that you receive a new version from the update where the status is stored, but in next reconciliation you might get even newer resource if someone update the resource meanwhile.

docs/content/en/docs/documentation/reconciler.md
csviri: Ahh sorry missed that, will remove that.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,124 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+
+[`PrimaryResourceCache`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/support/PrimaryResourceCache.java)
+is designed for this purpose. 
+As shown in the example above, it is up to you to provide a predicate to determine if the resource is more recent than the one available.
+In other words, when to evict the resource from the cache. Typically, as show in the [integration test](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/primarycache)
+you can have a counter in status to check on that. 
+
+Since all of this happens explicitly, you cannot use this approach for managed dependent resources and workflows; 
+Since passing of the primary resource to the dependent resource always comes directly from the underlying informer event
+source cache.
```
afalhambra-hivemq, line 283: Thanks for adding this @csviri. This is pretty important. ❤️

csviri: We migh iterate on this to support it also in dependent resources, but I think it should be quite rare that some will need this explicit cache.

afalhambra-hivemq: In our case, in the DRs we check the status from the primary and react based upon it, and sometimes we update the primary status there as well. 
But yeah, maybe we can improve this in future iterations.

csviri: Note that the defatul approach works with DR. (only this alternative is not) since there is explicit call to get the resource:
```java
primary = cache.getFreshResource(primary);
```

csviri: Also you can still used non managed DR + WFs

csviri:
cc @gyfora

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -13,57 +13,61 @@
...
+ *       {@link #handleRecentResourceUpdate(ResourceID, HasMetadata, HasMetadata)} and {@link #handleRecentResourceCreate(ResourceID, HasMetadata)} need to be called explicitly
+ *       after a resource is created or updated using the kubernetes client. These calls are done
+ *       automatically by the KubernetesDependentResource implementation. In the background this will
+ *       store the new resource in a temporary cache {@link TemporaryResourceCache} which does
+ *       additional checks. After a new event is received the cached object is removed from this
+ *       cache, since it is then usually already in the informer cache.
+ *   <li>Avoiding unneeded reconciliations after resources are created or updated. This filters out events that are the results of updates and creates made
+ *       by the controller itself because we typically don't want the associated informer to trigger an event causing a useless reconciliation (as the change originates from the reconciler itself). This is achieved
+ *       by
+ *       TODO: update as this mentions methods that don't exist anymore and isn't very clear
```
metacosm, line 48: This section needs to be updated as per the TODO.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,117 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+their associated primary resource from the underlying informer event source cache.
+
+#### Additional remarks
+
+As shown in the integration tests, there is no optimistic locking used when updating the
+[resource](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/statuscache/internal/StatusPatchCacheReconciler.java#L41)
+(in other words `metadata.resourceVersion` is set to `null`). This is desired since you don't want the patch to fail on
+update.
+
+In addition, you can configure the Fabric8 client retry.
```
metacosm, line 285: This probably needs to be explained more and, possibly, point to the appropriate documentation.

metacosm (changes requested):

docs/content/en/docs/documentation/reconciler.md
csviri: added link

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -0,0 +1,225 @@
...
+
+    context
+        .eventSourceRetriever()
+        .getControllerEventSource()
+        .handleRecentResourceUpdate(ResourceID.fromResource(primary), res, primary);
+    return res;
+  }
+
+  /**
+   * Patches the resource and adds it to the {@link PrimaryResourceCache} provided.
```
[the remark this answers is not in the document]
csviri: removed provided from the end of sentences.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -169,3 +169,123 @@ You can specify the name of the finalizer to use for your `Reconciler` using the
...
+
+However, there are cases when you would like to store some state in the status, typically generated 
+IDs of external resources. 
+See related topic in Kubernetes docs: [Representing Allocated Values](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#representing-allocated-values).
+In this case, it is reasonable to expect to have the state always available for the next reconciliation,
+to avoid generating the resource again and other race conditions.
+
+Therefore,
+the framework provides facilities
+to cover these use cases with [`PrimaryUpdateAndCacheUtils`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java#L16).
```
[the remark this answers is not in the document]
csviri: was removed, yeah this is always a challenge but luckily in this case we don't need the line number :)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: updated, just directing to the related method, not sure if it makes sense to describe more in depth since it would be not simpler than the code it self.

metacosm (approved):

PR #2762, opened 2025-04-11
[WIP] feat: support delete event reconiliation

csviri:
The idea is that reconcile method will be called on resource delete event.
What would allow to clear inMemory caches without using a finalizer (thus Cleaner interface)


csviri:
This we might want to schedule for 5.1

PR #2763, opened 2025-04-11
fix: event marking bug

csviri:


csviri:
Will try to add an unit test for this, but it might be the case that this did caused a real issue or just in extremely rare circumstances

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -196,7 +196,7 @@ private void handleEventMarking(Event event, ResourceState state) {
         // event as below.
         markEventReceived(state);
       }
-    } else if (!state.deleteEventPresent() || !state.processedMarkForDeletionPresent()) {
+    } else if (!state.deleteEventPresent() && !state.processedMarkForDeletionPresent()) {
```
metacosm, line 199: Didn't we fix this already? I guess not but it rings a bell.

metacosm (approved):

PR #2766, opened 2025-04-11
rough in of tracking things that are defaults and normalizations

shawkins:
closes: #2249

This is an alternative to fleshing out the approach in #2749

Here the desired state and the state returned from kubernetes is used to determine what values have been normalized and defaulted. It's stored in a cache on the SSA matcher - that cache will need to be size / time bound, but that has not been added yet.

We can use that information in a couple of ways.

The first roughed in here is to try and augment our matches decision making - if it looks like the same desired state was requested, and the actual contains all the overriden values, then we can assume that things still match. To cut down on the memory footprint this is based upon a hash of the desired state, rather than the full desired state.

Another, simpler and potentially safer approach, would be to just use this cache to determine if the previous annotation should be used.

Finally rather than caching or trying to use this in some novel way, we could at least log that what is happening and make it clear what will happen if use of the previous annotation is enabled.

WDYT @xstefank @csviri @metacosm

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
@@ -4,6 +4,7 @@ import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -27,6 +28,7 @@ import io.fabric8.kubernetes.client.utils.KubernetesSerialization;
 import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
 import io.javaoperatorsdk.operator.processing.LoggingUtils;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
 
 import com.github.difflib.DiffUtils;
 import com.github.difflib.UnifiedDiffUtils;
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java
```diff
@@ -79,16 +87,134 @@ public static <L extends HasMetadata> SSABasedGenericKubernetesResourceMatcher<L
...
   public boolean matches(R actual, R desired, Context<?> context) {
+    SSAState state = getSSAState(actual, desired, context);
+    
+    if (state == null) {
+      return false;
+    }
+    
+    SSAState overrides = defaultsAndNormalizations.get(new CacheKey(ResourceID.fromResource(actual), desired.hashCode()));
+    if (overrides != null) {
+      // if containsAll(overrides.desired, state.desired) && containsAll(overrides.actual, state.actual)
```
csviri, line 186: I assume here goes `return true`?

csviri: If it is true might not always mean match or?

csviri:
Not sure if I follow all the details, especially the missing ones. 
But generally the idea that record the transformation based on response  makes sense to me. 

Wonder if this would not require optimistic locking to be absolutely correct, is when we apply a desired state that is based  on "actual" from the cache, but let say this cache is stale, than we calculate the state based on outdated values (since it is based on response from the server) , or?

csviri:
@shawkins maybe we could also take a look together on our community meeting.

shawkins:
> But generally the idea that record the transformation based on response is great.
> 
> Wonder if this would not require optimistic locking to be absolutely correct, is when we apply a desired state that is based on "actual" from the cache, but let say this cache is stale, than we calculate the state based on outdated values (since it is based on response from the server) , or?

You're talking specifically here about the changes in SSA matching were we try to determine if we can skip the event correct?

The other two potential usages, wouldn't have any locking concerns:

- just log at a high level when we detect that looping could occur - this could even just reuse the debug logging that's already in the SSA matcher.

- avoid using the previous annotation.

For the SSA matching logic I can't quite see a locking concern either - maybe you assumed that we were skipping the rest of the matching logic because I didn't flesh things out fully.

> (so theoretically another manager/controller could have added a field that is now considered as default, I think defaults do not show up in managed fields, maybe also we should check if the defaults are not in managed fields or other managers? - not sure just thinking out loud)

I can confirm inconsistent behavior wrt defaults - at least for my latest keycloak issue along these lines, the default for a feildRef apiVersion ends up looking like it is managed by the operator even when it is not specified.

If the field is owned by another manager, then it will get pruned as expected. 

As you are getting if there were some case where a non-managed default could be added later (not by the operator create / update), then yes that would not be understood by this additional logic - it would have to be covered instead by the SSA matche's built-in logic.

PR #2774, opened 2025-04-28
improve: blocklist of problematic resources for previous version annotation

csviri:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -474,8 +474,7 @@ public interface ConfigurationService {
    *
    * @return a Set of resource classes where the previous version annotation won't be used.
    */
-  default Set<Class<? extends HasMetadata>>
-      previousAnnotationForDependentResourcesEventFilteringBlocklist() {
+  default Set<Class<? extends HasMetadata>> withPreviousAnnotationForDependentResourcesBlocklist() {
     return Set.of(Deployment.class, StatefulSet.class);
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -1,7 +1,6 @@
 package io.javaoperatorsdk.operator.api.config;
 
 import java.time.Duration;
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
@@ -464,10 +463,14 @@ public interface ConfigurationService {
    * resources. Note that if you are sure that the matcher works (most of the cases does) for your
    * case, you can remove the resource from the blocklist.
    *
+   * <p>The consequence of adding a resource type to this list is that it will not use event
+   * filtering in dependent resources, so will process also events which are results from updates of
+   * the dependent.
+   *
    * @return blocklist of resource classes where the previous version annotation won't be used.
    */
-  default List<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
-    return List.of(Deployment.class, StatefulSet.class);
+  default Set<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
+    return Set.of(Deployment.class, StatefulSet.class);
   }
 
   /**
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -14,7 +14,6 @@ import org.slf4j.LoggerFactory;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Secret;
-import io.fabric8.kubernetes.api.model.apps.DaemonSet;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.api.model.apps.StatefulSet;
 import io.fabric8.kubernetes.client.Config;
@@ -468,7 +467,7 @@ public interface ConfigurationService {
    * @return blocklist of resource classes where the previous version annotation won't be used.
    */
   default List<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
-    return List.of(Deployment.class, StatefulSet.class, DaemonSet.class);
+    return List.of(Deployment.class, StatefulSet.class);
   }
 
   /**
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -474,7 +474,8 @@ public interface ConfigurationService {
    *
    * @return a Set of resource classes where the previous version annotation won't be used.
    */
-  default Set<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
+  default Set<Class<? extends HasMetadata>>
+      previousAnnotationForDependentResourcesEventFilteringBlocklist() {
     return Set.of(Deployment.class, StatefulSet.class);
   }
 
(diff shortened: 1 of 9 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -448,6 +452,10 @@ default boolean previousAnnotationForDependentResourcesEventFiltering() {
     return true;
   }
 
+  default List<Class<? extends HasMetadata>> previousAnnotationUsageBlacklist() {
+    return List.of(Deployment.class, StatefulSet.class, ReplicaSet.class);
```
csviri, line 456: @shawkins do I remember correctly these are the ones you had issues with? thx!

shawkins: Deployment and StatefulSet I believe are the ones we encounted problems with before. DaemonSet may need to be included as well.

csviri: thank you!

csviri: just for the record a I did not add DaemonSet since was not able to reproduce the behavior regarding generation increase with it.

csviri:
related issues:
https://github.com/operator-framework/java-operator-sdk/issues/2509
https://github.com/operator-framework/java-operator-sdk/issues/2553

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -448,6 +451,25 @@ default boolean previousAnnotationForDependentResourcesEventFiltering() {
...
+   * correctly with the resource matcher yet again results in an update and a previous version
+   * annotation change, thus results in an infinite loop.
+   *
+   * <p>As a workaround, we automatically skip adding previous annotation for those well-known
+   * resources. Note that if you are sure that the matcher works (most of the cases does) for your
+   * case, you can remove the resource from the blocklist.
+   *
+   * @return blocklist of resource classes where the previous version annotation won't be used.
+   */
+  default List<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
```
xstefank, line 469: I would suggest also documenting the default implementation in the javadoc.

csviri: Not sure that is needed / that helpful since users can just open the code and see the defaults.

metacosm: What would make more sense, though, is to explain what is the consequence for a resource type to be in the block list because this isn't clear at all and people might be hesitant adding a resource to the block list without more details of what happens when they do.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -448,6 +451,25 @@ default boolean previousAnnotationForDependentResourcesEventFiltering() {
...
+    return List.of(Deployment.class, StatefulSet.class);
```
metacosm, line 470: Another aspect is whether or not these 2 classes should *always* be added to the block list regardless of what the user might set this list to (i.e. what is returned by this method should be added to the default list, not replace it completely).

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
metacosm: Also, this should probably be a set rather than a list as the order is meaningless whereas we only want one instance of a given resource type in this collection.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
```diff
@@ -188,6 +190,12 @@ public ConfigurationServiceOverrider withCloneSecondaryResourcesWhenGettingFromC
     return this;
   }
 
+  public ConfigurationServiceOverrider previousAnnotationUsageBlocklist(
+      List<Class<? extends HasMetadata>> previousAnnotationUsageBlacklist) {
+    this.previousAnnotationUsageBlocklist = previousAnnotationUsageBlacklist;
```
metacosm, line 195: See the previous comment on how this should probably be adding to the default set rather than completely replace it.

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/prevblocklist/DeploymentDependent.java
```diff
@@ -0,0 +1,95 @@
...
+                                                .withLimits(Map.of("cpu", new Quantity("2000m")))
+                                                .build())
+                                        .build())
+                                .build())
+                        .build())
+                .build())
+        .build();
+  }
+
+  // for testing purposes replicating the matching logic but with the special matcher
```
metacosm, line 69: This should use the configuration option from #2760 instead.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
csviri: Those should not be always added, if the user is sure that the matcher is right those can be removed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
csviri: Responded above.

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/prevblocklist/DeploymentDependent.java
csviri: If that is merged, can change it (might be a new PR). thx!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
csviri: Explained the consequence and changed it to set, thank you!

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/prevblocklist/PrevAnnotationBlockReconcilerIT.java
```diff
@@ -0,0 +1,50 @@
...
+class PrevAnnotationBlockReconcilerIT {
+
+  public static final String TEST_1 = "test1";
+
+  @RegisterExtension
+  LocallyRunOperatorExtension extension =
+      LocallyRunOperatorExtension.builder()
+          // Removing resource from blocklist List would result in test failure
+          //          .withConfigurationService(
+          //              o -> o.previousAnnotationUsageBlocklist(Collections.emptyList()))
```
xstefank, line 24: Should this be a new test then?

xstefank (approved):

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/prevblocklist/PrevAnnotationBlockReconcilerIT.java
csviri: Yes, it could be an additional test, I did not add that since I did not find it important to show that something is not working.

csviri:
@metacosm are there any other concerns or we can merge this?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
metacosm: Not sure I understand. From what I understand, Deployment and StatefulSet are always problematic for the matcher, no?

csviri: No, only if it uses fields that are normalized later by the API, (and maybe if some defaults filled in, not sure about that)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -474,7 +474,8 @@ default boolean previousAnnotationForDependentResourcesEventFiltering() {
    *
    * @return a Set of resource classes where the previous version annotation won't be used.
    */
-  default Set<Class<? extends HasMetadata>> previousAnnotationUsageBlocklist() {
+  default Set<Class<? extends HasMetadata>>
+      previousAnnotationForDependentResourcesEventFilteringBlocklist() {
```
csviri, line 478: I think this is fine to have also with shorter form and keep it consistent. (withPreviousAnnotationForDependentResourcesBlocklist)

metacosm (approved):

PR #2779, opened 2025-04-30
feat: allow reconcilers to override their own configuration

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
@@ -112,10 +112,18 @@ public class AbstractConfigurationService implements ConfigurationService {
   public <R extends HasMetadata> ControllerConfiguration<R> getConfigurationFor(
       Reconciler<R> reconciler) {
     final var key = keyFor(reconciler);
-    final var configuration = configurations.get(key);
+    var configuration = configurations.get(key);
     if (configuration == null) {
       logMissingReconcilerWarning(key, getReconcilersNameMessage());
+    } else {
+      if (reconciler instanceof ConfigurableReconciler<?> configurableReconciler) {
+        final var overrider = ControllerConfigurationOverrider.override(configuration);
+        configurableReconciler.updateConfigurationFrom(overrider);
+        configuration = overrider.build();
+        configurations.put(key, configuration);
+      }
     }
+
     return configuration;
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
@@ -116,6 +116,7 @@ public class AbstractConfigurationService implements ConfigurationService {
     if (configuration == null) {
       logMissingReconcilerWarning(key, getReconcilersNameMessage());
     } else {
+      // if a reconciler is also a ConfigurableReconciler, update and replace its configuration
       if (reconciler instanceof ConfigurableReconciler<?> configurableReconciler) {
         final var overrider = ControllerConfigurationOverrider.override(configuration);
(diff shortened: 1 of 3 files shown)


csviri (changes requested):
What is the use case for this change?

Would be good to document it at least as Javadoc on the interface

metacosm:
> What is the use case for this change?

This allows for `Reconciler` implementations to customize their configuration based on runtime information before being registered in situations where programmatic configuration via manually created `ControllerConfigurationOverrider` might not be easy or straightforward, like the Quarkus extension (and could potentially simplify the Spring Boot starter as well).

xstefank (approved):
The only thing I don't understand is why it's not applied the first time `getConfigurationFor` is called.

csviri:
> > What is the use case for this change?
> 
> This allows for `Reconciler` implementations to customize their configuration based on runtime information before being registered in situations where programmatic configuration via manually created `ControllerConfigurationOverrider` might not be easy or straightforward, like the Quarkus extension (and could potentially simplify the Spring Boot starter as well).

I see thank you, makes sense, just some remarks:

Just for the record: In general in v5 we aimed to remove interface implementation for reconciler since are not easy to discover. However since this is mainly intended for the extension (quarkus/SB) I think it is better to do this way as it is then having it as a default method on reconciler interface.

What however, might be considered as an alternative ( and let me know this if this is also feasible on Quarkus side), I'm quite sure this would work in SB, but can prepare a prototype; is this:

To not bake this mechanism into the core, rather have this interface in the extensions only, and provide it as a bean:

```java
@Singleton
@ForController(MyController.class)
@ForController(MyOtherController.class)
public class MyConfigOverrode implements ConfigurableReconciler {

@Inject 
Object anyBeanOrRuntimeValue

void updateConfigurationFrom(ControllerConfigurationOverrider configOverrider) {
 // my code to use override, possibly using runtime injected beans.
}

}
```

Then is bean instance could be easily used with the current API when the reconciler is registered in the background.

Note the `@ForController(MyController.class)` part. This could be used to identify the controller(s), thus possibly more controllers that share some common values, like label selectors.

Can expand on this later, hope it is clear how I mean it. Pls let me know what do you think.  The pros is that we don't have to introduce an other construct to the core, also might be a more flexible approach for the users (maybe with a more modern API? ). On the other hand the config is not tightly coupled with the reconciler (which might be a good and also a bad thing).

metacosm:
Putting the interface in the extension(s) is something that we discussed, actually. It can make sense but I thought that adding it to the core also made sense since it allows for the feature to be implemented once instead of having to duplicate the logic in slightly different ways in the extensions. It also provides a way for users to tie the configuration code to their reconciler instead of having it distributed in several spots (though, you could certainly organize your code to have it in a single spot, the current implementation doesn't really encourage this). 

That said, it is true that introducing yet another configuration mechanism in the core might confuse users, especially considering that there are some subtleties that would need to be documented so it might indeed make more sense to put this features in extensions that need them.

metacosm:
Started looking into moving that feature into extensions. It would require #2784 at the very least, and most likely a way for extensions to provide their custom configuration instances to the overrider instead of always using `ResolvedControllerConfiguration`.

metacosm:
> Started looking into moving that feature into extensions. It would require #2784 at the very least, and most likely a way for extensions to provide their custom configuration instances to the overrider instead of always using `ResolvedControllerConfiguration`.

Actually, that might not be required if we can continue living with the fact that it is possible to get 2 possibly different versions of a `Reconciler`'s configuration depending on whether it is retrieved from the `ConfigurationService` or `RuntimeInfo`.

metacosm:
See quarkiverse/quarkus-operator-sdk#1090 for how this might look like for the Quarkus extension. Note that I didn't bother with the `ForController` annotation because I tend to think it's quite overkill as a much easier implementation would be to use a shared implementation / bean instead of having the extension have to do the wiring, adding complexity for no obvious benefits, imo.

csviri:
Makes sense, I think it is much better this way at this point to have this in the extension, thank you!!

metacosm:
Closing this PR as this will instead be supported in extensions instead of JOSDK core.

PR #2784, opened 2025-05-07
fix: update configuration service when overriding configuration

metacosm:


metacosm:
Currently, the configuration information related to a reconciler can be inconsistent between what's returned by the `ConfigurationService` and what can be returned by `RuntimeInfo` from `RegisteredController` (which should be the correct version, in case they differ). The problem is `RuntimeInfo` is not very well known and not as widely used as `ConfigurationService`, which means that users can observe outdated configuration.

xstefank (approved):

csviri:
How can be that outdated, if it is immutable?

csviri:
Note that only change that now allow in runtime is changing watched namespaces through `RegisteredController`. After start no configuration other than this should be allowed.

PR #2785, opened 2025-05-08
improve: increase bounded cache IT GC timeout

csviri:
the test fails some times because of this


metacosm:
I don't really like that we need to keep increasing the time outs on the tests… to me, this seems to indicate that we maybe have some performance regression that we should look into. 😞

csviri:
This is has nothing to do with us ( fortunately ) this is a test which is quite separated from others and the value was low. But all we do is just call delete on primary and GC of K8S is taking care. But indeed on k8s side it seems it got slower but not sure.

metacosm (approved):

PR #2790, opened 2025-05-09
improve: remove fabric8 related daily build

csviri:
This never worked properly as far I can remember and creates lots of noise (daily fail email).
One of the problem is that the card generator maven plugin does not seems to be released in snapshot repo.


metacosm:
I'm fine removing it.

metacosm (approved):

PR #2799, opened 2025-05-14
improve: PrimaryUpdateAndCacheUtils locks, naming and doc improvements

csviri:
TODO
- cache utils test

### DIFF
--- a/docs/content/en/docs/documentation/reconciler.md
+++ b/docs/content/en/docs/documentation/reconciler.md
@@ -211,7 +211,7 @@ public UpdateControl<StatusPatchCacheCustomResource> reconcile(
 ```
 
 After the update `PrimaryUpdateAndCacheUtils.ssaPatchStatusAndCacheResourceWithLock` puts the result of the update into an internal
-cache and will the framework will make sure that the next reconciliation will contain the most recent version of the resource.
+cache and the framework will make sure that the next reconciliation contains the most recent version of the resource.
 Note that it is not necessarily the version of the resource you got as response from the update, it can be newer since other parties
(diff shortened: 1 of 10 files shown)


csviri:
@xstefank @metacosm this is the next iteration of the primary caches. 

It provides an additional and I think preferred way to update the internal caches using update method with retryable optimistic locking. I think this should be the default what users should use.

To be honest I'm a bit hesitant if we should provide the other approaches at this point. But would do this in a way that we merge this and in case I remove it in a subsequent PR.

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -201,27 +204,32 @@ public UpdateControl<StatusPatchCacheCustomResource> reconcile(
...
     
     return UpdateControl.noUpdate();
   }
 ```
 
-In the background `PrimaryUpdateAndCacheUtils.ssaPatchAndCacheStatus` puts the result of the update into an internal
-cache and will make sure that the next reconciliation will contain the most recent version of the resource. Note that it
-is not necessarily the version of the resource you got as response from the update, it can be newer since other parties
+After the update `PrimaryUpdateAndCacheUtils.ssaPatchStatusAndCacheResourceWithLock` puts the result of the update into an internal
+cache and will the framework will make sure that the next reconciliation will contain the most recent version of the resource.
```
metacosm, line 214:
```suggestion
cache and the framework will make sure that the next reconciliation contains the most recent version of the resource.
```

metacosm:
Can we delete the branch?

csviri:
sure

PR #2806, opened 2025-05-16
fix: restore backwards compatibility

metacosm:


csviri (approved):

metacosm:
> Hmm didn't we move the matcher to configuration object?

This is the configuration object.

csviri:
I mean if we can remove the constructor below.

metacosm:
> I mean if we can remove the constructor below.

No, it's needed by both the builder and the converter.

PR #2809, opened 2025-05-19
docs: document annotation-based dependent resource configuration

metacosm:
- **docs: document annotation-based dependent resource configuration**
- **fix: remove now unneeded interface**

### DIFF
--- a/docs/content/en/docs/documentation/configuration.md
+++ b/docs/content/en/docs/documentation/configuration.md
@@ -113,11 +113,9 @@ for this feature.
 
 ## DependentResource-level configuration
 
-It is possible to define custom annotations to configure custom `DependentResource` implementations. For example, the
-SDK provides the `@KubernetesDependent` annotation which allows configuring `KubernetesDependentResource` instance.
-
-In order to provide such a configuration mechanism for your own `DependentResource` implementations, they must be
-annotated with the `@Configured` annotation. This annotation defines 3 fields that tie everything together:
+It is possible to define custom annotations to configure custom `DependentResource` implementations. In order to provide
+such a configuration mechanism for your own `DependentResource` implementations, they must be annotated with the
+`@Configured` annotation. This annotation defines 3 fields that tie everything together:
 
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/documentation/configuration.md
```diff
@@ -113,14 +113,33 @@ for this feature.
...
-`DependentResource` implementations can implement the `DependentResourceConfigurator` interface 
-to pass information to the implementation. For example, the SDK 
-provides specific support for the `KubernetesDependentResource`, which can be configured via the 
-`@KubernetesDependent` annotation. This annotation is, in turn, converted into a 
-`KubernetesDependentResourceConfig` instance, which is then passed to the `configureWith` method 
-implementation.
-
-TODO
+It is possible to define custom annotations to configure custom `DependentResource` implementations. For example, the
+SDK provides the `@KubernetesDependent` annotation which allows configuring `KubernetesDependentResource` instance.
```
csviri, line 117: Would be nice to add a link here to the source code.

csviri (approved):
Added just one comment otherwise LGTM

docs/content/en/docs/documentation/configuration.md
metacosm: To point to what, though? I'm not sure that pointing to the classes sources would help (people can easily find them themselves) and it wouldn't really help, unless we're pointing to the interesting parts, but that would become a maintenance issue, since the references would then need to be updated.

csviri: To this:

https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L31-L34

> people can easily find them themselves

This is way it is much easier, just clicking on link, to see a sample usage. At least for me it helps a lot. Yes it requires maintenance. Another way to do it is add code snippets directly to markdown.

PR #2814, opened 2025-05-21
fix: primary cache utils mechanism

csviri:
Reading from API server was not correct, this works in all cases only if the informer cache has the up to date resources.
If we don't have the up to date resource in the cache, and don't do the update based on that, we cannot say for sure if we can remove the resource for the next event or not from overlay cache.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
@@ -27,8 +27,8 @@ import io.javaoperatorsdk.operator.processing.event.ResourceID;
 public class PrimaryUpdateAndCacheUtils {
 
   public static final int DEFAULT_MAX_RETRY = 10;
-  public static final int RESOURCE_CACHE_POLL_TIMEOUT = 10000;
-  public static final int DEFAULT_SLEEP_FOR_CACHE_POLL_MILLIS = 50;
+  public static final int DEFAULT_RESOURCE_CACHE_TIMEOUT_MILLIS = 10000;
+  public static final int DEFAULT_RESOURCE_CACHE_POLL_PERIOD_MILLIS = 50;
 
   private PrimaryUpdateAndCacheUtils() {}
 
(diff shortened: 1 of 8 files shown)


csviri:
see expalnation https://github.com/operator-framework/java-operator-sdk/pull/2815

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -180,14 +192,35 @@ public static <P extends HasMetadata> P updateAndCacheResource(
...
+              .getControllerConfiguration()
+              .getConfigurationService()
+              .getResourceCloner()
+              .clone(cachedResource);
+        }
+        Thread.sleep(pollDelayMillis);
       }
+      throw new OperatorException("Timeout of resource polling from cache for resource");
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
xstefank, line 222: Is this needed if there InterruptedException?

xstefank (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
csviri: According to Sonar yes.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -25,6 +27,8 @@
 public class PrimaryUpdateAndCacheUtils {
 
   public static final int DEFAULT_MAX_RETRY = 10;
+  public static final int RESOURCE_CACHE_POLL_TIMEOUT = 10000;
```
csviri, line 30: @shawkins do you know some real life data, if on how long it would take on larger cluster to an informer to receive the event? On small one is almost instant (if watch works), not sure how the cluster size influences this.

xstefank (approved):

metacosm (approved):

Issue #2813, opened 2025-05-21
Support field selector in InformerEventSource


csviri:
see: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

PR #2835, opened 2025-06-19
feat: field selectors support for InformerEventSource

csviri:

closes https://github.com/operator-framework/java-operator-sdk/issues/2813

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
@@ -2,7 +2,9 @@ package io.javaoperatorsdk.operator.api.config.informer;
 
 public @interface Field {
 
-  String field();
+  String path();
 
   String value();
+
+  boolean negate() default false;
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
@@ -6,5 +6,5 @@ public @interface Field {
 
   String value();
 
-  boolean negate() default false;
+  boolean negated() default false;
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
@@ -1,0 +1,8 @@
+package io.javaoperatorsdk.operator.api.config.informer;
+
+public @interface Field {
+
+  String field();
+
+  String value();
+}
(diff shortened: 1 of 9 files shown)


csviri:
closes #2813

xstefank (approved):

csviri:
@metacosm @xstefank  this is ready now. Targeting `next` with the features.

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/fieldselector/FieldSelectorTestReconciler.java
```diff
@@ -0,0 +1,67 @@
...
+import io.javaoperatorsdk.operator.api.reconciler.UpdateControl;
+import io.javaoperatorsdk.operator.processing.event.source.EventSource;
+import io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource;
+import io.javaoperatorsdk.operator.support.TestExecutionInfoProvider;
+
+@ControllerConfiguration(
+    informer =
+        @Informer(
+            withFields =
+                @Field(field = "type", value = FieldSelectorTestReconciler.MY_SECRET_TYPE)))
```
xstefank: This seems a little strange to me to read. Maybe "name" or "key" so it's not `@Field(field = ...` but of course this is just recommendation so feel free to discard it.

xstefank (approved):

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/fieldselector/FieldSelectorTestReconciler.java
csviri: Yes, it is a bit smelly. 

Alternatively, we could have:

`@FieldSelector(field="type", value="value", negated=false)`

So there would not be `with` and `without` fields in `@Informer` just a list of `@FieldSelector`.
We might want to have similar api for java api for informers.

csviri: @metacosm what do you think?

csviri: refactored it to a more explicit model, PTAL

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
```diff
@@ -0,0 +1,10 @@
+package io.javaoperatorsdk.operator.api.config.informer;
+
+public @interface Field {
+
+  String path();
+
+  String value();
+
+  boolean negate() default false;
```
metacosm: I would use `negated` here as well to be consistent (and I think it makes more sense that way)

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
csviri: ahh, for sure, thx! fixed

PR #2845, opened 2025-06-30
feat: add feature switch to use SSA for managing finalizer

csviri:
Discovered a weird bug (will describe more in detail later) when adding finalizer with SSA.
Therefore adding explicit feature switch for finalizer management (SSA / no SSA). 
Would like to have a patch release with this, but for next minor release (5.2) would set this flag to `false`, so by default
not use SSA for finalizer managment. 

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -505,16 +505,14 @@ public interface ConfigurationService {
   }
 
   /**
-   * {@link io.javaoperatorsdk.operator.api.reconciler.UpdateControl} patch resource or status can
-   * either use simple patches or SSA. Setting this to {@code true}, controllers will use SSA for
-   * adding finalizers.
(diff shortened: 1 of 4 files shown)


metacosm (approved):
Not a big fan of adding yet another flag to control SSA behavior…

csviri:
yeah me neither TBH, would be better without but, this can be life saving in future.
For v6 we might be able to remove the previous flag (https://github.com/operator-framework/java-operator-sdk/issues/2797)

csviri:
will work on a reproducer before I merge this

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -505,6 +504,18 @@ default boolean useSSAToPatchPrimaryResource() {
     return true;
   }
 
+  /**
+   * Setting this to {@code true}, controllers will use SSA for adding finalizers.
```
metacosm, line 508: It might help to document why it might be a good idea to set this to false, in particular point to the Kubernetes issue.

csviri:
was thinking more about this, and for now this might be not that good idea, let's discuss it further before doing related changes.

PR #2853, opened 2025-07-09
feat: add experimental annotation

csviri:
Annotation would serve to mark experimental features.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Experimental.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Experimental.java
@@ -12,5 +12,13 @@ import java.lang.annotation.Target;
  * the feedback of the users.
  */
 @Retention(RetentionPolicy.SOURCE)
-@Target({ElementType.METHOD, ElementType.TYPE})
-public @interface Experimental {}
+@Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD, ElementType.PACKAGE})
+public @interface Experimental {
+
+  /**
+   * Describes why the annotated element is experimental.
+   *
+   * @return the experimental description.
+   */
+  String value();
+}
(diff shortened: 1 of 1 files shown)


xstefank (commented):
We have similar in SmallRye project - https://github.com/xstefank/smallrye-common/blob/main/annotation/src/main/java/io/smallrye/common/annotation/Experimental.java and I would suggest also giving this one a String argument that says why/what experimental feature this is.

csviri:
> We have similar in SmallRye project - https://github.com/xstefank/smallrye-common/blob/main/annotation/src/main/java/io/smallrye/common/annotation/Experimental.java and I would suggest also giving this one a String argument that says why/what experimental feature this is.

Makes sense, will add, thank you!

csviri:
@xstefank Why is there runtime retention?

xstefank:
@csviri we might pull this in logs or similar based on config.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Experimental.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marks experimental features.
+ *
+ * <p>Experimental features are not yet stable and may change in future releases. Usually based on
```
xstefank, line 11: Maybe we could also say that we keep changes to minor or even micro releases. But that's optional.

metacosm (commented):
Do we intend to do anything specific with this annotation? i.e. automated warning when using experimental features, or such? If not, why not use just documentation for that purpose?
To be clear, not saying we shouldn't add such an annotation, just wondering what we will use it for.

csviri:
> Do we intend to do anything specific with this annotation? i.e. automated warning when using experimental features, or such? If not, why not use just documentation for that purpose? To be clear, not saying we shouldn't add such an annotation, just wondering what we will use it for.

That is fair, we should define/agree on how we are going to use it. My intention is just to mark features which are either not well understood / discovered / standard feature of operator framewors as experimental, until we standardize them or possibly iterate on them a few cycles. IMHO, annotation is not just a little more visible than simple Javadoc, but helps us and the users easily search for the experimental features, and yes also possibly build some additional logic on top of it in case.

csviri:
> Seems like there are unrelated changes added to this PR.

yes will have to rebase or actually might re-create it, thank you!

csviri:
@metacosm can we merge this now? Would like to use it in some of the new PRs

metacosm (approved):

PR #2867, opened 2025-07-16
source bom

csviri:
- **fix: remove source plugin from bom release**
- **fix: temp removing bom release**


metacosm:
Have you checked that releasing will still publish the BOM, because, iirc, it is not using the JOSDK parent?

csviri:
> Have you checked that releasing will still publish the BOM, because, iirc, it is not using the JOSDK parent?

Hi yes, in final version it is using the release related plugins directly. It published the bom.

PR #2874, opened 2025-07-21
fix: minikube setup does not use github token

csviri:
Currently integration tests are being throttled, this should not happen when github token is used.
It seems that sub-workflows it should be passed or used as `github.token`, see here:
https://docs.github.com/en/actions/tutorials/use-github_token-in-workflows#using-the-github_token-in-a-workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,9 @@ jobs:
     strategy:
       matrix:
         java: [ 17, 21, 24 ]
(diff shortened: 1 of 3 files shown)


xstefank (approved):

csviri:
It seems that the issue was with minikube itself; it makes HTTP calls to the GitHub API if a newer version of Kubernetes is passed to it than is officially supported. (for example 1.33.3 is not supported but 1.33.1 is)

cc @xstefank @manusa @metacosm

PR #2881, opened 2025-07-31
fix: startup all resource indexing

csviri:
Indexes all the resources in the cache on startup, making sure that all resources are on the index for first reconiliation.

### DIFF
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefIT.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefIT.java
@@ -44,22 +44,6 @@ class DependentResourceCrossRefIT {
                 assertThat(operator.get(Secret.class, TEST_RESOURCE_NAME + i)).isNotNull();
               }
             });
-
-    for (int i = 0; i < EXECUTION_NUMBER; i++) {
-      operator.delete(testResource(i));
-    }
-    await()
-        .timeout(Duration.ofSeconds(30))
-        .untilAsserted(
-            () -> {
-              for (int i = 0; i < EXECUTION_NUMBER; i++) {
-                assertThat(
-                        operator.get(
-                            DependentResourceCrossRefResource.class,
-                            testResource(i).getMetadata().getName()))
-                    .isNull();
-              }
-            });
   }
 
   DependentResourceCrossRefResource testResource(int n) {
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefReconciler.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefReconciler.java
@@ -5,9 +5,6 @@ import java.util.Base64;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
 import io.fabric8.kubernetes.api.model.Secret;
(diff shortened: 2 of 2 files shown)


operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/startsecondaryaccess/StartupSecondaryAccessIT.java
```diff
@@ -0,0 +1,53 @@
...
+              })
+          .build();
+
+  @Test
+  void reconcilerSeeAllSecondaryResources() {
+    var reconciler = extension.getReconcilerOfType(StartupSecondaryAccessReconciler.class);
+
+    await().untilAsserted(() -> assertThat(reconciler.isReconciled()).isTrue());
+
+    assertThat(reconciler.isSecondaryAndCacheSameAmount()).isTrue();
```
xstefank, line 51: Are the CMs cleaned if they are created like this?

xstefank (approved):

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/startsecondaryaccess/StartupSecondaryAccessIT.java
csviri: CM-s are cleared by K8S garbage collector, on delete even clears then our caches.

xstefank (approved):

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -424,6 +432,15 @@ public Builder withAdditionalCRD(String... paths) {
       return this;
     }
 
+    /**
+     * Used to initialize resources when the namespace is generated but the operator is not started
+     * yet.
+     */
+    public Builder withBeforeStartHook(Consumer<LocallyRunOperatorExtension> beforeStartHook) {
```
metacosm, line 439: Might need a "before-shutdown" hook as well to perform clean-up if needed.

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/startsecondaryaccess/StartupSecondaryAccessReconciler.java
```diff
@@ -0,0 +1,75 @@
...
+    var cached = cmInformer.list().toList();
+
+    log.info(
+        "Secondary number: {}, cached: {}, expected: {}",
+        secondary.size(),
+        cached.size(),
+        SECONDARY_NUMBER);
+
+    if (secondary.size() != cached.size()) {
+      secondaryAndCacheSameAmount = false;
```
metacosm, line 49: why not simply:
```java
secondaryAndCacheSameAmount = secondary.size() == cache.size();
```

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefIT.java
```diff
@@ -44,6 +44,22 @@ void dependentResourceCanReferenceEachOther() {
                 assertThat(operator.get(Secret.class, TEST_RESOURCE_NAME + i)).isNotNull();
               }
             });
+
+    for (int i = 0; i < EXECUTION_NUMBER; i++) {
```
metacosm: This seems unrelated?

metacosm (approved):
LGTM with minor nitpicks.

operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/startsecondaryaccess/StartupSecondaryAccessReconciler.java
csviri: That would not work if later is the same value would overwrite the false, so here the aim is that if once false therefore always false

operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/dependentresourcecrossref/DependentResourceCrossRefIT.java
csviri: Yeah, this test was failing with a different version of the fix, and improved it while investigating, I can make it as a separate PR.

operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
csviri: That is usually not needed strictly , since one can clean the resources at the end of the test, so functionally that is doable. But agree that might be a nicer api.

PR #2882, opened 2025-08-01
feat: native open telemetry metrics

### LINKED ISSUE #2135: Investigate integration possibilities with Opentelemetry

csviri:


csviri:
Started to prototype this, the thing is that there is a OTel adapter for micrometer:
https://docs.micrometer.io/micrometer/reference/implementations/otlp.html

So it should be already possible to integrate with OTel. A native support still would be nice.

However instantly hit this, there is a workaround for sure where we will manage our counts, but it is not nice that counter cannot be closed in sync api.
https://stackoverflow.com/questions/76652117/how-to-close-or-destroy-a-specific-counter-in-opentelemetry

cc @metacosm @xstefank 

We might want to discuss it on weekly dev call

csviri:
This is supported through micrometer otel adaptor, see related e2e tests. That seems to be good enough.

Issue #2885, opened 2025-08-05
Ignored readyPostcondition by dependent


ds-akloskowski:
#### What did you do?

I set up a workflow as in example below (just to show dependencies between resources).

```java
@Workflow(dependents = {
    @Dependent(type = AlphaSecretDependent.class,
        name = "Alpha",
        readyPostcondition = AlphaSecretReadyCondition.class,
        useEventSourceWithName = "alpha-source"
    ),
    @Dependent(type = BetaSecretDependent.class,
        name = "Beta",
        readyPostcondition = BetaSecretReadyCondition.class,
        useEventSourceWithName = "beta-source"
    ),
    @Dependent(type = GammaJWTDependent.class,
        name = "Gamma",
        dependsOn = {"Alpha", "Beta"}
    )
})
```

`AlphaSecretDependent` and `BetaSecretDependent` are secrets made to be created outside an operator and further used by `GammaJWTDependent` to produce JWT. Ready post conditions check if resources met requirements like if keys exist in secrets.

The workflow above I used to make a test in which `BetaSecretDependent` is not provided and `BetaSecretReadyCondition` was always false to check if code to search secrets worked properly and if `GammaJWTDependent` is not reconciled.

#### What did you expect to see?

I expected to see in test that `GammaJWTDependent` was not reconciled and there was no attempt to reconcile it.

#### What did you see instead? Under which circumstances?

In a few cases, looking non-deterministic, there was an attempt to reconcile `GammaJWTDependent`. Looked like `BetaSecretReadyCondition` was true (but wasn't) or ignored, but it was more like a race condition I believe. (More about it in "possible solution").

#### Environment

**Kubernetes cluster type:**
kube-apiserver

`$ Mention java-operator-sdk version from pom.xml file`
java-operator-sdk 5.1.1

`$ java -version`
java version "21.0.7" 2025-04-15 LTS
Java(TM) SE Runtime Environment (build 21.0.7+8-LTS-245)
Java HotSpot(TM) 64-Bit Server VM (build 21.0.7+8-LTS-245, mixed mode, sharing)

`$ kubectl version`
Client Version: v1.32.2
Kustomize Version: v5.5.0

#### Possible Solution

[...]

PR #2886, opened 2025-08-06
fix: Resolve race condition between visited flag and result of ready postcondition

ds-akloskowski:
Fix for a problem described in #2885. 

The changed code causes that when `visited` flag is changed to true, we are sure that readyPostconditionResult was checked before. In previous version, `visited` flag set to true and readyPostconditionResult equals null could describe two situations:
1. A resource was visited and its readyPostcondition is null.
2. A resource was visited and its readyPostcondition exists, but wasn't wasn't checked (race condition).

The changed code doesn't let the second situation to happen.


xstefank (approved):

csviri:
Makes sense, thank you @ds-akloskowski

metacosm:
Thank you!

PR #2891, opened 2025-08-08
improve: remove misleading log message

csviri:
This warn message is logged even if missing config is expected

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/AbstractConfigurationService.java
@@ -113,20 +113,7 @@ public class AbstractConfigurationService implements ConfigurationService {
       Reconciler<R> reconciler) {
     final var key = keyFor(reconciler);
     final var configuration = configurations.get(key);
-    if (configuration == null) {
-      logMissingReconcilerWarning(key, getReconcilersNameMessage());
-    }
     return configuration;
-  }
-
-  protected void logMissingReconcilerWarning(String reconcilerKey, String reconcilersNameMessage) {
-    log.warn("Cannot find reconciler named '{}'. {}", reconcilerKey, reconcilersNameMessage);
-  }
-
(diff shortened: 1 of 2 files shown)


metacosm:
I don't think this is the appropriate fix. The appropriate fix would be to override `logMissingReconcilerWarning` in `BaseConfigurationService`to only output the warning if `createIfNeeded` returns `true`. We could simplify the code even more and maybe remove `createIfNeeded` altogether but this would be an API breaking change (in case someone depends on `BaseConfigurationService` for their own `ConfigurationService` implementation).

csviri:
The problem now is that this warning is logged always, even if everything is ok from JOSDK perspective. I think it is still better to remove it and not confuse users, and finetune it later.

metacosm:
See #2892, the warning should not be output anymore with that implementation

csviri:
What I fail to see how this message helps the user. If any we should have such log message maybe on debug level that helps development / troubleshooting, but not on warning level. But in addition to that, mainly we should have a test for the API, so configurations are correctly initialized, and would remove this log message. 

@metacosm in case could you explain how this help pls?

thank you!

cc @xstefank

metacosm:
The use case is that not all `ConfigurationService` implementations should or want to create the configuration on demand. In some cases, having that method allows implementation to even throw an exception for example if the requested configuration is missing when it shouldn't be (because, for example, all configurations were supposed to have already been resolved at the call point).

csviri:
Yes, throwing exception in some cases would be fine, but we should not log warning.  (In addition to that context of josdk it is expected that this method returns null in some cases.)

metacosm:
> Yes, throwing exception in some cases would be fine, but we should not log warning. (In addition to that context of josdk it is expected that this method returns null in some cases.)

And with the implementation in the other PR, no logging occurs.

csviri:
commented there, I don't see what is the problem we are solving with that PR. Pls choose where to continue the discussion :)

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/2892

PR #2892, opened 2025-08-08
fix: do not output warning when resolving a configuration

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -485,7 +485,6 @@ public interface ConfigurationService {
    *
    * @return if resource version should be parsed (as integer)
    * @since 4.5.0
-   * @return if resource version should be parsed (as integer)
    */
   default boolean parseResourceVersionsForEventFilteringAndCaching() {
     return false;


operator-framework/src/test/java/io/javaoperatorsdk/operator/config/BaseConfigurationServiceTest.java
```diff
@@ -81,6 +81,13 @@ private static KubernetesDependentResourceConfig extractDependentKubernetesResou
     return (KubernetesDependentResourceConfig) configuration.getConfigurationFor(spec);
   }
 
+  @Test
+  void test() {
+    final var service = new BaseConfigurationService();
+    final var config = service.getConfigurationFor(new NoDepReconciler());
+    System.out.println(config);
```
xstefank: This test should be asserting the logged message or removed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
```diff
@@ -318,6 +326,13 @@ private <P extends HasMetadata> ResolvedControllerConfiguration<P> controllerCon
         informerConfig);
   }
 
+  /**
+   * @deprecated This method was meant to allow subclasses to prevent automatic creation of the
+   *     configuration when not found. This functionality is now removed, if you want to be able to
+   *     prevent automated, on-demand creation of a reconciler's configuration, please use the
+   *     {@link AbstractConfigurationService} implementation instead as base for your extension.
+   */
+  @Deprecated(forRemoval = true)
```
xstefank, line 335: Somewhat strange to introduce a new usage of the method and deprecate it at the same time. How will this be handled when this method is removed?

xstefank (commented):
I'm not sure which solution is better for this problem, but if this would go in there are a few points.

operator-framework/src/test/java/io/javaoperatorsdk/operator/config/BaseConfigurationServiceTest.java
metacosm: Ah yes, this wasn't supposed to end up in the commit… 😅 
I thought about asserting the lack of logging but this would require changing the architecture (i.e. make the logger an instance variable instead of a class one) just for the tests so decided against it.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
metacosm: The current implementation uses the method but the documentation explains what should happen instead. It's only deprecated to avoid API breakage in case implementations use it but this method should really be removed, because, as far as I'm aware, there is no implementation that doesn't currently return `true` and the creation scenario should only be confined to this particular implementation (or its subclasses) anyway.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
```diff
@@ -149,10 +155,12 @@ private static void configureFromAnnotatedReconciler(
 
   @Override
   protected void logMissingReconcilerWarning(String reconcilerKey, String reconcilersNameMessage) {
-    logger.warn(
-        "Configuration for reconciler '{}' was not found. {}",
-        reconcilerKey,
-        reconcilersNameMessage);
+    if (!createIfNeeded()) {
```
csviri, line 158: what is the probelm we are solving this method evetually?

operator-framework/src/test/java/io/javaoperatorsdk/operator/config/BaseConfigurationServiceTest.java
csviri: if is does not test anything, should we remove the test?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/BaseConfigurationService.java
csviri: I mean in quarkus operator sdk

metacosm: Right now, the issue is preserving the possibility to log (or do something else, if needed) if the configuration doesn't exist in the **abstract** class, without logging when the configuration is created in `BaseConfigurationService`. Removing the logging and `createIfNeeded` methods would be API breaking changes so the goal is to have an implementation that doesn't break the existing API and doesn't log a warning when a configuration is created on-demand.

csviri: Thank you, approved

csviri (approved):

PR #2899, opened 2025-08-15
fix: create state only on resource event

csviri:
In general we cleanup state on delete event, so although in practice this probably not causing
issues in theory it can happen that we receive events where there is no related custom resource.
What effectivel means a memory leak, since that state never cleaned up - again it is more theoretical than practical issues, since it practice usually events are propagated related to a custom resource (or based on that).

This PR fixes it in a way that it checks that when state is created it is alway related to a resource event (thus for primary / custom resource). In that way avoiding that state is created without a primary resource. 


xstefank (approved):

csviri:
In close future I will iterate in this probably to clean things up more.

PR #2911, opened 2025-08-26
improve: add aider to git ignore

csviri:


xstefank (approved):
AI is finding its way to our JOSDK workflow I see :)

PR #2912, opened 2025-08-26
improve: Make MicrometerMetrics extendable

### LINKED ISSUE #2884: Custom bootstrap and CR lifecycle management
This is mostly a reality check if our custom bootstrap is using JOSDK as intended or not. Depending on the feedback we can close this issue and create more specific follow-up issues if needed. I just wanted to describe the context once.

```java
// configure operator
final var customMetrics = new CustomResourceMetrics(config, meterRegistry);
final var eventSender = new EventSender(config, client);
final var reconciler = new HiveMQPlatformReconciler(config, customResourceMetrics, client, eventSender);

final var dependentResourceFactory = new HiveMQPlatformOperatorDependentResourceFactory<>(config, eventSender);
final var metrics = MicrometerMetrics.newPerResourceCollectingMicrometerMetricsBuilder(meterRegistry).build();
final var operator = new Operator(override -> override //
        .withConcurrentReconciliationThreads(config.getConcurrentReconciliationThreads())
        .withConcurrentWorkflowExecutorThreads(config.getConcurrentWorkflowThreads())
        .withReconciliationTerminationTimeout(config.getReconciliationTerminationTimeout())
        .withCacheSyncTimeout(config.getCacheSyncTimeout())
        .withKubernetesClient(client)
        .withCloseClientOnStop(closeClient)
        .withDependentResourceFactory(dependentResourceFactory)
        .withMetrics(metrics)
        .withUseSSAToPatchPrimaryResource(useSSA)
        .withSSABasedCreateUpdateMatchForDependentResources(useSSA));
final var reconcilerConfig = operator.getConfigurationService().getConfigurationFor(reconciler); // (1)
final var overrideConfig = ControllerConfigurationOverrider.override(reconcilerConfig);
if (operatorNamespaces.equals(Constants.WATCH_CURRENT_NAMESPACE)) {
    overrideConfig.watchingOnlyCurrentNamespace();
} else if (operatorNamespaces.equals(Constants.WATCH_ALL_NAMESPACES)) {
    overrideConfig.watchingAllNamespaces();
} else {
    final var namespaces = Arrays.stream(operatorNamespaces.split(",")).collect(Collectors.toSet());
    overrideConfig.settingNamespaces(namespaces);
}
if (!operatorSelector.isBlank()) {
    overrideConfig.withLabelSelector(operatorSelector);
}
overrideConfig.withOnAddFilter(platform -> { // (3)
    customResourceMetrics.register(platform);
    return true;
});
final var controller = (Controller<HiveMQPlatform>) operator.register(reconciler, overrideConfig.build());
customResourceMetrics.setCache(controller.getEventSourceManager().getControllerEventSource()); // (2)
```

* `HiveMQPlatformOperatorDependentResourceFactory implements DependentResourceFactory` (it's the only place where we really had to replace Quarkus Arc, otherwise we don't need a CDI).
* `closeClient` is `true` in production and only set to `false` in tests to not close an injected K8s client (that is still used in the test).
* `useSSA` is `true` in production and only set to `false` in tests with the K8s mockserver.

1. Are we creating the configuration and overrides as intended? I tried to reverse engineer how Quarkus and the `LocallyRunOperatorExtension` are configuring JOSDK, but we still get a warning on the `operator.getConfigurationService().getConfigurationFor(reconciler)` invocation:
    ```
    12:29:37.423 [main] WARN  Default ConfigurationService implementation - Configuration for reconciler
    'hivemq-controller' was not found. Known reconcilers: None.
    ```
    It feels wrong to get that warning, but I found no other way to create a `ControllerConfigurationOverrider`.
2. How to properly implement custom metrics on all custom resources in an efficient way? For example, in `CustomResourceMetrics` we have a global `Gauge` for each state of our operator state machine. To calculate the values we count all custom resources that are in that state:
    ```java
    cache.list()
        .map(CustomResource::getStatus)
        .filter(Objects::nonNull)
        .filter(status -> status.getState().equals(state))
        .count();
    ````
    That `cache` is the `ControllerEventSource` that we set via `customResourceMetrics.setCache(controller.getEventSourceManager().getControllerEventSource())`. This feels a bit illegal but works very fine for us.
    In the previous implementation we kept a shadow copy of all custom resources in a CHM, but these were the cloned instances from the reconciliation loop. This impacted the GC and needed constant updates of the cached instances to retrieve the current state. Using the original instances from the cache solved this problem very elegantly for us, including the lifecycle management.
3. The same problem hits us now again with a new `Gauge` for an aggregated health status metric per custom resource. We need to register this `Gauge` once when the CR is added, de-register it when it's removed and in between have access to the current `CustomResource::getStatus` to get the health state. So we're back to a `Map<String, GaugeHolder>` with `<namespace>-<name>` as key.
For the `Gauge` creation I'm using `overrideConfig.withOnAddFilter()`, which again feels very wrong, but seems to work fine. The `Gauge` removal is done in the `Cleaner::cleanup` method of our reconciler.
Is there a better way for the lifecycle control and where to put that `Gauge`? It feels like we cannot put this into the custom resource itself, because of the cloning (the `Gauge` needs to be a singleton) and to update the correct instance. This is how it looks like in the `CustomResourceMetrics` right now:
    ```java
    private final @NotNull Map<String, GaugeHolder> healthMetricGauges = new ConcurrentHashMap<>();

    // called from the bootstrap
    public void register(final @NotNull HiveMQPlatform platform) {
        healthMetricGauges.put(getKey(platform), new GaugeHolder(platform));
    }

    // called from Cleaner::cleanup
    public void deregister(final @NotNull HiveMQPlatform platform) {
        if (healthMetricGauges.get(getKey(platform)) instanceof GaugeHolder gaugeHolder) {
            meterRegistry.remove(gaugeHolder.gauge);
        }
    }

    // called from the main reconciler when the health state has changed
    public void updateHealthMetric(final @NotNull HiveMQPlatform platform) {
        if (healthMetricGauges.get(getKey(platform)) instanceof GaugeHolder gaugeHolder) {
            gaugeHolder.value.set(platform.getStatus().getHealthStatus().getMetricsValue());
        }
    }

    private static @NotNull String getKey(final @NotNull HiveMQPlatform platform) {
        return platform.getMetadata().getNamespace() + "-" + platform.getMetadata().getName();
    }

    private class GaugeHolder {

        private final @NotNull AtomicInteger value = new AtomicInteger(HealthStatus.UNKNOWN.getMetricsValue());
        private final @NotNull Gauge gauge;

        public GaugeHolder(final @NotNull HiveMQPlatform platform) {
            this.gauge = Gauge.builder("hivemq.platform.health.system.current", value::get)
                    .tag("namespace", platform.getMetadata().getNamespace())
                    .tag("name", platform.getMetadata().getName())
                    .description("Custom resource health status")
                    .register(meterRegistry);
        }
    }
    ```

Donnerbart:
Partially fixes #2884

As mentioned in the issue:
> I think implementing a custom metrics interface, or extending the current implementation with micrometer, should do the job

The latter is not possible due to the private constructor and private Cleaner implementations. This is an attempt to make the `MicrometerMetrics` class extendable without exposing all `Cleaner` implementations.

### DIFF
--- a/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
+++ b/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
@@ -96,14 +96,13 @@ public class MicrometerMetrics implements Metrics {
 
   /**
    * Creates a micrometer-based Metrics implementation that cleans up {@link Meter}s associated with
-   * deleted resources as specified by the (possibly {@code null}) provided {@link Cleaner}
-   * instance.
+   * deleted resources as specified by the provided {@link Cleaner} instance.
    *
    * @param registry the {@link MeterRegistry} instance to use for metrics recording
    * @param cleaner the {@link Cleaner} to use
    * @param collectingPerResourceMetrics whether to collect per resource metrics
    */
-  private MicrometerMetrics(
+  protected MicrometerMetrics(
       MeterRegistry registry, Cleaner cleaner, boolean collectingPerResourceMetrics) {
     this.registry = registry;
     this.cleaner = cleaner;
@@ -315,11 +314,9 @@ public class MicrometerMetrics implements Metrics {
   public static class PerResourceCollectingMicrometerMetricsBuilder
       extends MicrometerMetricsBuilder {
 
-    private int cleaningThreadsNumber;
-    private int cleanUpDelayInSeconds;
-
     private PerResourceCollectingMicrometerMetricsBuilder(MeterRegistry registry) {
       super(registry);
+      collectingPerResourceMetrics = true;
     }
 
     /**
@@ -329,7 +326,7 @@ public class MicrometerMetrics implements Metrics {
      */
     public PerResourceCollectingMicrometerMetricsBuilder withCleaningThreadNumber(
         int cleaningThreadsNumber) {
-      this.cleaningThreadsNumber = cleaningThreadsNumber <= 0 ? 1 : cleaningThreadsNumber;
+      cleanerBuilder.withCleaningThreadNumber(cleaningThreadsNumber);
       return this;
     }
 
@@ -343,30 +340,24 @@ public class MicrometerMetrics implements Metrics {
      */
     public PerResourceCollectingMicrometerMetricsBuilder withCleanUpDelayInSeconds(
         int cleanUpDelayInSeconds) {
-      this.cleanUpDelayInSeconds = Math.max(cleanUpDelayInSeconds, 1);
+      cleanerBuilder.withCleanUpDelayInSeconds(cleanUpDelayInSeconds);
       return this;
-    }
-
-    @Override
-    public MicrometerMetrics build() {
-      final var cleaner =
-          new DelayedCleaner(registry, cleanUpDelayInSeconds, cleaningThreadsNumber);
-      return new MicrometerMetrics(registry, cleaner, true);
     }
   }
 
   public static class MicrometerMetricsBuilder {
     protected final MeterRegistry registry;
-    private boolean collectingPerResourceMetrics = true;
+    protected final CleanerBuilder cleanerBuilder;
+    protected boolean collectingPerResourceMetrics = true;
 
     private MicrometerMetricsBuilder(MeterRegistry registry) {
       this.registry = registry;
+      this.cleanerBuilder = new CleanerBuilder(registry);
     }
 
     /** Configures the instance to collect metrics on a per-resource basis. */
     @SuppressWarnings("unused")
     public PerResourceCollectingMicrometerMetricsBuilder collectingMetricsPerResource() {
-      collectingPerResourceMetrics = true;
       return new PerResourceCollectingMicrometerMetricsBuilder(registry);
     }
 
@@ -381,11 +372,52 @@ public class MicrometerMetrics implements Metrics {
     }
 
     public MicrometerMetrics build() {
-      return new MicrometerMetrics(registry, Cleaner.NOOP, collectingPerResourceMetrics);
+      return new MicrometerMetrics(registry, cleanerBuilder.build(), collectingPerResourceMetrics);
     }
   }
 
-  interface Cleaner {
+  public static class CleanerBuilder {
+
+    private final MeterRegistry registry;
+    private int cleaningThreadsNumber = 0;
+    private int cleanUpDelayInSeconds = 0;
+
+    public CleanerBuilder(MeterRegistry registry) {
+      this.registry = registry;
+    }
+
+    /**
+     * @param cleaningThreadsNumber the maximal number of threads that can be assigned to the
+     *     removal of {@link Meter}s associated with deleted resources, defaults to 1 if not
+     *     specified or if the provided number is lesser or equal to 0
+     */
+    public CleanerBuilder withCleaningThreadNumber(int cleaningThreadsNumber) {
+      this.cleaningThreadsNumber = cleaningThreadsNumber <= 0 ? 1 : cleaningThreadsNumber;
+      return this;
+    }
+
+    /**
+     * @param cleanUpDelayInSeconds the number of seconds to wait before {@link Meter}s are removed
+     *     for deleted resources, defaults to 1 (meaning meters will be removed one second after the
+     *     associated resource is deleted) if not specified or if the provided number is lesser than
+     *     0. Threading and the general interaction model of interacting with the API server means
+     *     that it's not possible to ensure that meters are immediately deleted in all cases so a
+     *     minimal delay of one second is always enforced
+     */
+    public CleanerBuilder withCleanUpDelayInSeconds(int cleanUpDelayInSeconds) {
+      this.cleanUpDelayInSeconds = Math.max(cleanUpDelayInSeconds, 1);
+      return this;
+    }
+
+    public Cleaner build() {
+      if (cleanUpDelayInSeconds > 0 || cleaningThreadsNumber > 0) {
+        return new DelayedCleaner(registry, cleanUpDelayInSeconds, cleaningThreadsNumber);
+      }
+      return Cleaner.NOOP;
+    }
+  }
+
+  public interface Cleaner {
     Cleaner NOOP = new Cleaner() {};
 
     default void removeMetersFor(ResourceID resourceID) {}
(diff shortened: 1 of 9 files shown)


micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -381,11 +373,79 @@ public MicrometerMetricsBuilder notCollectingMetricsPerResource() {
     }
 
     public MicrometerMetrics build() {
-      return new MicrometerMetrics(registry, Cleaner.NOOP, collectingPerResourceMetrics);
+      return new MicrometerMetrics(registry, cleanerBuilder.build(), collectingPerResourceMetrics);
+    }
+  }
+
+  public static class CleanerBuilder {
```
metacosm, line 379: Does this need to be exposed?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
```diff
@@ -381,11 +373,79 @@ public MicrometerMetricsBuilder notCollectingMetricsPerResource() {
...
+
+    public CleanerBuilder(MeterRegistry registry) {
+      this.registry = registry;
+    }
+
+    /**
+     * @param cleanerType the type of cleaner to use, defaults to {@link CleanerType#NOOP} if not
+     *     specified
+     */
+    public CleanerBuilder withCleanerType(CleanerType cleanerType) {
```
metacosm: I would rather not expose the type if possible but rather switch the type internally based on whether the user sets the number of cleaning threads or the delay, which would switch the implementation rather than throwing an exception if the type is not amenable to set these values.

metacosm (changes requested):

Donnerbart:
My main motivation to extend the `MicrometerMetrics` class is to avoid using a delegate pattern. This is how we've implemented this right now:
```java
public class CustomResourceMetrics implements Metrics {

  private final @NotNull Metrics delegate;

  @Override
  public void controllerRegistered(final @NotNull Controller<? extends HasMetadata> controller) {
    delegate.controllerRegistered(controller);
    ...
  }

  @Override
  public void reconcileCustomResource(
          final @NotNull HasMetadata resource,
          final @NotNull RetryInfo retryInfo,
          final @NotNull Map<String, Object> metadata) {
      delegate.reconcileCustomResource(resource, retryInfo, metadata);
  }
}
```
This works fine, but we need to implement all methods of `Metrics` to preserve the JOSDK provided metrics. And if there would ever be a new default method in `Metrics` that is implemented in `MicrometerMetrics`, it would silently break something.

An alternative approach would be to allow multiple `Metrics` implementations. Then we could just configure the `MicrometerMetrics` as they are and our custom metrics together. But this might actually break public APIs, especially the getter signatures:
```java
public class ConfigurationServiceOverrider {

  private List<Metrics> metrics;

  public ConfigurationServiceOverrider withMetrics(Metrics metrics) {
    return withMetrics(List.of(metrics));
  }
  
  public ConfigurationServiceOverrider withMetrics(List<Metrics> metrics) {
    this.metrics = metrics;
    return this;
  }
  
  public ConfigurationService build() {
    return new BaseConfigurationService(original.getVersion(), cloner, client) {
      @Override
      public List<Metrics> getMetrics() {
        return metrics != null ? metrics : original.getMetrics();
      }
   }
}
```
```java
public interface ConfigurationService {

  default List<Metrics> getMetrics() {
    return List.of(Metrics.NOOP);
  }
}
```

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
Donnerbart: The problem is that the constructor needs a `Cleaner` implementation:
```java
protected MicrometerMetrics(
      MeterRegistry registry, Cleaner cleaner, boolean collectingPerResourceMetrics) {
```
So how can we re-use the existing implementations (beside the static `Cleaner.NOOP`)?

I didn't want to make them all public, so I tried to hide them behind that new `CleanerBuilder`.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
Donnerbart: That might actually work, yes! With that we might not need the types at all. The `build()` method could just check if a delay parameter is set, otherwise use Cleaner.NOOP. The `DefaultCleaner` is actually not directly used and exposed right now at all, so no need to change that.

metacosm:
> An alternative approach would be to allow multiple `Metrics` implementations. Then we could just configure the `MicrometerMetrics` as they are and our custom metrics together. But this might actually break public APIs, especially the getter signatures:
> 
> ```java
> public class ConfigurationServiceOverrider {
> 
>   private List<Metrics> metrics;
> 
>   public ConfigurationServiceOverrider withMetrics(Metrics metrics) {
>     return withMetrics(List.of(metrics));
>   }
>   
>   public ConfigurationServiceOverrider withMetrics(List<Metrics> metrics) {
>     this.metrics = metrics;
>     return this;
>   }
>   
>   public ConfigurationService build() {
>     return new BaseConfigurationService(original.getVersion(), cloner, client) {
>       @Override
>       public List<Metrics> getMetrics() {
>         return metrics != null ? metrics : original.getMetrics();
>       }
>    }
> }
> ```
> 
> ```java
> public interface ConfigurationService {
> 
>   default List<Metrics> getMetrics() {
>     return List.of(Metrics.NOOP);
>   }
> }
> ```

Alternatively, one could implement an aggregate `Metrics` implementation that would allow to register multiple `Metrics` implementations and would then dispatch to all registered implementations for each method call. This would be similar to the delegate pattern I guess but a little bit more flexible. 

For that matter, such an aggregate implementation could even be provided by JOSDK.

Donnerbart:
> Alternatively, one could implement an aggregate `Metrics` implementation that would allow to register multiple `Metrics` implementations and would then dispatch to all registered implementations for each method call. This would be similar to the delegate pattern I guess but a little bit more flexible.

I think that might actually be the smartest approach with the least impact to the existing classes and APIs! I'm happy to rewrite this PR or create an alternative one with that approach.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java
Donnerbart: I've updated the PR to get rid of the CleanerType and derive the type dynamically from the set builder parameters.

metacosm:
> > Alternatively, one could implement an aggregate `Metrics` implementation that would allow to register multiple `Metrics` implementations and would then dispatch to all registered implementations for each method call. This would be similar to the delegate pattern I guess but a little bit more flexible.
> 
> I think that might actually be the smartest approach with the least impact to the existing classes and APIs! I'm happy to rewrite this PR or create an alternative one with that approach.

Whatever you think is more appropriate but I do think that this approach would be better, indeed. What do you think @csviri, @xstefank?

Issue #1999, opened 2023-08-07
Dependent Resource Creation behavior configurable


csviri:
With sever side apply in dependent resources, it can happen that the resource is already created ( if no owner reference present, the cache will not contain it). Currently in this case the creation will fail. Although there might be use cases when simply desired that this resource is overwritten. To delegate this decision to the user a feature flag can be introduced to make this configurable.
This is feature flag per resource (not a global one).

see also discussion: https://github.com/operator-framework/java-operator-sdk/discussions/1995

PR #2915, opened 2025-08-26
refactor: optimize retrieval of secondary resource

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
@@ -1,9 +1,7 @@
 package io.javaoperatorsdk.operator.processing.dependent.kubernetes;
 
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -264,17 +262,9 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
   }
 
   @Override
-  protected Optional<R> selectTargetSecondaryResource(
-      Set<R> secondaryResources, P primary, Context<P> context) {
-    ResourceID managedResourceID = targetSecondaryResourceID(primary, context);
-    return secondaryResources.stream()
-        .filter(
-            r ->
-                r.getMetadata().getName().equals(managedResourceID.getName())
-                    && Objects.equals(
-                        r.getMetadata().getNamespace(),
-                        managedResourceID.getNamespace().orElse(null)))
-        .findFirst();
+  public Optional<R> getSecondaryResource(P primary, Context<P> context) {
+    final var targetResourceID = targetSecondaryResourceID(primary, context);
+    return eventSource().flatMap(informer -> informer.get(targetResourceID));
   }
 
   /**
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -112,7 +112,6 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
 
   @Override
   public Optional<R> getSecondaryResource(P primary, Context<P> context) {
-
     var secondaryResources = context.getSecondaryResources(resourceType());
     if (secondaryResources.isEmpty()) {
       return Optional.empty();


xstefank (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -264,17 +262,9 @@ protected void addSecondaryToPrimaryMapperAnnotations(
...
-    return secondaryResources.stream()
-        .filter(
-            r ->
-                r.getMetadata().getName().equals(managedResourceID.getName())
-                    && Objects.equals(
-                        r.getMetadata().getNamespace(),
-                        managedResourceID.getNamespace().orElse(null)))
-        .findFirst();
+  public Optional<R> getSecondaryResource(P primary, Context<P> context) {
+    final var targetResourceID = targetSecondaryResourceID(primary, context);
```
csviri, line 266: This is a bit of a change in the behavior, since it skips the owner mechanism that we have with SecondaryToPrimary mappers and related indexing, I rather would not change that. The failing test is a real problem actually, since with current way we are deteciting that the resource is actually related (a.k.a. secondary to) primary resource.

IMO it is also by definition better to use the stanrd appraoch that user would use without dependen resources thus calling the `getSecondaryResource`.

metacosm: This indexing should occur regardless of how secondary resources are retrieved. This new way is much more efficient as it doesn't require to iterate over all the resources and filter them. For the purposes of retrieval, nothing should occur apart from identifying the target resource and then retrieving it from the informer cache. If processing was previously done during that phase, then I would think that the design needs to be revised so that it occurs at some other time, not during secondary retrieval. 

Another aspect is that it should be possible for users to override this method easily without breaking anything. 

Regarding the failing test, I don't understand your point and I actually think that the test is wrong, as discussed separately.

csviri: There are multiple aspects of this:
1. Practical: the change from this PR would simply not work in this scenario: the config map. (as in the test) is created prior the dependent resource is reconciled. From the PR the logic would recognize that the config map is there so won't create / update it. Now we delete the primary resource, since we did not create/update the resource the owner reference it not added, so the config map is not deleted. This is inconsistent with the current and desireg behavior, so simply wrong.

2. The semantical aspect:  we added createifNotExists flag, especially to cover this situation, that we don't override the resource based on desired if we did not create it. Again the devil is in details, we did not create it since there is not owner reference (in other words the resource is not related to primary), so based on this flag we don't want to override it. This is what the failing test is testing, and it is correct.

3. performance:
For exmaple on failing test this aspect does not give any performance gain, we receiving the target resource id, from the primaryToSecondary index in constant time, and getting it from cache directly, this is about the same performance. Note that we usually getting a constant number of resources, looping through is also constant, so from performance perspective of a single dependent is negligible. If you compare the execution time with http call there will be a factor of million.
In addition to that user can simply provide a PrimaryToSeconday mapper, and achive the same result, with a built in appraoch. Not saying that would be much better, just that there is also that option.


So in general, I think this PR should not be merged, since simply breaks the logic point (1.). Would require to change the actual logic for (2.).
We can further discuss performance gains further in case we solve (1,2), but those are really quastinable also in practice / real world examples.

metacosm: > There are multiple aspects of this:
> 
>     1. Practical: the change from this PR would simply not work in this scenario: the config map. (as in the test) is created prior the dependent resource is reconciled. From the PR the logic would recognize that the config map is there so won't create / update it. Now we delete the primary resource, since we did not create/update the resource the owner reference it not added, so the config map is not deleted. This is inconsistent with the current and desireg behavior, so simply wrong.

The scenario as tested by the test is wrong, though. The ConfigMap exists so it shouldn't be created, that's understood. However, the ConfigMapDependentResource defines a desired state which doesn't match the existing CM. So, according to the proper behavior, the CM should be updated. With the current implementation, it is not and that is a bug. The test is wrong and actually shows that the current implementation is not doing what's expected in this instance.

> 
>     2. The semantical aspect:  we added createifNotExists flag, especially to cover this situation, that we don't override the resource based on desired if we did not create it. Again the devil is in details, we did not create it since there is not owner reference (in other words the resource is not related to primary), so based on this flag we don't want to override it. This is what the failing test is testing, and it is correct.

I don't agree with this. The desired method should prevail in all instances since it encapsulates what the operator thinks the proper state is. If an existing state is to be preserved, then the desired method should take this into account, not a flag that breaks the rest of the behavior. Or this resource shouldn't be a dependent resource if a desired state cannot appropriately be computed from the primary resource. I guess I should have paid more attention when that flag was added because I think it is wrong and goes against the rest of the design, especially with it being true by default. Not only that but the name of the constant and the flag do not appropriately convey what will happen.

> 
>     3. performance:
>        For exmaple on failing test this aspect does not give any performance gain, we receiving the target resource id, from the primaryToSecondary index in constant time, and getting it from cache directly, this is about the same performance. Note that we usually getting a constant number of resources, looping through is also constant, so from performance perspective of a single dependent is negligible. If you compare the execution time with http call there will be a factor of million.

Performance matter everywhere. Indeed, most of our processing will be faster than a network call but is that a valid excuse to have inefficient code? Also, all those little things add up. In any case, the core of the issue is not so much performance than the facts that:
1. users should be able to override this public method without breaking things as was previously the case
2. the flag that supposedly deals with SSA creation of resources really doesn't and is wrong with the rest of the framework design. Either it needs to be removed, set to false by default and/or properly justified with a valid use case. Again, if the desired state cannot be computed from a primary resource then the associated resource should **NOT** be a dependent resource, that's the base of the dependent resource feature.


> 
> So in general, I think this PR should not be merged, since simply breaks the logic point (1.). Would require to change the actual logic for (2.). We can further discuss performance gains further in case we solve (1,2), but those are really quastinable also in practice / real world examples.

I argue that the logic exhibited in the breaking test is wrong and should be reverted and the associated flag removed as such use cases should not be modeled as dependent resources. Adding that flag was a mistake we need to fix.

csviri: This is the matter of definition of `creator` that is the reason of that flag. Since you can define it either:
1. creator is always the one creates the resource
2. creates creates the resource if not exists (otherwaise can uptate)

To distinguise the two bahavior we added the flag, so we cover all the use cases.

csviri: Going further into the details. Kubernetes defines a flag `controller` on owner references,
see: https://stackoverflow.com/a/65825463/737591
So, for example, there would be a resource with an owner reference where controller is true, we should not update the spec / resource - this we don't handle explicitly yet in DR (probably we should even for case (2) - will create an issue for that to further discuss.

metacosm: I didn't know about the controller field but
1. it's not completely implemented / supported in Kubernetes
2. this isn't what the flag does at all

Imo, the proper way to support this would be to throw an exception in that case because the desired state cannot be satisfied if the target resource is "controlled" by a different controller as our controller will never be able to match the desired state. The current behavior is improper because it leads the controller to believe that the resource is in the proper state when it's not.

csviri: what do you mean by `it's not completely implemented / supported in Kubernetes` ?

metacosm: > what do you mean by `it's not completely implemented / supported in Kubernetes` ?

At least according to https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/controller-ref.md#implementation
Also, the official documentation doesn't really provide semantics for that field (at least, not that I could find with a quick search)

csviri: I'm pretty sure that you cannot add two owner references with controller: true flag. Maybe something to try out.

csviri:
just putting here for the record related stuff:

https://github.com/operator-framework/java-operator-sdk/pull/2001
https://github.com/operator-framework/java-operator-sdk/issues/1999
https://github.com/operator-framework/java-operator-sdk/discussions/1995

So basically this was discussed in context of KeyCloak operator, and we agreed on this approach.

csviri:
I was thinking more about this, and here I'm probably not right:

> Practical: the change from this PR would simply not work in this scenario: the config map. (as in the test) is created prior the dependent resource is reconciled. From the PR the logic would recognize that the config map is there so won't create / update it. Now we delete the primary resource, since we did not create/update the resource the owner reference it not added, so the config map is not deleted. This is inconsistent with the current and desireg behavior, so simply wrong.

So basically update would not match and would add the owner reference.

So having this directly from cache could work as in this PR, but  it is hard to be backwards compatible with this feature - since we using getSecondaryResource not only "read the resource" but read "related resources". 

Maybe this we could discuss further also on the community meeting.
  
 ( TBH the create / update disctinguising does not make that much sense in the context of SSA, so there is bit logical issue also there. )

metacosm:
> just putting here for the record related stuff:
> 
> #2001 #1999 #1995
> 
> So basically this was discussed in context of KeyCloak operator, and we agreed on this approach.

I guess I didn't follow that discussion that closely then. I don't understand how the owner reference not being present would prevent the resource from being updated. The failing test in this PR shows that this isn't an issue: the ConfigMap that is manually created is properly updated by the dependent resource once JOSDK notices that it doesn't match the desired state.

Maybe we should check that there is not an owner reference with controller set to true and then fail the workflow in that case if an update is required because the reconciler cannot make an appropriate decision in that case: a resource exists that doesn't match the desired state but that is controlled by a different operator. This requires a human intervention to decide what needs to be done in that case, imo.

csviri:
Was thinking more about this, and actually, what we could do is do this way as you propose when that feature flag `createResourceOnlyIfNotExistingWithSSA` is set to false. And do as now (so with `getSecondaryResources` if this is true.)

metacosm:
Somewhat related: another example of the `getSecondaryResources` behavior being confusing: https://github.com/quarkiverse/quarkus-operator-sdk/issues/1161. There have been several such instances since the v5 update. In my opinion, the `createResourceOnlyIfNotExistingWithSSA` should be `false` by default.

csviri:
> Somewhat related: another example of the `getSecondaryResources` behavior being confusing: [quarkiverse/quarkus-operator-sdk#1161](https://github.com/quarkiverse/quarkus-operator-sdk/issues/1161). There have been several such instances since the v5 update. In my opinion, the `createResourceOnlyIfNotExistingWithSSA` should be `false` by default.

That is probably related to this https://github.com/operator-framework/java-operator-sdk/pull/2881 not v5

PR #2920, opened 2025-08-28
improve: [Dependent Resources] Set controller flag to true for owner reference

csviri:


xstefank (approved):

csviri:
I closed this, will re-open if a request comes from the users.

Issue #2895, opened 2025-08-11
[Question] `desired` method of `CRUDKubernetesDependentResource` called twice when dependent resource exists


michaelkoepf:
Hi,

I have an operator that uses a [standalone dependent resource](https://javaoperatorsdk.io/docs/documentation/dependent-resource-and-workflows/dependent-resources/#standalone-dependent-resources). The dependent resource is of type `CRUDKubernetesDependentResource<ConfigMap, T>` and only overrides method `protected ConfigMap desired(T primary, Context<T> context)`. When the dependent resource does not exist yet, `desired` is called once. However, when the dependent resource exists, `desired` is called twice in a row. Why is that? According to the [flowchart in the docs](https://javaoperatorsdk.io/docs/documentation/dependent-resource-and-workflows/dependent-resources/#motivations-and-goals), I would have expected only one call.

**Minimal reproducible example based on the [WebPage Example Operator](https://github.com/operator-framework/java-operator-sdk/tree/main/sample-operators/webpage)**

Full source code can be found here: https://github.com/michaelkoepf/dependent-resource-test. Most important classes can be found below.

```
@ControllerConfiguration
public class WebPageStandaloneDependentsReconciler implements Reconciler<WebPage> {

  private static final Logger LOG = LoggerFactory.getLogger(WebPageStandaloneDependentsReconciler.class);

  private final Workflow<WebPage> workflow;

  public WebPageStandaloneDependentsReconciler() {
    workflow = createDependentResourcesAndWorkflow();
  }

  @Override
  public List<EventSource<?, WebPage>> prepareEventSources(EventSourceContext<WebPage> context) {
    return EventSourceUtils.eventSourcesFromWorkflow(context, workflow);
  }

  @Override
  public UpdateControl<WebPage> reconcile(WebPage webPage, Context<WebPage> context)
      throws Exception {
    LOG.info("Starting reconciliation at {}", System.currentTimeMillis());
    workflow.reconcile(webPage, context);

    return UpdateControl.noUpdate();
  }

  @SuppressWarnings({"unchecked", "rawtypes"})
  private Workflow<WebPage> createDependentResourcesAndWorkflow() {
    // create the dependent resources
    var configMapDR = new ConfigMapDependentResource();

    // configure them with our label selector
    Arrays.asList(configMapDR)
        .forEach(
            dr ->
                dr.configureWith(
                    new KubernetesDependentResourceConfigBuilder()
[...]

PR #2922, opened 2025-08-28
feat: record desired state in Context

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
@@ -164,7 +164,7 @@ public class DefaultContext<P extends HasMetadata> implements Context<P> {
   }
 
   @SuppressWarnings("unchecked")
-  public <R> R desiredStateFor(
+  public <R> R getOrComputeDesiredStateFor(
       DependentResource<R, P> dependentResource, Function<P, R> desiredStateComputer) {
     return (R)
         desiredStates.computeIfAbsent(
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -86,7 +86,7 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
     if (creatable() || updatable()) {
       if (actualResource == null) {
         if (creatable) {
-          var desired = doDesired(context);
+          var desired = getOrComputeDesired(context);
           throwIfNull(desired, primary, "Desired");
           logForOperation("Creating", primary, desired);
           var createdResource = handleCreate(desired, primary, context);
@@ -96,7 +96,8 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
         if (updatable()) {
           final Matcher.Result<R> match = match(actualResource, primary, context);
           if (!match.matched()) {
-            final var desired = match.computedDesired().orElseGet(() -> doDesired(context));
+            final var desired =
+                match.computedDesired().orElseGet(() -> getOrComputeDesired(context));
             throwIfNull(desired, primary, "Desired");
             logForOperation("Updating", primary, desired);
             var updatedResource = handleUpdate(actualResource, desired, primary, context);
@@ -212,10 +213,25 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
             + " updated");
   }
 
-  protected R doDesired(Context<P> context) {
+  /**
+   * Retrieves the desired state from the {@link Context} if it has already been computed or calls
+   * {@link #desired(HasMetadata, Context)} and stores its result in the context for further use.
+   * This ensures that {@code desired} is only called once per reconciliation to avoid unneeded
+   * processing and supports scenarios where idempotent computation of the desired state is not
+   * feasible.
+   *
+   * <p>Note that this method should normally only be called by the SDK itself and exclusively (i.e.
+   * {@link #desired(HasMetadata, Context)} should not be called directly by the SDK) whenever the
+   * desired state is needed to ensure it is properly cached for the current reconciliation.
+   *
+   * @param context the {@link Context} in scope for the current reconciliation
+   * @return the desired state associated with this dependent resource based on the currently
+   *     in-scope primary resource as found in the context
+   */
+  protected R getOrComputeDesired(Context<P> context) {
     assert context instanceof DefaultContext<P>;
     DefaultContext<P> defaultContext = (DefaultContext<P>) context;
-    return defaultContext.desiredStateFor(this, p -> desired(p, defaultContext));
+    return defaultContext.getOrComputeDesiredStateFor(this, p -> desired(p, defaultContext));
   }
 
   public void delete(P primary, Context<P> context) {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
@@ -105,7 +105,7 @@ public abstract class AbstractExternalDependentResource<
 
   @Override
   public Matcher.Result<R> match(R resource, P primary, Context<P> context) {
-    var desired = doDesired(context);
+    var desired = getOrComputeDesired(context);
     return Matcher.Result.computed(resource.equals(desired), desired);
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
@@ -138,7 +138,7 @@ public class GenericKubernetesResourceMatcher<R extends HasMetadata, P extends H
       Context<P> context,
       boolean labelsAndAnnotationsEquality,
       String... ignorePaths) {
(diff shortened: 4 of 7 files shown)


csviri:
I think we should agree that on these optimization issues we should see the numbers, so we can decide if it is worth it.
First of all on large resources this will have a huge memory overhead. 
So we need to see, how faster is it (if it is faster at all), and how much more memory we consume.

But as we I think discussed before, there should be no blocking operation in desired state computation, that goes agains the pattern, calculating the desired (thus creating a pojo) is extremely fast - might be faster than getting a resource from a hash map (we need to measure), we should think though what we are actually optimizing with this.

metacosm:
Agreed but it might also be a good idea to look at the different cache layers which are becoming too complex to properly maintain.

csviri:
Note that without this:
https://github.com/operator-framework/java-operator-sdk/issues/2893
if the reconciler does not implement cleaner, there is no place where to clean those caches, since we don't receive the delete event now, when it should happen.

metacosm:
Thinking about this some more, the memory usage should actually be minimal and there is no need to clean things because this isn't a new cache per se as the desired states are put in the `Context` object which doesn't survive past the reconciliation so these are short-lived versions of objects that are already computed so it shouldn't incur any additional memory than what is already used during the reconciliation.

csviri:
> Thinking about this some more, the memory usage should actually be minimal and there is no need to clean things because this isn't a new cache per se as the desired states are put in the `Context` object which doesn't survive past the reconciliation so these are short-lived versions of objects that are already computed so it shouldn't incur any additional memory than what is already used during the reconciliation.

if it does not survive the reconiliation, why to cache it?
Now it does no use the context by a HashMap

metacosm:
> > Thinking about this some more, the memory usage should actually be minimal and there is no need to clean things because this isn't a new cache per se as the desired states are put in the `Context` object which doesn't survive past the reconciliation so these are short-lived versions of objects that are already computed so it shouldn't incur any additional memory than what is already used during the reconciliation.
> 
> if it does not survive the reconiliation, why to cache it? 
To avoid recomputing a potentially costly and non-idempotent desired state, which is something that people have run into several times already.

csviri:
> To avoid recomputing a potentially costly and non-idempotent desired state, which is something that people have run into several times already.

In what use case is the desired state cumputation costy?
Could you pls link the issue here, so we have better clarity? 
All that information should come from EventSource caches.

csviri:
TBH I don't like the fact that we are calling the desired either - and by definition idempotent function. But I'm not convinced that this is the way to solving or should be solved in generic way, users are able to optimize that in some edge cases already. But this is not a strong opinion, so if we are able to solve this elegantly withing the Context I'm open for that.

also maybe there is something I just don't see, so happy to dicuss further. Would be nice to see the users specific issue.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
```diff
@@ -24,6 +28,7 @@ public class DefaultContext<P extends HasMetadata> implements Context<P> {
   private final ControllerConfiguration<P> controllerConfiguration;
   private final DefaultManagedWorkflowAndDependentResourceContext<P>
       defaultManagedDependentResourceContext;
+  private final Map<DependentResource<?, P>, Object> desiredStates = new ConcurrentHashMap<>();
```
csviri, line 48: This probably should be removed.

metacosm:
> > To avoid recomputing a potentially costly and non-idempotent desired state, which is something that people have run into several times already.
> 
> In what use case is the desired state cumputation costy? Could you pls link the issue here, so we have better clarity? All that information should come from EventSource caches.

I don't have specific issues on hand but it's not unexpected that the desired state could be costly to compute if it requires accessing external systems. There's also the expectation since the beginning of the dependent resource feature that the desired method is only called once during the reconciliation process as it was the case until recently. Here are some issues related to this problem:
https://github.com/operator-framework/java-operator-sdk/issues/2895
https://github.com/operator-framework/java-operator-sdk/issues/1758

csviri:
> > > To avoid recomputing a potentially costly and non-idempotent desired state, which is something that people have run into several times already.
> > 
> > 
> > In what use case is the desired state cumputation costy? Could you pls link the issue here, so we have better clarity? All that information should come from EventSource caches.
> 
> I don't have specific issues on hand but it's not unexpected that the desired state could be costly to compute if it requires accessing external systems. There's also the expectation since the beginning of the dependent resource feature that the desired method is only called once during the reconciliation process as it was the case until recently. Here are some issues related to this problem: #2895 #1758

makes sense, thank you! yeah if this can be solved elegantly with such caches, might be a nice improvement

metacosm:
Can we merge this, @csviri?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
csviri, line 48: Why are we leaking this into an explicit construct? 
Shall this be part rather of `ManagedWorkflowAndDependentResourceContext` ?

csviri: Could we reuse only the already existing map there:
https://github.com/operator-framework/java-operator-sdk/blob/0dee1bf93f84acb959e5b77acf5c56d66599a0e1/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/ManagedWorkflowAndDependentResourceContext.java#L48

csviri:
> Can we merge this, @csviri?

added ine comment, sorry I though for some reason this is WIP

csviri:
I would also target `next` since this is a feature not a bugfix.

csviri:
Could we also add some integration test, or which test should cover this functionality?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
metacosm: > Why are we leaking this into an explicit construct? Shall this be part rather of `ManagedWorkflowAndDependentResourceContext` ?

It could indeed make sense there even though this isn't specific to managed dependents.

metacosm: > Could we reuse only the already existing map there:
> 
> https://github.com/operator-framework/java-operator-sdk/blob/0dee1bf93f84acb959e5b77acf5c56d66599a0e1/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/dependent/managed/ManagedWorkflowAndDependentResourceContext.java#L48

Sure but that'd be slightly more complex to handle and less type-safe.

metacosm:
> Could we also add some integration test, or which test should cover this functionality?

I'll look into it but basically the test should be that `desired` is only called once per reconciliation.

csviri:
> > Could we also add some integration test, or which test should cover this functionality?
> 
> I'll look into it but basically the test should be that `desired` is only called once per reconciliation.

yeah, if you think unit tests are enough, it is fine by me also. thx!

metacosm:
> I would also target `next` since this is a feature not a bugfix.

I disagree: this is a bug fix because the documentation states, and it was the case in v4, that `desired` should only be called once per reconciliation.

csviri:
> > I would also target `next` since this is a feature not a bugfix.
> 
> I disagree: this is a bug fix because the documentation states, and it was the case in v4, that `desired` should only be called once per reconciliation.

I don't see where does it states in the documentation. But it was obvious from the changes for v5 that we can call this multiple times. So it is then more of a documentation problem. Note that that was a major version change, so changes is behavior are fine, and we released also a minor version since then, therefore I believe would be a nicer to have this change in the behavior in minor verison update, even if such behavior existed before.

metacosm:
> > > I would also target `next` since this is a feature not a bugfix.
> > 
> > 
> > I disagree: this is a bug fix because the documentation states, and it was the case in v4, that `desired` should only be called once per reconciliation.
> 
> I don't see where does it states in the documentation. 

I guess it's not stated directly but the state diagram implies that the `desired` state is only computed once. 

> But it was obvious from the changes for v5 that we can call this multiple times. 

Where? That's definitely not true as users keep getting confused by the fact that it can be called multiple times.

csviri:
> Where? That's definitely not true as users keep getting confused by the fact that it can be called multiple times.

Which users? Could you please link the related issue here?

> I guess it's not stated directly but the state diagram implies that the desired state is only computed once.

> But it was obvious from the changes for v5 that we can call this multiple times.

And don't see why that diagram implies that, but mainly we explicitly state that in the docs that what users should do when to want to avoid calling desired multiple times:

https://javaoperatorsdk.io/docs/documentation/dependent-resource-and-workflows/dependent-resources/#multiple-dependent-resources-of-same-type

There is nothing wrong calling that multiple times, since it is inherently idempotent, and as we dicussed before should not be a blocking operation, we might want to also articulate in the docs. So I see this more of problem from user side we should point him to the docs, but we could also improve the documentation and/or also write a guide about it. We can add it to the FAQ.

I also created to PR that improves the docs by adding pointers and adding similar capability that we have for KubernetesDependetResources for ExternalDependentResources, that is a good point that we should have it also there.
https://github.com/operator-framework/java-operator-sdk/pull/3000


But in any means since there were further improvements around this please target next branch.

csviri:
But I see this as a good improvement, we can get rid of that mechanism, pls target next and also update the documentation.

csviri:
@metacosm could we target `next` branch with this feature please.

metacosm:
> @metacosm could we target `next` branch with this feature please.

Why? It didn't get into the last release so it's now targeting what was `next` at the time.

csviri:
> > @metacosm could we target `next` branch with this feature please.
> 
> Why? It didn't get into the last release so it's now targeting what was `next` at the time.

We usually release features in minor releases; the ongoing development for that is on `next`, `main` is reserved for bug fixes thus patch releases

metacosm:
Replaced by #3082

PR #2934, opened 2025-09-07
improve: GenericRetry does not provide mutable singleton instance

csviri:
This might easily result in unintended side effect if someone uses `DEFAULT` instance in one controller, changes some paramter; then uses the instance for an other controller, in that case the changed parameter will apply also for other controller what is quite unintuitive


csviri:
added also issue for v6 to refactor:
https://github.com/operator-framework/java-operator-sdk/issues/2935

xstefank (approved):

PR #2946, opened 2025-09-16
docs: fix reconiliation termination override snippet

csviri:


csviri:
> I would add spaces for readability but not pressing.

true, will do thx

PR #2947, opened 2025-09-17
feat: CI for java 25

csviri:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Set up Java and Maven
         uses: actions/setup-java@v5
         with:
-          distribution: zulu
+          distribution: temurin
           java-version: ${{ matrix.java }}
       - name: Run Special Integration Tests
         run: ./mvnw ${MAVEN_ARGS} -B package -P minimal-watch-timeout-dependent-it --file pom.xml
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -43,8 +43,8 @@ jobs:
       - name: Set up Java and Maven
         uses: actions/setup-java@v5
         with:
-          java-version: 17
-          distribution: zulu
+          java-version: 25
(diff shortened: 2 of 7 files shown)


metacosm (changes requested):
The commit message should be changed as this isn't an actual support of Java 25, just running CI on Java 25.

metacosm:
Also, at this point, I don't think that there's a Temurin Java 25 distribution available.

csviri:
> The commit message should be changed as this isn't an actual support of Java 25, just running CI on Java 25.

It is not the same thing? So basically the SDK works with java 25

csviri:
> Also, at this point, I don't think that there's a Temurin Java 25 distribution available.

No, will wait for that before this is merged.

metacosm:
> > The commit message should be changed as this isn't an actual support of Java 25, just running CI on Java 25.
> 
> It is not the same thing? So basically the SDK works with java 25

Well, not quite. We're just checking that it works with Java 25, not the same thing as actually doing something to support it. For that matter, have you actually checked that the SDK builds on Java 25?

csviri:
> Well, not quite. We're just checking that it works with Java 25, not the same thing as actually doing something to support it.

What I mean is this what we do for any java >17. But changed the title of the PR

metacosm:
For info, I've tried building with Java 25 and it appears the spotless plugin doesn't work on Java 25.

csviri:
fixed

metacosm:
Cool! We should still wait for the Temurin releases.

metacosm:
Ah, I thought that the Temurin releases were available now, according to https://adoptium.net/temurin/releases/. Maybe it takes some time to propagate images?

csviri:
> Ah, I thought that the Temurin releases were available now, according to https://adoptium.net/temurin/releases/. Maybe it takes some time to propagate images?

There is a github issue somehwere where they are tracking the exact progress, just can't find it anymore, but last week there was still a lot to done there

PR #2950, opened 2025-09-17
refactor: avoid unneeded initializations in injection contexts

metacosm:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
@@ -24,17 +24,17 @@ import io.javaoperatorsdk.operator.processing.LifecycleAware;
 public class Operator implements LifecycleAware {
   private static final Logger log = LoggerFactory.getLogger(Operator.class);
 
-  private final ControllerManager controllerManager;
-  private final LeaderElectionManager leaderElectionManager;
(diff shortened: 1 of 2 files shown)


metacosm:
Can we merge this, @csviri?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
```diff
@@ -46,12 +46,23 @@ public Operator() {
    *     operator
    */
   public Operator(ConfigurationService configurationService) {
-    this.configurationService = configurationService;
+    init(configurationService, false);
+  }
+
+  private void init(ConfigurationService configurationService, boolean allowDeferredInit) {
+    if (configurationService == null) {
+      if (!allowDeferredInit) {
```
csviri, line 80: Can we add a unit test for this?

csviri (approved):
LGTM, if there is an elegant way pls add unit test so we don't break this in the future.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/Operator.java
metacosm: Added a test, though it's not great and additional documentation, along with a protected method for subclasses to set the configuration service after intialization, if needed.

csviri (approved):

PR #2951, opened 2025-09-18
improve: PrimaryToSecondayMapper test improvements

csviri:
This change showcases that changes in `Cluster` resource triggers reconiliation of Job resources.

### DIFF
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/primarytosecondary/Cluster.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/primarytosecondary/Cluster.java
@@ -9,4 +9,4 @@ import io.fabric8.kubernetes.model.annotation.Version;
 @Group("sample.javaoperatorsdk")
 @Version("v1")
 @ShortNames("clu")
-public class Cluster extends CustomResource<Void, Void> implements Namespaced {}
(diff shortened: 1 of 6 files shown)


csviri:
@metacosm @xstefank not sure how that integration test is readable for you / users, we might want to separate the IT for the two appraoch types.

csviri:
@metacosm @xstefank can I merge this?

metacosm:
> @metacosm @xstefank not sure how that integration test is readable for you / users, we might want to separate the IT for the two appraoch types.

I think that it'd be better indeed to split this to make things clearer.

csviri:
> > @metacosm @xstefank not sure how that integration test is readable for you / users, we might want to separate the IT for the two appraoch types.
> 
> I think that it'd be better indeed to split this to make things clearer.

can do it as a separate PR

xstefank:
@csviri just making sure if you are going to separate it? Or should I review this PR?

csviri:
> @csviri just making sure if you are going to separate it? Or should I review this PR?

Pls review, we can later separate it, but it is PR anyways already done, and show that the change on `Cluster` reconiliation works. 

Later we can break it down to two separate tests. One that uses the `PrimaryToSecondaryMapper` one that does not, but both of those cases we want to have this test case

PR #2954, opened 2025-09-21
feat(docs): add @Sample annotation for documentation

Choconaut:
Marks integration test classes so that markdown pages can be auto generated.

### DIFF
--- a/operator-annotations/pom.xml
+++ b/operator-annotations/pom.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.javaoperatorsdk</groupId>
+        <artifactId>java-operator-sdk</artifactId>
+        <version>5.1.3-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>operator-annotations</artifactId>
+
+    <properties>
+        <maven.compiler.source>22</maven.compiler.source>
+        <maven.compiler.target>22</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+</project>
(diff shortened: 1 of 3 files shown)


Choconaut:
I wasn't sure where to place @Sample, so I put it in a temporary module for now. I thought about putting it in sample-operators. Do you have any recommendations? I'm happy to move it.

csviri:
I think that is fine by now, since the implementation will be I guess an annotation processor.

operator-annotations/pom.xml
```diff
@@ -0,0 +1,20 @@
...
+    <parent>
+        <groupId>io.javaoperatorsdk</groupId>
+        <artifactId>java-operator-sdk</artifactId>
+        <version>5.1.3-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>operator-annotations</artifactId>
+
+    <properties>
+        <maven.compiler.source>22</maven.compiler.source>
```
metacosm: I'm not sure these properties are needed here since they should probably be inherited from the parent. Also, this code should work with Java 21.

metacosm (changes requested):

csviri:
Hi @Choconaut , looks good maybe what we could do is to use this now in `operator-framework` module and annotate one test with the annotation and see how the result looks.

Choconaut:
> Hi @Choconaut , looks good maybe what we could do is to use this now in `operator-framework` module and annotate one test with the annotation and see how the result looks.

Hi, I added the sample annotations to these 2 classes: RDMappingInTestExtensionIT.java, ChangeNamespaceIT.java in `operator-framework`. Let me know if there's anything I should change.

csviri:
> > Hi @Choconaut , looks good maybe what we could do is to use this now in `operator-framework` module and annotate one test with the annotation and see how the result looks.
> 
> Hi, I added the sample annotations to these 2 classes: RDMappingInTestExtensionIT.java, ChangeNamespaceIT.java in `operator-framework`. Let me know if there's anything I should change.

thank you will take a look next week

operator-framework/pom.xml
```diff
@@ -84,6 +84,12 @@
       <artifactId>kube-api-test-client-inject</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.javaoperatorsdk</groupId>
+      <artifactId>operator-annotations</artifactId>
+      <version>5.1.3-SNAPSHOT</version>
```
csviri, line 90: We should use `${project.version}` here.

csviri:
Hi @Choconaut pls rebase it on to main and fix the build, I also tried it to locally but does not seem to generate the file, is it working for you locally?

csviri:
Hi @Choconaut I'm sorry but wanted to make progress with this, and quickly implemented this today:

https://github.com/operator-framework/java-operator-sdk/pull/3052

feel free to make improvements on that later.

PR #2961, opened 2025-09-24
feat: allow DependentResourceNode creation override

xstefank:
We want to provide CDI beans in the `Condition` subclasses as requested in https://github.com/quarkiverse/quarkus-operator-sdk/issues/720.


csviri:
I'm fine with this but @metacosm is more competent to review it, let's wait for him too

PR #2965, opened 2025-09-25
fix: handle all exceptions to support Kotlin

dvob:
When using JOSDK with Kotlin I ran into the problem that the retry did not work when the create method of a DependentResource throws an Exception not inherited from RuntimeException.

The change switches the handling from RuntimeException to Exception so that it works with Kotlin where you can throw Exceptions from methods which don't declare that they throw.

I changed this in the NodeExecutor to address the problem I specifically ran into. Then I also grepped for other places where RuntimeException is catched and then changed it in PollingEventSource as well.

Maybe there are other places (e.g. OperatorException) which should be changed.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
@@ -111,8 +111,9 @@ abstract class AbstractWorkflowExecutor<P extends HasMetadata> {
     actualExecutions.put(dependentResourceNode, future);
   }
 
+  // Exception is required because of Kotlin
   protected synchronized void handleExceptionInExecutor(
-      DependentResourceNode<?, P> dependentResourceNode, RuntimeException e) {
+      DependentResourceNode<?, P> dependentResourceNode, Exception e) {
     createOrGetResultFor(dependentResourceNode).withError(e);
   }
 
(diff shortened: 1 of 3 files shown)


csviri:
Hi @dvob I'm not against this, we probably don't want to change `OperatorException`. So you were able to test this PR, with properly with kotlin?

csviri (changes requested):
I think that we should at least comment on those places that `Exception is required because of kotlin`

csviri:
also created a followup issue: https://github.com/operator-framework/java-operator-sdk/issues/2967

dvob:
Theoretically this is for all JVM-languages which do not have the concept of checked vs. unchecked exceptions and not just for Kotlin. But not sure if in practice any other language is used.

Yes, I tested this with Kotlin and it fixed the problem (reconcile of DependentResource which throws Exception now causes a retry)

I will go ahead and add the comments at the places where we catch Exception, ok?
Do you want me to put this into the same commit or add a second commit and then you can squash it on merging?

csviri:
> I will go ahead and add the comments at the places where we catch Exception, ok?

Yes, pls.

> Do you want me to put this into the same commit or add a second commit and then you can squash it on merging?

Does not matter; we squash commits.

Thank you!

dvob:
I added the comments and also fixed the typos in the commit message

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/AbstractWorkflowExecutor.java
```diff
@@ -112,7 +112,7 @@ protected void markAsExecuting(
   }
 
   protected synchronized void handleExceptionInExecutor(
-      DependentResourceNode<?, P> dependentResourceNode, RuntimeException e) {
+      DependentResourceNode<?, P> dependentResourceNode, Exception e) {
```
csviri, line 116: @dvob pls add also comment here

xstefank (approved):
I'm OK with this but shouldn't this be considered a breaking change? If someone indeed is now throwing RuntimeException it will stop working?

csviri:
> I'm OK with this but shouldn't this be considered a breaking change? If someone indeed is now throwing RuntimeException it will stop working?

I don't see how it would stop working, could you pls elaborate?

xstefank:
scratch that, forgot which exception extends which exception

PR #2968, opened 2025-09-25
Use eventsource for secondary resource by default

florianknecht:
AbstractEventSourceHolderDependentResource may have the current state of the dependent resource in its event sources cache. Let's use that by default.

Otherwise the super implementation in AbstractDependentResource runs empty via context.getSecondaryResources, never looking for information in the event source's cache. Which forces all inheritors to implement this lookup themselves (e.g. PerResourcePollingExternalDependentResource, where I noticed this).


csviri:
This is a bot problematic, since the target event source could still hold multiple resources for this type.
This if indeed better for the case where are multiple event sources for the same type.
But would need to add also this check:
https://github.com/operator-framework/java-operator-sdk/blob/7397be67bd28314bd56b696b3f53b372955c4c53/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java#L140

csviri:
Note that the current design check on `equals` what is not ideal, since it is easy to miss that users need to override that, so in future we might have a more explicit solution for that.

florianknecht:
Got it - digging a little deeper with your hint made me understand how it's supposed to work. Thanks a lot!
I think I found something else though - will add a new PR for discussion.

metacosm:
> Note that the current design check on `equals` what is not ideal, since it is easy to miss that users need to override that, so in future we might have a more explicit solution for that.

The current design is wrong because it will never properly match resources that need to be updated even if `equals` is overridden (because in this case, `equals` will return `false` anyway).

florianknecht:
My issues here were caused by my Kotlin runtime not enforcing checked exceptions, see  https://github.com/operator-framework/java-operator-sdk/pull/2965 

@metacosm you're right, while a secondary resource is not reconciled it will not be returned by this method. Might be intentional though - I've not yet gotten deep enough into this. On the plus side, Kotlin DataClasses already implement a clean equals/hashcode for my external secondary resource type so I got that covered at least.

csviri:
> The current design is wrong because it will never properly match resources that need to be updated even if equals is overridden (because in this case, equals will return false anyway).

I checked this now on mysql sample, and it finds the resource, I see no issue there. Could you elaborate pls?

see:

image

csviri:
@florianknecht 
note that if `equals` does not work for you, you are always free to override `selectTargetSecondaryResource`:

https://github.com/operator-framework/java-operator-sdk/blob/9158b8e8e62b4670d37cb344515862cdc71e1eba/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java#L137

csviri:
@florianknecht 
create a draft with the rough idea I had on mind to improve this:
If this provides a nicer API, pls let us know, maybe we can continue discussion there.

csviri:
But the current approach is also functional, I don't see it broken. The API might be more elegant, though.

metacosm:
> > The current design is wrong because it will never properly match resources that need to be updated even if equals is overridden (because in this case, equals will return false anyway).
> 
> I checked this now on mysql sample, and it finds the resource, I see no issue there. Could you elaborate pls?
> 
> see:
> image

Have you tested with a scenario where the dependent resource needs to be updated (i.e. the current state doesn't match the desired state)?

csviri:
> > > The current design is wrong because it will never properly match resources that need to be updated even if equals is overridden (because in this case, equals will return false anyway).
> > 
> > 
> > I checked this now on mysql sample, and it finds the resource, I see no issue there. Could you elaborate pls?
> > see:
> > image
> 
> Have you tested with a scenario where the dependent resource needs to be updated (i.e. the current state doesn't match the desired state)?

That dependent resources is **NOT** an `Updater` so by definition not doing updates,see:
https://github.com/operator-framework/java-operator-sdk/blob/75405bf9d21cc50175e817dd253d5478e6e6e66e/sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java#L47-L49

PR #2973, opened 2025-09-29
fix: patchResourceAndStatus on status patch should generate diff only from status

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -486,7 +488,11 @@ private R editStatus(R resource, R originalResource) {
         clonedOriginal.getMetadata().setResourceVersion(null);
         resource.getMetadata().setResourceVersion(null);
         var res = resource(clonedOriginal);
-        return res.editStatus(r -> resource);
+        return res.editStatus(
+            r -> {
+              ((CustomResource) r).setStatus(((CustomResource) resource).getStatus());
```
metacosm: I don't think that this is appropriate as `R` is not always a `CustomResource`, if I'm not mistaken.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
csviri: yes working on it, it could be also built in resource. But basically can be only those 2. If someone calls it for example on `ConfigMap`, it is an issue on user side.

csviri: done.

metacosm (approved):
Not a fan of using reflection here, as mentioned before. We probably need to revive the associated Fabric8 issue.

csviri:
> Not a fan of using reflection here, as mentioned before. We probably need to revive the associated Fabric8 issue.

Sma here, that would be much nicer.

metacosm:
For reference, the Fabric8 issue: https://github.com/fabric8io/kubernetes-client/issues/3586

PR #2980, opened 2025-10-05
improve: add license headers to source files

csviri:
Was checking on this, and since other CNCF projects, what I checked at the end has this header, adding here too. 

- Adds license headers to source files.
- automatically adds the license header on project build to the new source file
- checks for the missing header in the PRs

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Field.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright Java Operator SDK Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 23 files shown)


xstefank (approved):
I didn't go through all the files, but I assume the header is the same everywher.

csviri:
> I didn't go through all the files, but I assume the header is the same everywher.

yes, using the plugin to add it.

metacosm:
I would prefer we do that for the next minor release, rather than for the next patch version and only when we're ready to release (i.e. once main and next have been merged) so that we don't lose changes that might not be in both branches.

csviri:
> I would prefer we do that for the next minor release, rather than for the next patch version and only when we're ready to release (i.e. once main and next have been merged) so that we don't lose changes that might not be in both branches.

I'm fine with that, will change the PR.

csviri:
updated the PR so it targets next. (also generated headers for new source files)

Issue #2981, opened 2025-10-06
Patch status errors on conflict logged too highly


shawkins:
## Bug Report

We are seeing frequent error logs such as:

```
2025-10-01 14:57:19,052 ERROR [io.jav.ope.pro.eve.ReconciliationDispatcher] (ReconcilerExecutor-keycloakcontroller-62) updateErrorStatus failed for resource: a86407e2-7e87-43a7-9465-52104c92c8ba with version: 22934977 for error Failure executing: PATCH at: https://10.96.0.1:443/apis/k8s.keycloak.org/v2alpha1/namespaces/auth/keycloaks/keycloak/status?fieldManager=keycloakcontroller&force=true. Message: Operation cannot be fulfilled on keycloaks.k8s.keycloak.org "keycloak": the object has been modified; please apply your changes to the latest version and try again. Received status: Status(apiVersion=v1, code=409, details=StatusDetails(causes=[], group=k8s.keycloak.org, kind=keycloaks, name=keycloak, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Operation cannot be fulfilled on keycloaks.k8s.keycloak.org "keycloak": the object has been modified; please apply your changes to the latest version and try again, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Conflict, status=Failure, additionalProperties={}).: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: PATCH at: https://10.96.0.1:443/apis/k8s.keycloak.org/v2alpha1/namespaces/auth/keycloaks/keycloak/status?fieldManager=keycloakcontroller&force=true. Message: Operation cannot be fulfilled on keycloaks.k8s.keycloak.org "keycloak": the object has been modified; please apply your changes to the latest version and try again. Received status: Status(apiVersion=v1, code=409, details=StatusDetails(causes=[], group=k8s.keycloak.org, kind=keycloaks, name=keycloak, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=Operation cannot be fulfilled on keycloaks.k8s.keycloak.org "keycloak": the object has been modified; please apply your changes to the latest version and try again, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Conflict, status=Failure, additionalProperties={}).
[...]

PR #2983, opened 2025-10-09
fix: reducing log level of failure to patch in handleErrorStatusHandler

shawkins:
closes: #2981

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
@@ -1,10 +1,12 @@
 package io.javaoperatorsdk.operator.processing.event;
 
 import java.lang.reflect.InvocationTargetException;
+import java.net.HttpURLConnection;
 import java.util.function.Function;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.event.Level;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.KubernetesResourceList;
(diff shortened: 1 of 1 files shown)


csviri:
Thank you for the PR.
I see your point and I'm not against this, just some points to discuss / consider:

1. if sombody (most of the time) will have logger for the framework on info level won't see any log message regarding what happened. It will just retry but there will be no info about that the conflict happened whatsoever. So the reconiliation will "silently fail".  Maybe for this case just rather log a short log message without stack trace that this happened? thus:
like:
```java
log.info("Conflict while updating custom resource, but next reconiliation is imminent").
```
Would this help in your case?

2. Note that usually retry is always in place (well it is considered a bad practice if turned off), so in most the cases it will retry this error right after the conflict anyways (regardless if reconciliation is imminent (other event received meanwhile) or not). 

3.  another option I see is to have a dedicated logger, this where we log now the conflic we could have a dedicated logger:

```java
private static final Logger conflictLogger = LoggerFactory.getLogger(ReconciliationDispatcher.class.getCanonicalName()+".conflict");

// omittes code


// modified log in this PR:
conflictLogger.info(
                "updateErrorStatus failed for resource: {} with version: {} for error {}",
                getUID(resource),
                getVersion(resource),
                e.getMessage(),
                ex);
```
Now you can change the log level of this logger to "warn" in your config you won't see these log messages, however we will be backwards compatible. (but maybe this approach is overkill)


Please let me know what do you think about these.

shawkins:
1. Fair enough, I'll follow the same pattern from https://github.com/operator-framework/java-operator-sdk/pull/2748/files here. At worst you'll see two highly related info logs - but it should be understandable what is happening.

2. Agreed.

3. That does seem like overkill.

shawkins:
See if this looks right. There will be three different logging behaviors, which is very similar to the other processing.

If it looks like this will be retried and is based upon a conflict, then it's a INFO message and DEBUG log of the exception.
If it looks like this will be retried and is any other reason, then it's a WARN log.
Otherwise it's an ERROR log.

csviri:
LGTM, let's wait also for a while for @xstefank  @metacosm

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java
```diff
@@ -216,12 +218,35 @@ public boolean isLastAttempt() {
...
+        String failedMessage = "";
+        if (context.isNextReconciliationImminent()
+            || !(errorStatusUpdateControl.isNoRetry() || retryInfo.isLastAttempt())) {
+          if (code == HttpURLConnection.HTTP_CONFLICT
+              || (originalResource.getMetadata().getResourceVersion() != null && code == 422)) {
+            exceptionLevel = Level.DEBUG;
+            failedMessage = " due to conflict";
+            log.info(
+                "ErrorStatusUpdateControl.patchStatus of {} failed due to a conflict, but the next"
+                    + " reconiliation is imminent.",
```
metacosm, line 232:
```suggestion
                    + " reconciliation is imminent.",
```

Issue #2984, opened 2025-10-09
Remove the usage of TemporaryResourceCache.knownResourceVersions


shawkins:
**Is your feature request related to a problem? Please describe.**
Explicit tracking of the knownResourceVersions is not needed. Instead we should simply check the tombstones and version comparison logic.

csviri:
Thnak you @shawkins

Issue #2944, opened 2025-09-15
Comparable Resource Versions in Kubernetes


csviri:
In future (k8s v1.35) going to support officially comparable resources versions:
https://github.com/kubernetes/enhancements/pull/5505

This has implications on our implementations - we already support this with feature flag (by default turned off) in informer temporary caches. But we will be able to optimize better for example in `PrimaryUpdateAndCacheUtils` - that would no longer require to do the update through optimistic locking.

- [x] how to support non complient resources. Automatic detection vs explicit configuration. -> explicit per informer event source
- [x] which version to ship? (basically this is supported also now in Kubernetes so we might target 5.2+) -> 5.3
- [x] Identify all the improvements we can do based on this. (some PR-s already added) -> see sub tasks and PRs here
- [x] review if on event lockstriping can be relaxed more
- [x] concurrent modification support on event filtering

Notes:
- related issue in controller runtime: https://github.com/kubernetes-sigs/controller-runtime/issues/3320
- https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5647-stale-controller-handling

csviri:
Added for now into 6.0 but we can check on this as soon as officially available in Kubernetes

csviri:
see kep: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5504-comparable-resource-version

PR #2985, opened 2025-10-09
refactor: remove the use of knownResourceVersions

shawkins:
closes: #2984

After seeing that comparable version support is coming in https://github.com/operator-framework/java-operator-sdk/issues/2981#issuecomment-3384461777 I reviewed the current logic and decided that it could be simplified.

We don't need to explicitly track known versions as a check of whether it's an unknown later version will work just as well.

A reason not to do this is if locking is not used, and you want to react to an older version that may have contained competing changes. However that doesn't seem like a valid thing to do from an eventual consistency perspective.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -192,9 +192,6 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
   }
 
   private boolean canSkipEvent(R newObject, R oldObject, ResourceID resourceID) {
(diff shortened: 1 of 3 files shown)


csviri:
checking the KEP:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5504-comparable-resource-version#conformance-tests

THey are actually not changing anything, just adding utility and tests, so we are fine with this IMHO, I will also do some related PRs.

csviri:
@shawkins could you pls rather target `next` branch for next minor release.

shawkins:
> @shawkins could you pls rather target `next` branch for next minor release.

The changes here a valid in general - they do not depend on #2944

csviri:
> > @shawkins could you pls rather target `next` branch for next minor release.
> 
> The changes here a valid in general - they do not depend on #2944

ahh ok, soory will unlink

PR #2987, opened 2025-10-10
Event skipping and temp resource caching based on resourceVersion comparison

### LINKED ISSUE #2944: Comparable Resource Versions in Kubernetes

csviri:
At this point it is safe to say that we can compare resource versions, that is simplifies lots of our algortihms in JOSDK. This PR simplifies the dependent resource related event skipping and caching parts.

Will clearly describe this in release notes, along with the removed config properties.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -164,20 +159,12 @@ public synchronized void putResource(T newResource, String previousResourceVersi
...
-      if (parseResourceVersions
-          && Long.parseLong(newResource.getMetadata().getResourceVersion())
-              > Long.parseLong(cachedResource.getMetadata().getResourceVersion())) {
-        return true;
-      }
+      return Long.parseLong(newResource.getMetadata().getResourceVersion())
+          > Long.parseLong(cachedResource.getMetadata().getResourceVersion());
     } catch (NumberFormatException e) {
-      log.debug(
+      log.warn(
```
shawkins: Could this even be an error?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -209,7 +196,7 @@ private synchronized void onAddOrUpdate(
   private boolean canSkipEvent(R newObject, R oldObject, ResourceID resourceID) {
     var res = temporaryResourceCache.getResourceFromCache(resourceID);
     if (res.isEmpty()) {
-      return isEventKnownFromAnnotation(newObject, oldObject);
+      return false;
```
shawkins: This handling doesn't seem directly related to resourceVersion comparability. It's to handle the case where the event comes in before the create / update puts the result in the cache.

Do you want to not worry about this case now, or is there some other locking or other logic not in this pr handling this case?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: hmm, maybe even throw an exception at this point?

shawkins: Yes, the only problematic behavior could be poorly written mock servers that are providing invalid resourceVersions.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: yes, this alone is not enough, but we might simplify it anyways, and just react on event that our last updated resource event.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -148,45 +145,45 @@ public synchronized void putResource(T newResource, String previousResourceVersi
...
   public synchronized Optional<T> getResourceFromCache(ResourceID resourceID) {
     return Optional.ofNullable(cache.get(resourceID));
   }
+
+  public boolean isNewerThenKnownResource(T newResource, ResourceID resourceID) {
+    var resource = getResourceFromCache(resourceID);
+    if (resource.isPresent()) {
+      return isLaterResourceVersion(newResource, resource.get());
+    }
+    var latestUpdated = latestUpdatedVersion.get(resourceID);
```
shawkins, line 186: This strategy can also be used instead of tombstones. If we attempt to put something that already has been deleted, the delete event (and the latest in general) is guarenteed to have been based upon a resourceVersion that is newer.

csviri: if you have the time pls add a pr for that.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
shawkins: Can you explain this a bit more? I'm not sure at the point in time if we have enough information to know if the incoming event is from our reconciliation, or another actor.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: > if you have the time pls add a pr for that.

Since that would be based upon the latest tracking, could you introduce / merge that (which will effectively shadow the latest tracked by the informer, but I'm not sure if you have anyone trying to use that directly) without the changes to the previous annotation? Or do you think you can complete that work in the near term?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: This is an outdated version but the current one: 
this implementation assumes that we want to trigger reconiliation only on newer version of resources, that came after our update. So basically we store the latest updated version even if we don't cache the resource in the temporal cache in case resource versions are equal. So new resources can be compared with this one. Essentially dropping events for our update and the updates happened before.

shawkins: To clarify here's the timing sequence that the code is guarding against:

- The reconciliation starts to do an update of a resource at v1.
- Before the kubernetes client receives that result, the informer receives the event for the transition from v1 -> v2
- That event is later than the lastest known event to the TemporaryResourceCache because it has yet to transition to v2, so canSkipEvent will return false.

csviri: ahh, yes, thank you pointing this out, might separate the PR and see if It is possible to handle that without the annotation.

shawkins: As far as I remember the only two approaches are locking or the annotation. Locking is more intrusive in other parts of the code, whereas the current annotation approach is prone to reconciliation loops if SSA change detection isn't perfect. We could probably come up with a way to reason over a stable hash, but that starts seeming expensive for this problem.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: Actually I think this can just be something like https://github.com/operator-framework/java-operator-sdk/pull/2989

PR #2988, opened 2025-10-10
Resource version comparison utility

csviri:
KEP compatible implementation of resourceVersion comparator: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5504-comparable-resource-version#helper-function

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
@@ -458,7 +458,7 @@ public class PrimaryUpdateAndCacheUtils {
     }
     var v2Length = v2.length();
     if (v2Length == 0) {
-      throw new NonComparableResourceVersionException("Resource version (1) is empty");
+      throw new NonComparableResourceVersionException("Resource version (2) is empty");
     }
     var maxLength = Math.max(v1Length, v2Length);
     boolean v1LeadingZero = true;


shawkins:
Should this include the numeric sanity checks as well?

csviri:
> Should this include the numeric sanity checks as well?

added, it is not perfect this way, but I would say good enough, what do you think?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -450,4 +450,33 @@ public static <P extends HasMetadata> P addFinalizerWithSSA(
...
+      }
+
+      if (char1 > char2) {
+        return 1;
+      }
+      if (char1 < char2) {
+        return -1;
+      }
+    }
+    return 0;
```
shawkins: > Loop through both strings, ensuring they are at least one character long, have no leading 0's, and all characters are 0-9, and return an error otherwise

To fully validate using a single iteration would look like:

```suggestion
    var v1Length = v1.length();
    if (v1Length == 0) {
      throw new IllegalStateException("Resource version (1) is empty");
    }
    var v2Length = v2.length();
    if (v2Length == 0) {
      throw new IllegalStateException("Resource version (2) is empty");
    }
    var maxLength = Math.max(v1Length, v2Length);
    boolean v1LeadingZero = true;
    boolean v2LeadingZero = true;
    int lengthComparison = 0;
    int comparison = 0;
    for (int i = 0; i < maxLength; i++) {
      char char1 = 0;
      if (i < v1Length) {
        char1 = v1.charAt(i);
        if (v1LeadingZero) {
          if (char1 == '0') {
            throw new IllegalStateException("Resource version (1) cannot begin with 0");
          }
          v1LeadingZero = false;
        }
        if (!Character.isDigit(char1)) {
          throw new IllegalStateException(
              "Non numeric characters in resource version (1): " + char1);
        }
      }
      if (i < v2Length) {
        var char2 = v2.charAt(i);
        if (v2LeadingZero) {
          if (char2 == '0') {
            throw new IllegalStateException("Resource version (1) cannot begin with 0");
          }
          v2LeadingZero = false;
        }
        if (!Character.isDigit(char2)) {
          throw new IllegalStateException(
              "Non numeric characters in resource version (2): " + char2);
        }
        if (char1 == 0) {
          lengthComparison = -1;
        } else if (comparison == 0) {
          comparison = Character.compare(char1, char2);
        }
      } else {
        lengthComparison = 1;
      }
    }
    return lengthComparison != 0 ? lengthComparison : comparison;
```

csviri: why would we check for leading zero?

shawkins: Otherwise "01" would be greater than "2".

csviri: Yes, but is than realistic in any means?

shawkins: I'm not sure what you mean by realistic - do we expect that to be returned by a conforming implementation, no. Would it proactively catch any invalid scenario, yes. 

Encapsulated in this method we can do:

1. No sanity checks - the initial state of the pr
2. Partial sanity checks - the second state of the pr
3. Full sanity checks - this suggested change

Of these 1 and 3 make the most sense to me.

The alternative to 3 is put the resourceVersion validation in a different spot, that will only get run once per event, and could even be made optional.

csviri: kk, I'm convinced, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -450,4 +450,57 @@ public static <P extends HasMetadata> P addFinalizerWithSSA(
...
+        if (!Character.isDigit(char1)) {
+          throw new IllegalStateException(
+              "Non numeric characters in resource version (1): " + char1);
+        }
+      }
+      if (i < v2Length) {
+        var char2 = v2.charAt(i);
+        if (v2LeadingZero) {
+          if (char2 == '0') {
+            throw new IllegalStateException("Resource version (1) cannot begin with 0");
```
shawkins:
```suggestion
            throw new IllegalStateException("Resource version (2) cannot begin with 0");
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -450,4 +450,57 @@ public static <P extends HasMetadata> P addFinalizerWithSSA(
...
+        if (char1 == 0) {
+          lengthComparison = -1;
+        } else if (comparison == 0) {
+          comparison = Character.compare(char1, char2);
+        }
+      } else {
+        lengthComparison = 1;
+      }
+    }
+    return lengthComparison != 0 ? lengthComparison : comparison;
```
shawkins: I realized some other changes made it so lengthComparison is not needed:

```suggestion
    var v1Length = v1.length();
    if (v1Length == 0) {
      throw new IllegalStateException("Resource version (1) is empty");
    }
    var v2Length = v2.length();
    if (v2Length == 0) {
      throw new IllegalStateException("Resource version (2) is empty");
    }
    var maxLength = Math.max(v1Length, v2Length);
    boolean v1LeadingZero = true;
    boolean v2LeadingZero = true;
    int comparison = 0;
    for (int i = 0; i < maxLength; i++) {
      char char1 = 0;
      if (i < v1Length) {
        char1 = v1.charAt(i);
        if (v1LeadingZero) {
          if (char1 == '0') {
            throw new IllegalStateException("Resource version (1) cannot begin with 0");
          }
          v1LeadingZero = false;
        }
        if (!Character.isDigit(char1)) {
          throw new IllegalStateException(
              "Non numeric characters in resource version (1): " + char1);
        }
      }
      if (i < v2Length) {
        var char2 = v2.charAt(i);
        if (v2LeadingZero) {
          if (char2 == '0') {
            throw new IllegalStateException("Resource version (2) cannot begin with 0");
          }
          v2LeadingZero = false;
        }
        if (!Character.isDigit(char2)) {
          throw new IllegalStateException(
              "Non numeric characters in resource version (2): " + char2);
        }
        if (char1 == 0) {
          comparison = -1;
        } else if (comparison == 0) {
          comparison = Character.compare(char1, char2);
        }
      } else {
        comparison = 1;
      }
    }
    return comparison;
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -450,4 +450,56 @@ public static <P extends HasMetadata> P addFinalizerWithSSA(
...
+  public static int compareResourceVersions(String v1, String v2) {
+    var v1Length = v1.length();
+    if (v1Length == 0) {
+      throw new IllegalStateException("Resource version (1) is empty");
+    }
+    var v2Length = v2.length();
+    if (v2Length == 0) {
+      throw new IllegalStateException("Resource version (2) is empty");
+    }
+    var maxLength = Math.max(v1Length, v2Length);
```
csviri, line 463: Assuming we don't have leading zeros, would it make sense to compare the lenghts upfront?
In general there would be not a bug win for that, but might simplify things.  
I'm pretty sure etcd does not generate versions with leading zeros

@shawkins

csviri: ( I mean I could leave with that assumption :) )

shawkins: > Assuming we don't have leading zeros, would it make sense to compare the lenghts upfront?

That also short circuits the check for non-digit characters. So if you want to go with that approach, then just revert to the initial state of the pr - and consider putting the resource version validation elsewhere if you want a sanity check.

csviri: For now reverted for sake of simplicity / readability, but we can discuss on next community meeting, and maybe decide with others which one to take.

csviri: But added the validation you proposed

shawkins: It's probably confusing to have the other sanity checks in there - I'd vote to either fully validate or not at all.

I would add that the performance difference between fully validating and partially validating should be minimal in the long run. As the api server is up longer, the larger resource versions will get, and the less likely it will be for there to be a difference in the length.

csviri: Yeah, at the end I removed the checks, I think we should not test the Kubernetes contract, if the contract say it is an  integer as string, we are safe to assume that.

csviri: but yeah, another way is what you provided, is exatly what is described in kep:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5504-comparable-resource-version#helper-function

But let's talk about the guaratees we want to give. Basically if we want to provide a feature flag (as now) where we support non-integer resourceVerions or not.

However reading this section it might be different on resource bases:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5504-comparable-resource-version#risks-and-mitigations

csviri: so at the end reverted it to your version :)

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtilsTest.java
```diff
@@ -176,4 +181,43 @@ void cachePollTimeouts() {
...
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("11", "ba"));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("", "22"));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("11", ""));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("01", "123"));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("123", "01"));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("3213", "123a"));
+    assertThrows(IllegalArgumentException.class, () -> compareResourceVersions("321", "123a"));
+  }
+
+  // naive performance that compares the works case scenario for non parsing variant
```
xstefank:
```suggestion
  // naive performance test that compares the work case scenario for the parsing and non-parsing variants
```

csviri: thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
```diff
@@ -450,4 +450,58 @@ public static <P extends HasMetadata> P addFinalizerWithSSA(
...
   }
+
+  public static int compareResourceVersions(String v1, String v2) {
+    var v1Length = v1.length();
+    if (v1Length == 0) {
+      throw new NonComparableResourceVersionException("Resource version (1) is empty");
+    }
+    var v2Length = v2.length();
+    if (v2Length == 0) {
+      throw new NonComparableResourceVersionException("Resource version (1) is empty");
```
xstefank:
```suggestion
      throw new NonComparableResourceVersionException("Resource version (2) is empty");
```
?

xstefank: or just merge it into one check and one exception

xstefank (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
csviri: committed, thank you!

csviri:
@metacosm can we merge this?

PR #2989, opened 2025-10-10
removing tombstones

shawkins:
@csviri I misread your working draft at some point and thought you were tracking a single latest revision - that's all I need to remove tombstone tracking. If we are not relying on comparable resource versions, then I think we can simply reuse the informer last synced version to reason about all puts - not just creates.

I've left the latest check as non-compiling as that will line up with changes you have in other prs.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -34,6 +34,7 @@ import io.javaoperatorsdk.operator.OperatorException;
 import io.javaoperatorsdk.operator.api.config.ControllerConfiguration;
 import io.javaoperatorsdk.operator.api.config.Informable;
 import io.javaoperatorsdk.operator.api.config.NamespaceChangeable;
+import io.javaoperatorsdk.operator.api.reconciler.PrimaryUpdateAndCacheUtils;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.RecentOperationCacheFiller;
 import io.javaoperatorsdk.operator.health.InformerHealthIndicator;
 import io.javaoperatorsdk.operator.health.InformerWrappingEventSourceHealthIndicator;
@@ -122,30 +123,38 @@ public abstract class ManagedInformerEventSource<
   @Override
   public void handleRecentResourceUpdate(
       ResourceID resourceID, R resource, R previousVersionOfResource) {
-    temporaryResourceCache.putResource(
-        resource, previousVersionOfResource.getMetadata().getResourceVersion());
+    temporaryResourceCache.putResource(resource);
   }
 
   @Override
   public void handleRecentResourceCreate(ResourceID resourceID, R resource) {
-    temporaryResourceCache.putAddedResource(resource);
+    temporaryResourceCache.putResource(resource);
   }
 
   @Override
   public Optional<R> get(ResourceID resourceID) {
+    var res = cache.get(resourceID);
     Optional<R> resource = temporaryResourceCache.getResourceFromCache(resourceID);
-    if (resource.isPresent()) {
-      log.debug("Resource found in temporary cache for Resource ID: {}", resourceID);
+    if (parseResourceVersions
+        && resource.isPresent()
+        && res.filter(
+                r ->
+                    PrimaryUpdateAndCacheUtils.compareResourceVersions(r, resource.orElseThrow())
+                        > 0)
+            .isEmpty()) {
+      log.debug("Latest resource found in temporary cache for Resource ID: {}", resourceID);
       return resource;
-    } else {
-      log.debug(
-          "Resource not found in temporary cache reading it from informer cache,"
-              + " for Resource ID: {}",
-          resourceID);
-      var res = cache.get(resourceID);
-      log.debug("Resource found in cache: {} for id: {}", res.isPresent(), resourceID);
-      return res;
     }
+    log.debug(
+        "Resource not found, or older, in temporary cache. Found in informer cache {}, for"
+            + " Resource ID: {}",
+        res.isPresent(),
+        resourceID);
+    return res;
+  }
+
+  public Optional<String> getLastSyncResourceVersion(Optional<String> namespace) {
+    return cache.getLastSyncResourceVersion(namespace);
   }
 
   @SuppressWarnings("unused")
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
@@ -15,7 +15,6 @@
  */
 package io.javaoperatorsdk.operator.processing.event.source.informer;
 
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -24,7 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.javaoperatorsdk.operator.api.config.ConfigurationService;
+import io.javaoperatorsdk.operator.api.reconciler.PrimaryUpdateAndCacheUtils;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResource;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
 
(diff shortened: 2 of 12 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -131,17 +86,21 @@ public synchronized void putAddedResource(T newResource) {
    */
   public synchronized void putResource(T newResource, String previousResourceVersion) {
     var resourceId = ResourceID.fromResource(newResource);
+
     var cachedResource = managedInformerEventSource.get(resourceId).orElse(null);
 
     boolean moveAhead = false;
+    String latest = managedInformerEventSource.getLastSyncResourceVersion(resourceId.getNamespace()).orElse(null);
+    if (latest != null && latest > newResource.getMetadata().getResourceVersion()) {
```
csviri: Is it guaranteed that those events (maybe for different resource names) come in order in watches?

csviri: But I guess if we going to compare the resource versions in informers that does not matter.

shawkins: My understanding is that ordering is guarenteed. So this is applicable not only to getting rid of tombstones, but putting anything into the cache in general - added that as another simplification.

Also the javadoc on getLastSyncResourceVersion could use some improvement. Updates to the cache are made prior to async event processing. So the underlying cache state should be correct up to the given resourceVersion at the time we check getLastSyncResourceVersion.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -131,36 +86,20 @@ public synchronized void putAddedResource(T newResource) {
...
+          "Resource {}: resourceVersion {} is not later than latest {}",
+          resourceId,
+          newResource.getMetadata().getResourceVersion(),
+          latest);
+    } else {
       log.debug(
           "Temporarily moving ahead to target version {} for resource id: {}",
           newResource.getMetadata().getResourceVersion(),
           resourceId);
       cache.put(resourceId, newResource);
```
shawkins, line 126: This should be safe to do without further conditions as long as there is no chance that the operator sdk is making effectively concurrent updates to the same resource - if that's possible, then we still need to compare the cache item to what is being put.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -133,21 +133,25 @@ public void handleRecentResourceCreate(ResourceID resourceID, R resource) {
 
   @Override
   public Optional<R> get(ResourceID resourceID) {
+    var res = cache.get(resourceID);
     Optional<R> resource = temporaryResourceCache.getResourceFromCache(resourceID);
-    if (resource.isPresent()) {
+    if (resource.isPresent() && res.filter(r -> r.getMetadata().getResourceVersion() < resource.get().getMetadata().getResourceVersion()).isPresent()) {
```
shawkins: If events are processed async from the cache update, it's possible for the informer cache to be more up-to-date than the temporary resource cache.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: > My understanding is that ordering is guarenteed. So this is applicable not only to getting rid of tombstones, but putting anything into the cache in general - added that as another simplification.

I see, I know that it is true for a single resource, but for example there are multiple resources in the same type, if that still comes in order, but I guess yes, can also quickly take a look today on some cluster and watch some events.

Thank you, this is great!!

csviri: ( also if you could join us tomorrow on community meeting we could discuss strategy would be great, thinking if these improvements should go to 5.2 or rather the next 5.3 so we can iterate and discuss more, and there is no pressure, since might be good to do a minor release soonish )

shawkins: Is that on zoom at 15:00 CEST?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: Converted to a comment in the code

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: yes

csviri:
added branch for 5.3 (as we discuss on community meeting) to target resource version comparison related issues:
https://github.com/operator-framework/java-operator-sdk/pull/2996

thank you!

shawkins:
@csviri just wanted to double check - what if anything do you want in next? I was thinking that only the annotation removal would go into 5.3.

I have enabled parse versions by default, which does need some refinement - perhaps a better name and we lack a configuration value for on the primary resources. If it is not enabled, the the temporary resource cache won't do anything.

There's no smart exception handling yet. If non comparable versions are seen, then you'll just get an exception - the message could mention disabling the feature.

csviri:
> just wanted to double check - what if anything do you want in next? I was thinking that only the annotation removal would go into 5.3.

Although this will work, in almost all cases, since:

`There's no smart exception handling yet. If non comparable versions are seen, then you'll just get an exception - the message could mention disabling the feature.`

We might want to ship this as a whole in 5.3.
see also: https://github.com/operator-framework/java-operator-sdk/issues/3011

So we have also more time and refine it. We can then release 5.3 with those feature quite quickly don't have to wait for other feature. What do you think?

csviri:
Do I see that this PR is same as: https://github.com/operator-framework/java-operator-sdk/pull/3010

Should we close one of those?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -116,74 +71,68 @@ synchronized void onEvent(T resource, boolean unknownState) {
     cache.computeIfPresent(
         ResourceID.fromResource(resource),
         (id, cached) ->
-            (unknownState || !isLaterResourceVersion(id, cached, resource)) ? null : cached);
-  }
-
-  public synchronized void putAddedResource(T newResource) {
-    putResource(newResource, null);
+            (unknownState
+                    || PrimaryUpdateAndCacheUtils.compareResourceVersions(resource, cached) > 0)
```
shawkins: Answering https://github.com/operator-framework/java-operator-sdk/pull/3010#discussion_r244420367 here - yes we can switch to >= because we're checking canSkipEvent first. I was starting to think more about the implications of the annotation removal and wondering if we should hold on to the resource that was created until it was obsolete, but we'll discuss that more on the annotation removal pr.

shawkins:
> Do I see that this PR is same as: #3010
> 
> Should we close one of those?

Closed the other one for now - if enabling parse versions seems like it should be separated, then it will be reopened with all the other changes stripped out.

shawkins:
> Added just one comment, I lean towards 5.3 regarding this feature, and completely remove the flag for parsing fro that version. Otherwise LGTM

I don't think it can remove the flag if you want to allow for non-conforming sources. The configuration would also need expanded to the controller itself.

csviri:
> > Added just one comment, I lean towards 5.3 regarding this feature, and completely remove the flag for parsing fro that version. Otherwise LGTM
> 
> I don't think it can remove the flag if you want to allow for non-conforming sources. The configuration would also need expanded to the controller itself.

My idea is that we would detect non-conforming sources dynamically, so if the comparison algorithm throws an error we can flip a flag and don't compare the resources anymore.

csviri:
But for sake of simplicity we could configure it on InformerEventSources for now. (maybe as next PR)

shawkins:
> My idea is that we would detect non-conforming sources dynamically, so if the comparison algorithm throws an error we can flip a flag and don't compare the resources anymore.

That could still be two modes of operation:
- strict (the default)
- lenient 

Related to that in this pr there is: https://github.com/operator-framework/java-operator-sdk/pull/2989/files#diff-0b20da91b3c31881f7f78a0ecffefc3811e1916f8e401a047d335fffa8593b71R91-R97

Which is lenient, or it could be done as a breaking change an throw an exception.

csviri:
> * lenient

Yes, I think lenient is fine, but maybe this would require more discussions. (therefore proposing to move also this pr agains `5.3` branch, and to that version.

> Related to that in this pr there is: https://github.com/operator-framework/java-operator-sdk/pull/2989/files#diff-0b20da91b3c31881f7f78a0ecffefc3811e1916f8e401a047d335fffa8593b71R91-R97

I guess the null check is not enough rather a validation of that version, and catching the `NonComparableResourceVersionException` also at 

```java
 public synchronized void putAddedResource(T newResource) {
    putResource(newResource, null);
            (unknownState
                    || PrimaryUpdateAndCacheUtils.compareResourceVersions(resource, cached) >= 0)
                ? null
                : cached);
  }
```

But the configuration if want to have explicit, should not be on controller level, rather on InformerEventSource+ControllerEventSource level.

Otherwise this looks great!!

shawkins:
> Yes, I think lenient is fine, but maybe this would require more discussions. (therefore proposing to move also this pr agains 5.3 branch, and to that version.

Ok, I'll target to the 5.3 branch.

> I guess the null check is not enough rather a validation of that version

Null - seems like a simple programatic error, which was already present in the external example
Non-null, but invalid - seems like a non-conforming source, so it seems fine to allow the NonComparableResourceVersionException

>  and catching the NonComparableResourceVersionException also at

Doesn't it make more sense to start with enabled / disabled behavior, then add a lenient mode later if users want that instead of strictly disabled?

csviri:
> Doesn't it make more sense to start with enabled / disabled behavior, then add a lenient mode later if users want that instead of strictly disabled?

I'm fine with that, as far it is per InformerEventSource.

PR #2999, opened 2025-10-15
showing code to possibly remove the previous annotation

shawkins:
Pros:
- no additional locking
- could be adapted to not even rely on resourceVersion comparisons

Cons:
- additional state accumulation on ResourceState during processing. This can be narrowed, but that could end up looking like similar to a locking based solution.
- isNextReconciliationImminent is weakened - without further refinement we won't commit to dependent events triggering a next reconciliation until we're done with the current reconcilitation

Another thought - in what situations will dependents have multiple primary resources? We are skipping the reconciliation for all of them if canSkipEvent returns true from the temporary resource cache, but that add/update occurred for just a single primary resource.


csviri:
thank you will try to take a look soon

PR #3000, opened 2025-10-15
improve: possibility to provide id for external resource

csviri:
PR introduces similar mechanism for external resource as for KubernetesDependetResource to easily override a method to provide an ID if user want't to avoid computing desired states multiple times.
Also improves the docs with pointers where we explain how to avoid calling desired multiple times.

### DIFF
--- a/docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
+++ b/docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
@@ -306,10 +306,10 @@ it is always possible to override this automated discrimination using several me
 - Override the [`targetSecondaryResourceID`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L282)
   method, if your `DependentResource` extends `KubernetesDependentResource`,
   where it's very often possible to easily determine the `ResourceID` of the secondary resource. This would probably be
-  the easiest solution if you're working with Kubernetes resources. Similarly, for you can override the 
+  the easiest solution if you're working with Kubernetes resources. Similarly, you can override the 
   [`targetSecondaryResourceID`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java#L138)
   for external resources. See below the related ID handling
-- If the approach from above doesn't fit your needs, you can override the target resource selection mechanism by overriding
+- If the approach above doesn't fit your needs, you can override the target resource selection mechanism by overriding
   `selectTargetSecondaryResource` for both [`KubernetesDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L282)
   and [`AbstractExternalDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java#L148). 
 - As last resort, you can implement your own `getSecondaryResource` method on your `DependentResource` implementation from scratch.
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java
@@ -127,7 +127,7 @@ public abstract class AbstractExternalDependentResource<
   }
 
   /**
-   * Override this method to optimize and not compute the desired when selecting the target
+   * Override this method to optimize and avoid computing the desired when selecting the target
    * secondary resource. Return the target id.
    *
    * @see ExternalDependentIDProvider


docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
```diff
@@ -303,14 +303,15 @@ they should be reconciled.
...
+- Override the [`targetSecondaryResourceID`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L282)
+  method, if your `DependentResource` extends `KubernetesDependentResource`,
   where it's very often possible to easily determine the `ResourceID` of the secondary resource. This would probably be
-  the easiest solution if you're working with Kubernetes resources.
-- Override the `selectTargetSecondaryResource` method, if your `DependentResource` extends `AbstractDependentResource`.
-  This should be relatively simple to override this method to optimize the matching to your needs. You can see an
-  example of such an implementation in
-  the [`ExternalWithStateDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/externalstate/ExternalWithStateDependentResource.java)
-  class. 
+  the easiest solution if you're working with Kubernetes resources. Similarly, for you can override the
```
```suggestion
  the easiest solution if you're working with Kubernetes resources. Similarly, you can override the 
```

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
```diff
@@ -303,14 +303,15 @@ they should be reconciled.
...
-  example of such an implementation in
-  the [`ExternalWithStateDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework/src/test/java/io/javaoperatorsdk/operator/dependent/externalstate/ExternalWithStateDependentResource.java)
-  class. 
+  the easiest solution if you're working with Kubernetes resources. Similarly, for you can override the 
+  [`targetSecondaryResourceID`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java#L138)
+  for external resources. See below the related ID handling
+- If the approach from above doesn't fit your needs, you can override the target resource selection mechanism by overriding
+  `selectTargetSecondaryResource` for both [`KubernetesDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L282)
```
xstefank, line 313: https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java#L269

Noting also that we should propably look into how to automate this with some kind of annotation comments in code to track correct line numbers.

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
```diff
@@ -303,14 +303,15 @@ they should be reconciled.
...
+  and [`AbstractExternalDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java#L148).
```
xstefank, line 314:
```suggestion
  and [`AbstractExternalDependentResource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java). 
```

this class has only 109 lines.

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
csviri: yes, if we could solve that, would be awesome.

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
csviri: not after this PR :)

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
xstefank: @csviri but this link still needs to be fixed or not?

csviri:
@metacosm I would like to merge this, we can have the desired caching as a followup pr on `next`. thx

metacosm:
Reviewing this, I'm not sure I like the current design of putting the provider on the external resource class… I think it might be more appropriate to put it on the related dependent resource implementation since this is where the logic for this should be, imo. It doesn't feel right to put JOSDK-related logic on the external resource class, especially considering that maybe that class is not modifiable (if it's coming from a 3rd party library, for example). It also makes sense to gather the resource identification logic on the dependent resource implementation, imo.

csviri:
> Reviewing this, I'm not sure I like the current design of putting the provider on the external resource class… I think it might be more appropriate to put it on the related dependent resource implementation since this is where the logic for this should be, imo. It doesn't feel right to put JOSDK-related logic on the external resource class, especially considering that maybe that class is not modifiable (if it's coming from a 3rd party library, for example). It also makes sense to gather the resource identification logic on the dependent resource implementation, imo.

Yes, I was also thinking about that. Maybe we could generalize the `cacheKeyMapper` concept, see also generalization here:

https://github.com/operator-framework/java-operator-sdk/pull/3007

WDYT?

But those are not necessarily contradicting nor we have to choose, we can have the (renamed to ResourceIDMapper ) `cacheKeyMapper` in dependent resources, where there will be still a posibility to use the `ResourceIdProvider`, just not limited to that.

csviri:
So what  I would do is to merge this and https://github.com/operator-framework/java-operator-sdk/pull/3007  to next, but iterate over it for 5.2 to achieve what you also proposed.

xstefank:
I think Chris has a point. So why it would be worth merging this now to next and then changing it later?

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3007

@xstefank so I still intend to kepp the `ExternalResourceVersionProvider` just also use the `CacheKeyMapper`/`ResourceIDMapper` for DRs

let's continue the discussion under the linked PR, will prepare one with the mentioned changes

metacosm:
Should the branch be deleted?

PR #3001, opened 2025-10-16
wip

metacosm:


metacosm:
@csviri can you tell quickly what is wrong with that test? To me, it looks reasonable and on par with what a user would expect…

csviri:
> @csviri can you tell quickly what is wrong with that test? To me, it looks reasonable and on par with what a user would expect…

I don't see an obvious issue on first look, can take a look next week more in depth

PR #3002, opened 2025-10-16
[WIP] feat: generalizing id provider across dependent resource and external resource event sources

csviri:
For 5.2 we introduced id provider for external resources:

but we use such id mechanism across multiple components, this PR shows how we could we have such ID provider unified across components like the BulkDependentResources and PerResourcesEventSource and ExternalDependentResources that handles those. 

An other approach would could take is to have ExternalResourceIDProvider is to return string instead of generic type, what would be less general and a bit opionated, however the hierarchy would be much simpler.


csviri:
see: simpler version where the the bulk dr is not covered

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3020

metacosm:
Should this branch be deleted?

PR #3003, opened 2025-10-16
fix: flaxy EventProcessorTest.triggerOnAllEventDeleteEventInstantlyAfterEvent

csviri:


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/EventProcessorTest.java
```diff
@@ -562,7 +562,9 @@ void triggerOnAllEventDeleteEventInstantlyAfterEvent() {
     eventProcessor.handleEvent(prepareCRDeleteEvent1());
 
     waitUntilProcessingFinished(eventProcessor, TestUtils.testCustomResource1Id());
-    verify(reconciliationDispatcherMock, times(1)).handleExecution(any());
+    // it es expected to be usually called only once, but it can happen based on timing that called
```
metacosm:
```suggestion
    // it is expected to be usually called only once, but it can happen based on timing that called
```

PR #3007, opened 2025-10-18
improve: ExternalResourceIDProvider extended to be used in CacheKeyMapper

csviri:
For 5.2 we introduced id provider for external resources:

This PR extends is to be used also for `ExternalResourceCachingEventSource` and it's subclasses.
Also:
- renames the `ExternalDependentIDProvider` to `ResourceIDProvider` to reflect that is used also for different purpose than dependent resources (event sources). Also having in mind forward compatiblity if we would still want to extend it's scope also for BulkDependentResources


csviri:
@metacosm @xstefank pls review this PR, would like to go forward with this for sure, will maybe add an additional PR to extend it to BulkDependentResource, but that we not necessarily want. But this is quite reasonable to do IMO.

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3019

metacosm:
Should the branch be deleted?

PR #3008, opened 2025-10-18
fix: disable performance test on resourceVersion comparison

csviri:
the test sometimes fails, we will iterate on this in any ways


metacosm:
This test pretty much fails all the time on my system and the difference is considerable (i.e.`Long.parseLong` is consistently more than 10 times faster) so we definitely need to review the implementation and figure out what to do. It might not make sense to hand-craft a parser if it cannot be made substantially faster than the "default" parsing + some extra validation.

csviri:
Yeah, it was not failing for me locally, we will iterate on this also based on prs of @shawkins in 5.3
simple parsing won't validate all the cases though.

PR #3010, opened 2025-10-19
enabling parse versions by default

shawkins:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -116,74 +71,68 @@ synchronized void onEvent(T resource, boolean unknownState) {
     cache.computeIfPresent(
         ResourceID.fromResource(resource),
         (id, cached) ->
-            (unknownState || !isLaterResourceVersion(id, cached, resource)) ? null : cached);
-  }
-
-  public synchronized void putAddedResource(T newResource) {
-    putResource(newResource, null);
+            (unknownState
+                    || PrimaryUpdateAndCacheUtils.compareResourceVersions(resource, cached) > 0)
```
csviri, line 75: This shouln't be `>=` , so we remove the cached also when it is the same?

PR #3012, opened 2025-10-21
Improve PodTemplateSpec sanitizer for GKE Autopilot compatibility

### LINKED ISSUE #3006: Early return in PodTemplateSpecSanitizer causes constant resource mismatch due to enforced resource requests/limits in GKE Autopilot clusters
#### What did you do?

Install a [HiveMQ Platform](https://github.com/hivemq/helm-charts/tree/master/charts/hivemq-platform) and [HiveMQ Platform Operator](https://github.com/hivemq/helm-charts/tree/master/charts/hivemq-platform-operator) Helm chart in a GKE Autopilot cluster.

I expect a smooth reconciliation.

We see a constant mismatch of our `StatefulSet` resource, so it's updated on every reconciliation:
```
15:04:28.712 [INFO] c.h.p.o.d.StatefulSetResourceMatcher - Detected changes in StatefulSet specification:
  Path: /spec/template/spec/containers/0/resources/limits/cpu
    Actual value: "1"
    Desired value: "1000m"

  Path: /spec/template/spec/containers/0/resources/requests/cpu
    Actual value: "1"
    Desired value: "1000m"
```
(`StatefulSetResourceMatcher` extends `SSABasedGenericKubernetesResourceMatcher` and uses the internal, pruned actual and desired maps for the diff logging)

This mismatch should be prevented by the `PodTemplateSpecSanitizer`. The actual root cause for the mismatch is hidden, due to an unlucky configuration of resource requests/limits and the interference of GKE Autopilot:

* The HiveMQ Platform Helm chart configures `cpu` requests/limits of `1000m` that will be serialized as `1` by K8s. So we require the `PodTemplateSpecSanitizer` in JOSDK to sanitize the `actualMap`, to prevent false positive mismatches on our `StatefulSet` resource.

* The HiveMQ Platform Helm chart doesn't configure `ephemeral-storage` requests/limits by default, but GKE Autopilot enforces this and updates our `StatefulSet` accordingly on-the-fly.

Under the hood we end up with these values in the matcher:
```yaml
desired:
  resources:
    limits:
      cpu: 1000m
      memory: 2048M
    requests:
      cpu: 1000m
      memory: 2048M
actual:
  resources:
    limits:
      cpu: 1                 # changed by K8s
      ephemeral-storage: 1Gi # added by GKE Autopilot
      memory: 2048M
    requests:
      cpu: 1                 # changed by K8s
      ephemeral-storage: 1Gi # added by GKE Autopilot
      memory: 2048M
```

The size mismatch of the actual and desired maps trigger [this early return in PodTemplateSpecSanitizer](https://github.com/operator-framework/java-operator-sdk/blob/520770084efbbcd8bed2fd6b29120712fc46998a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/PodTemplateSpecSanitizer.java#L127). So the `cpu` values are not sanitized and we end up with a false positive mismatch of the `StatefulSet`.

Since the desired state doesn't contain `ephemeral-storage`, there are no managed fields for this key in the requests/limits resources of our container. The `SSABasedGenericKubernetesResourceMatcher` then correctly prunes `ephemeral-storage` from the actual map, but also hides it as the actual root cause for the wrong `cpu` mismatch. For example, even with [debug logging](https://github.com/operator-framework/java-operator-sdk/blob/520770084efbbcd8bed2fd6b29120712fc46998a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcher.java#L116) the `ephemeral-storage` won't show up in the diff, because that uses the pruned actual map: `var diff = getDiff(prunedActual, desiredMap, objectMapper);`. The same applies to our custom logging, that also uses the pruned actual map.

#### Environment

**Kubernetes cluster type:** K8s 1.33.5 on GKE with Autopilot

`5.1.4`

`$ java -version`

```
openjdk version "21.0.8" 2025-07-15
OpenJDK Runtime Environment (build 21.0.8+9-Ubuntu-0ubuntu124.04.1)
OpenJDK 64-Bit Server VM (build 21.0.8+9-Ubuntu-0ubuntu124.04.1, mixed mode, sharing)
```

`$ kubectl version`

```
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.33.5-gke.1080000
```

#### Possible Solution

The easiest solution would be to remove the early return: `.filter(m -> m.size() == desiredResource.size())`.

This shouldn't cost much performance, since we still have two more early returns before we call `equals()` check that invokes the expensive `getNumericalAmount()`.

Donnerbart:
Fixes #3006

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/PodTemplateSpecSanitizer.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/PodTemplateSpecSanitizer.java
@@ -124,7 +124,6 @@ class PodTemplateSpecSanitizer {
                 "resources",
                 quantityPath))
         .map(Map.class::cast)
-        .filter(m -> m.size() == desiredResource.size())
         .ifPresent(
             m ->
                 actualResource.forEach(
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/PodTemplateSpecSanitizerTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/PodTemplateSpecSanitizerTest.java
@@ -133,19 +133,6 @@ class PodTemplateSpecSanitizerTest {
   }
 
   @Test
-  void testSanitizeResourceRequirements_whenResourceSizeMismatch_doNothing() {
-    final var actualMap =
-        sanitizeRequestsAndLimits(
-            ContainerType.CONTAINER,
-            Map.of("cpu", new Quantity("2")),
-            Map.of(),
-            Map.of("cpu", new Quantity("4")),
-            Map.of("cpu", new Quantity("4"), "memory", new Quantity("4Gi")));
-    assertContainerResources(actualMap, "requests").hasSize(1).containsEntry("cpu", "2");
-    assertContainerResources(actualMap, "limits").hasSize(1).containsEntry("cpu", "4");
-  }
-
-  @Test
   void testSanitizeResourceRequirements_whenResourceKeyMismatch_doNothing() {
     final var actualMap =
         sanitizeRequestsAndLimits(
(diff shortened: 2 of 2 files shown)


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcherTest.java
```diff
@@ -391,7 +385,7 @@ private static <R> R loadResource(String fileName, Class<R> clazz) {
         clazz, SSABasedGenericKubernetesResourceMatcherTest.class, fileName);
   }
 
-  private static class ConfigMapDR extends KubernetesDependentResource<ConfigMap, ConfigMap> {
+  private static class ConfigMapDR extends KubernetesDependentResource<ConfigMap, HasMetadata> {
```
csviri, line 388: nit: What is the reason for this change?

csviri (approved):
LGTM

thank you!!

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/SSABasedGenericKubernetesResourceMatcherTest.java
Donnerbart: To make the existing field `mockedContext` re-useable without any casts or warnings.

I've changed the context from `Context<?>` to `Context<HasMetadata>`, which works for all existing tests. And then aligned `ConfigMapDR` to use `HasMetadata` as primary resource as well.

The whole test class is now warning free in my IDEA.

PR #3015, opened 2025-10-22
Annotation removal using locking

shawkins:
Replacement / alternative to #2999.

Layered on the tombstone removal changes / default enablement of parsing resource versions. It turns out that some of the tombstone changes needed refinement anyway. For example, we need to compare to a localized latest version, not to the cache.

This seems like the most straight-forward way to do this with locking - but of course that means we have to be very careful with what locks are already held. If this seems too risky, then queuing managed in the TemporaryResourceCache could be used instead, but that will need more code restructuring.

These changes also assume we only want to deal with this skipping problem when using the InformerEventSource. That could be generalized if needed for #3011

Much like the PrimaryUpdateAndCacheUtils, users will need to call InformerEventSource.PrimaryUpdateAndCacheUtils, to properly use this via custom logic.

This does not do any cleanup the extraneous puts into the cache because of onCreated / onUpdated being called after the updateAndCacheResource method.

This also does not try to reason about deletes.

cc @csviri @metacosm

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -28,8 +28,6 @@ import org.slf4j.LoggerFactory;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Secret;
-import io.fabric8.kubernetes.api.model.apps.Deployment;
-import io.fabric8.kubernetes.api.model.apps.StatefulSet;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.CustomResource;
@@ -450,45 +448,15 @@ public interface ConfigurationService {
   }
 
   /**
-   * For dependent resources, the framework can add an annotation to filter out events resulting
-   * directly from the framework's operation. There are, however, some resources that do not follow
-   * the Kubernetes API conventions that changes in metadata should not increase the generation of
-   * the resource (as recorded in the {@code generation} field of the resource's {@code metadata}).
-   * For these resources, this convention is not respected and results in a new event for the
-   * framework to process. If that particular case is not handled correctly in the resource matcher,
-   * the framework will consider that the resource doesn't match the desired state and therefore
-   * triggers an update, which in turn, will re-add the annotation, thus starting the loop again,
-   * infinitely.
-   *
-   * <p>As a workaround, we automatically skip adding previous annotation for those well-known
-   * resources. Note that if you are sure that the matcher works for your use case, and it should in
-   * most instances, you can remove the resource type from the blocklist.
-   *
-   * <p>The consequence of adding a resource type to the set is that the framework will not use
-   * event filtering to prevent events, initiated by changes made by the framework itself as a
-   * result of its processing of dependent resources, to trigger the associated reconciler again.
-   *
-   * <p>Note that this method only takes effect if annotating dependent resources to prevent
-   * dependent resources events from triggering the associated reconciler again is activated as
-   * controlled by {@link #previousAnnotationForDependentResourcesEventFiltering()}
-   *
-   * @return a Set of resource classes where the previous version annotation won't be used.
-   */
-  default Set<Class<? extends HasMetadata>> withPreviousAnnotationForDependentResourcesBlocklist() {
-    return Set.of(Deployment.class, StatefulSet.class);
-  }
-
-  /**
-   * If the event logic should parse the resourceVersion to determine the ordering of dependent
-   * resource events.
+   * If the event logic can compare resourceVersions.
    *
    * <p>Enabled by default as Kubernetes does support this interpretation of resourceVersions.
-   * Disable only if your api server provides non comparable resource versions..
+   * Disable only if your api server provides non comparable resource versions.
    *
-   * @return if resource version should be parsed (as integer)
-   * @since 4.5.0
+   * @return if resource versions are comparable
+   * @since 5.3.0
    */
-  default boolean parseResourceVersionsForEventFilteringAndCaching() {
+  default boolean comparableResourceVersions() {
     return DEFAULT_COMPARABLE_RESOURCE_VERSIONS;
   }
 
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -97,7 +97,7 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
   private InformerEventSource(
       InformerEventSourceConfiguration<R> configuration,
       KubernetesClient client,
-      boolean parseResourceVersions) {
+      boolean comparableResourceVersions) {
     super(
         configuration.name(),
         configuration
@@ -105,7 +105,7 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
             .map(gvk -> client.genericKubernetesResources(gvk.apiVersion(), gvk.getKind()))
             .orElseGet(() -> (MixedOperation) client.resources(configuration.getResourceClass())),
         configuration,
-        parseResourceVersions);
+        comparableResourceVersions);
     // If there is a primary to secondary mapper there is no need for primary to secondary index.
     primaryToSecondaryMapper = configuration.getPrimaryToSecondaryMapper();
     if (useSecondaryToPrimaryIndex()) {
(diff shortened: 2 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -125,6 +121,22 @@ private InformerEventSource(
...
 
+  public R updateAndCacheResource(
+      R resourceToUpdate, Context<?> context, UnaryOperator<R> updateMethod) {
+    ResourceID id = ResourceID.fromResource(resourceToUpdate);
+    if (log.isDebugEnabled()) {
+      log.debug("Update and cache: {}", id);
+    }
+    try {
+      temporaryResourceCache.startModifying(id);
+      var updated = updateMethod.apply(resourceToUpdate);
```
csviri, line 134: We might want to iterate over this, the start modifyn and recent resource update should be called from an utils, but fine by me now as it is

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -46,144 +46,135 @@
...
 
-    if (moveAhead
-        || (cachedResource != null
-                && (cachedResource
-                    .getMetadata()
-                    .getResourceVersion()
-                    .equals(previousResourceVersion))
-            || isLaterResourceVersion(resourceId, newResource, cachedResource))) {
+    // first check against the source in general - this also prevents resurrecting resources when
+    // we've already seen the deletion event
```
csviri: As mentioned on other PR , would be nice to describe the resurrection scenario more in detial

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -46,144 +46,135 @@
...
+  boolean onEvent(T resource, boolean unknownState) {
+    ReentrantLock lock = activelyModifying.get(ResourceID.fromResource(resource));
+    if (lock != null) {
+      lock.lock(); // wait for the modification to finish
+      lock.unlock();
+    }
+    boolean[] known = new boolean[1];
+    synchronized (this) {
+      if (!unknownState) {
+        latestResourceVersion = resource.getMetadata().getResourceVersion();
```
csviri, line 113: While is this better this way as using the latest sync version?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerEventSourceConfiguration.java
```diff
@@ -96,18 +96,21 @@ class DefaultInformerEventSourceConfiguration<R extends HasMetadata>
     private final GroupVersionKind groupVersionKind;
     private final InformerConfiguration<R> informerConfig;
     private final KubernetesClient kubernetesClient;
+    private final boolean comparableResourceVersions;
```
csviri, line 100: We should add this also to `@Informer` annotation

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -149,16 +159,11 @@ public void onUpdate(R oldObject, R newObject) {
...
-        Operation.UPDATE,
-        newObject,
-        oldObject,
-        () -> InformerEventSource.super.onUpdate(oldObject, newObject));
+    onAddOrUpdate(Operation.UPDATE, newObject, oldObject);
   }
 
   @Override
-  public void onDelete(R resource, boolean b) {
+  public synchronized void onDelete(R resource, boolean b) {
```
csviri, line 168: I don't understand why we have this synchronized, could you elaborate pls?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
shawkins: I had initially started down the path of a common method with the primary cache utils - but for clarity this seemed like a better place to start. 

Are you expecting that we'll want to push the event skipping logic down into ManagedInformerEventSource, and/or are you thinking that we'll effectively be able to use a Context rather than a direct reference to the event source?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: We are effectively using the TemporaryResourceCache to reason about things the reconciler was responsible for creating / updating.

If we skip putting something in the cache, then we may not skip that event as intended.

Let's say the informer has seen:
- create v1
- modify v2 - the reconciler did this
- modify v3 - something else did this

But has only delivered to the TemporaryResourceCache
- create v1

And we're now attempting a put with v2. If we use the informer latest version, we won't actually put v2 into the TemporaryResourceCache as it looks out-of-date. But then on onAddOrUpdateEvent we'll answer that the event wasn't known and trigger a reconciliation.

So by using a local notion of latestResourceVersion we're reasoning over what the TemporaryResourceCache actually knows.

Also because the get logic has changed to check both the TemporaryResourceCache and the informer cache, we won't actually return an out-of-date resource due to the put.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
shawkins: The current logic has start and the onAddOrUpdate methods synchronized. I'm assuming that the intention there is to effectively block event handling until the primaryToSecondaryIndex has been built from the initial listing. I don't think delete should be handled differently.

That could also be done with a more granular lock around the usages of primaryToSecondaryIndex.

In general there's effectively a single threaded model of event delivery, so unless users are calling the onXXX methods manually, those calls will be serialized.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: I think using the Context will be fine, from a utility method where it passed later. 
So not complaining, it is fine like this. And we will se if we want to iterate.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
csviri: > That could also be done with a more granular lock around the usages of primaryToSecondaryIndex.

Yes, that might be cleaner.

> In general there's effectively a single threaded model of event delivery, so unless users are calling the onXXX methods manually, those calls will be serialized. 

Yeah, synchronizing the start with those makes it a bit safer and easier to reason about (without a real performance impact )

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -491,18 +480,16 @@ default Set<Class<? extends HasMetadata>> withPreviousAnnotationForDependentReso
...
-   * operator.
+   * <p>Enabled by default as Kubernetes does support this interpretation of resourceVersions.
+   * Disable only if your api server provides non comparable resource versions..
    *
    * @return if resource version should be parsed (as integer)
    * @since 4.5.0
    */
   default boolean parseResourceVersionsForEventFilteringAndCaching() {
-    return false;
+    return DEFAULT_COMPARABLE_RESOURCE_VERSIONS;
```
csviri, line 460: This needs to go to `@Informer` configuration and into ControllerConfiguration with dynamic configuration, but we can do it as a separate PR

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
shawkins: Expanded on that, and updated the class description.

shawkins:
Tried to clean up a bit more of the configuration and went ahead and switched to the comparableResourceVersions term - not sure how much backwards compatibility you'll want.

I need to look more into the Informer and ControllerConfiguration - I had started to add stuff, but it looks to be a little involved. Perhaps a follow-up pr would be best.

csviri (approved):
Hi @shawkins , this looks good to me, we can cleanup configs in a separate PR. In case happy to help with that.

shawkins:
> Hi @shawkins , this looks good to me, we can cleanup configs in a separate PR. In case happy to help with that.

Thank you @csviri

csviri:
Thank you @shawkins !!

PR #3016, opened 2025-10-23
improve: refining resource version comparison

shawkins:
After expanding the micro benchmark to include warm-up, and a variety of comparison scenarios, it does appear there are a couple of improvements that can be made. This seems to be the optimal and most straight-forward.

It is about 45% faster than using parseLong, but that still may not be good enough to rely upon in github actions without warmup due to garbage collection and or cpu throttling.

Another reason, besides performance, to not use parseLong is that I believe the Kubernetes folks are allowing for the possiblity of arbitrary length resourceVersions.

cc @metacosm @csviri


csviri:
Sorry I did not mean to merge this yet, but since cannot easily undo, will stick with it

PR #3020, opened 2025-10-24
feat: ResourceIDMapper for external event sources, external dependents and bulk dependents

csviri:


docs/content/fileList.txt
xstefank, line 1: why is this needed?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+package io.javaoperatorsdk.operator.processing;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.ExternalResourceCachingEventSource;
+
+/**
+ * Provides id for the target resource. This mapper is used across multiple component of the
```
xstefank:
```suggestion
 * Provides id for the target resource. This mapper is used across multiple components of the
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java
```diff
@@ -0,0 +1,58 @@
...
+ * framework, like the {@link
+ * io.javaoperatorsdk.operator.processing.dependent.AbstractExternalDependentResource}, in {@link
```
xstefank:
```suggestion
 * io.javaoperatorsdk.operator.processing.dependent.AbstractExternalDependentResource}, {@link
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java
```diff
@@ -0,0 +1,58 @@
...
+ * ExternalResourceCachingEventSource}, and {@link
+ * io.javaoperatorsdk.operator.processing.dependent.KubernetesBulkDependentResource}.
+ */
+public interface ResourceIDMapper<R, ID> {
+
+  ID idFor(R resource);
```
xstefank, line 39: This should also be javadoced

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.javaoperatorsdk.operator.processing;
+
+/**
+ * Provides the identifier for an object that represents resource. This ID is used to select target
```
xstefank:
```suggestion
 * Provides the identifier for an object that represents a resource. This ID is used to select the target
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ * external resource for a dependent resource from the resources returned by `{@link
+ * io.javaoperatorsdk.operator.api.reconciler.Context#getSecondaryResources(Class)}`. But also for
```
xstefank:
```suggestion
 * io.javaoperatorsdk.operator.api.reconciler.Context#getSecondaryResources(Class)}`. It is also used in
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ * {@link ResourceIDMapper} for event sources in external resources. But also for bulk dependent
+ * resource see {@link
+ * io.javaoperatorsdk.operator.processing.dependent.ExternalBulkDependentResource} and external
+ * event sources, see {@link
+ * io.javaoperatorsdk.operator.processing.event.source.ExternalResourceCachingEventSource}
```
xstefank: I would rewrite this as a bullet list

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param <ID>
+ */
+public interface ResourceIDProvider<ID> {
+
+  ID resourceId();
```
xstefank, line 37: Maybe it would be better to move the javadoc to the method so it is available in autocomplete

docs/content/fileList.txt
csviri: This is a deletion of empty file, it is not needed.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java
csviri: done. thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
csviri: done. thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java
csviri: added javadoc also here, thx

docs/content/en/docs/documentation/dependent-resource-and-workflows/dependent-resources.md
```diff
@@ -369,19 +369,22 @@ or we can use a matcher based SSA in most of the cases if the resource is manage
...
+the target resources from an event source, we provide a [mechanism](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractExternalDependentResource.java#L133-L147) that helps with that logic.
+[`ResourceIDMapper`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java)
+maps the resource to and ID and the ID of desired and actual resource is checked for equality.
+
+Your POJO representing an external resource can implement [`ResourceIDProvider`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDProvider.java). 
+The default `ResourceIDMapper` implementation works on top of resource which implement `ResourceIDProvider`:
```
```suggestion
The default `ResourceIDMapper` implementation works on top of resource which implements the `ResourceIDProvider`:
```

docs/content/en/docs/migration/v5-2-migration.md
```diff
@@ -0,0 +1,44 @@
+---
+title: Migrating from v5.1 to v5.2
```
xstefank, line 2: should this be in this PR?

csviri: yeah, these are related to these changes, not sure if there will be any other, but those will be in case separate PR.

metacosm (changes requested):
Several comments:
- As previously mentioned, I don't think that putting the responsibility of providing the external resource id on the external resource class is a good idea. We need to decide which design to go with now because this is already a breaking change and it shouldn't change again soon after, if we decide that putting the id "generation" on the dependent class (for example) is a better design later.
- The documentation for this feature is rather lacking and doesn't really explain the concepts or how to use things so it needs to be improved, imo.

csviri:
> As previously mentioned, I don't think that putting the responsibility of providing the external resource id on the external resource class is a good idea. 

That is completely optional, but allows us to add meaningful default implementation for the mapper. What alternative would you propose?


> if we decide that putting the id "generation" on the dependent class (for example) is a better design later.

Not shure what do you mean by this. Can you pls elaborate?

> The documentation for this feature is rather lacking and doesn't really explain the concepts or how to use things so it needs to be improved, imo.

Yeah, in general we have quite badly documented, features around it, would rather followup with a comprehensive rework on the docs.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/ResourceIDMapper.java
```diff
@@ -46,7 +46,7 @@ public interface ResourceIDMapper<R, ID> {
    * @param <R> secondary resource type
    */
   static <R> ResourceIDMapper<R, String> singleResourceResourceIDMapper() {
-    return r -> "josdk:mapper:id";
+    return r -> "id";
```
metacosm: The value should be irrelevant and we should rather use a value that has not risk of conflicting with an identifier a user would create.

csviri: Will have to upgrade upgrade the javadocs, there is no situation where there can be a collision. Assuming that this is by definition not usable for BulkDependentResources - will add javadoc.

If you see some situation where it actually can lead to a collision pls describe it.

metacosm: What's the harm of choosing a value that we're sure won't cause a collision, regardless of whether one such collision can happen? I mean if the value doesn't matter, then why not choose one that doesn't look like it is meaningful? For that matter, it could be a totally random one.

csviri: I belive in this cases we should focus:

1. the readability and simplicity is the key; 
2. And of course we should not solve not solve issues which are not present - so if there is no situation where it can cause collision, then we should not solve the issue of collision; since that would cause confusion.

So to fulfill those two as you said, any string that is simple enough would do. I choose simply `id`, but could be for example `void`, `anyId`, `any string`.  But imho (and maybe this is a personal pref in some cases), should not be any string that if someone reads the code will have to start thinking why is that a "special value".

csviri: But added also a comment that it could be any values, feel free to choose any simple and expressive string literal, it is fine by me.

metacosm: > So to fulfill those two as you said, any string that is simple enough would do. I choose simply `id`, but could be for example `void`, `anyId`, `any string`. But imho (and maybe this is a personal pref in some cases), should not be any string that if someone reads the code will have to start thinking why is that a "special value".

That's my point as well: reading the "id" value made me think that this was somehow a special value, since `id` is related to the semantic field of the feature. So maybe something like: "irrelevant"?

csviri: Ahh, yes, that sounds great, will change it.

csviri: changed, thx!!

PR #3021, opened 2025-10-24
Small health check improvements

Donnerbart:
We're digging into a reported operator health issue right now, and I noticed the `for for` typo in the debug log.

I also noticed a `RetryAndRescheduleTimerEventSource` with status `UNKNOWN` in the event source health indicators, caused by `TimerEventSource` not implementing the `getStatus()` method. If that was intentional (since we don't know if the actual timer tasks have been executed successfully), I can drop that commit again.


csviri:
Hi @Donnerbart  I think this is reasonable to change it this way. Thank you!

csviri (approved):
LGTM

csviri:
I did not checked yet, but this might be also interesting:
https://github.com/operator-framework/java-operator-sdk/issues/2978

xstefank (approved):

csviri:
thank you @Donnerbart !

PR #3027, opened 2025-10-31
improve: complete comparable resource version configs

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Informer.java
```diff
@@ -131,4 +132,9 @@
 
   /** Kubernetes field selector for additional resource filtering */
   Field[] fieldSelector() default {};
+
+  /**
+   * true if we can consider resource versions as integers, therefore it is valid to compare them
+   */
+  boolean comparableResourceVersion() default DEFAULT_COMPARABLE_RESOURCE_VERSION;
```
metacosm: javadoc could use an `@since` tag

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -459,5 +468,10 @@ public Builder withFieldSelector(FieldSelector fieldSelector) {
       InformerConfiguration.this.fieldSelector = fieldSelector;
       return this;
     }
+
+    public Builder withComparableResourceVersions(boolean comparableResourceVersions) {
```
metacosm, line 472: Method name should match the annotation field (i.e. versions vs version)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
```diff
@@ -166,26 +165,6 @@ public ConfigurationServiceOverrider withDefaultNonSSAResource(
...
-    this.comparableResourceVersions = value;
-    return this;
-  }
-
-  /**
-   * @deprecated use withComparableResourceVersions
-   * @param value true if internal algorithms can use metadata.resourceVersion as a numeric value.
-   * @return this
-   */
-  @Deprecated(forRemoval = true)
```
metacosm, line 183: Not sure we can remove this method, yet.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Informer.java
csviri: added , thx!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
csviri: changed in the annotation! thx!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
csviri: It is not used anywhere anymore, so I would say it is better to remove to avoid confusion.

PR #3028, opened 2025-11-01
improve: marking long running tests as integration test

csviri:
It seems that something changed in fabric8 client at some point, therefore these tests try to connect to a cluster, when the client is created. We can simply mark them as integration tests, so runs part of that suit, and does not prolong the build


metacosm:
> It seems that something changed in fabric8 client at some point, therefore these tests try to connect to a cluster, when the client is created. We can simply mark them as integration tests, so runs part of that suit, and does not prolong the build

Can you point to an occurrence of the issue? Maybe this is something we should address with the Fabric8 team?

metacosm (approved):

csviri:
> > It seems that something changed in fabric8 client at some point, therefore these tests try to connect to a cluster, when the client is created. We can simply mark them as integration tests, so runs part of that suit, and does not prolong the build
> 
> 
> 
> Can you point to an occurrence of the issue? Maybe this is something we should address with the Fabric8 team?

It seems that when the client is created it already does some connection. Will follow up on in fabric8 when back from vacation

Issue #3038, opened 2025-11-13
Warnings when using MicrometerMetrics with PrometheusMeterRegistry


marcinpeck:
#### What did you do?

Using operator with metrics. Implementation of Metrics (io.javaoperatorsdk.operator.monitoring.micrometer.MicrometerMetrics) is using 
io.micrometer.prometheusmetrics.PrometheusMeterRegistry.

This produces a lot of warnings in logs:

> Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'operator_sdk_events_received' containing tag keys [action, event, group, kind, name, namespace, scope, version]. The meter you are attempting to register has keys [event, group, kind, name, namespace, scope, version]. 

It happens in the following method:
https://github.com/operator-framework/java-operator-sdk/blob/6eb248689aec7ef542dab002a4077b90b1151793/micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetrics.java#L169

Mentioned implementation allows registration of the same metric but with different set of tags. This is not allowed by PrometheusMeterRegistry - [see](https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html#_limitation_on_same_name_with_different_set_of_tag_keys) as prometheus treats this as separate time series [see](https://prometheus.io/docs/practices/naming/)

#### What did you expect to see?

I expect to see all metrics without plenty of warnings in logs (Splunk). 

#### Possible Solution

As operator uses 2 types of events - resource related event with action and non-resource related - I would expect to have the same set of tags, but in case of non-resource related maybe something general. 

```
  public void receivedEvent(Event event, Map<String, Object> metadata) {
    if (event instanceof ResourceEvent) {
      incrementCounter(
          event.getRelatedCustomResourceID(),
          EVENTS_RECEIVED,
          metadata,
          Tag.of(EVENT, event.getClass().getSimpleName()),
          Tag.of(ACTION, ((ResourceEvent) event).getAction().toString()));
    } else {
      incrementCounter(
          event.getRelatedCustomResourceID(),
          EVENTS_RECEIVED,
          metadata,
          Tag.of(EVENT, event.getClass().getSimpleName()));
          Tag.of(ACTION, "NA");
    }
  }
```

csviri:
thx for reporting @marcinpeck , makes sense, do you plan to create a PR for this, would be great.

marcinpeck:
[...]

PR #3049, opened 2025-11-20
fix: add unknown action to MicrometerMetrics (#3038)

marcinpeck:
Solves #3038


csviri:
thank you @marcinpeck

PR #3050, opened 2025-11-20
improve: remove resource version comparison from 5.2

csviri:
Since it is not used in the framework yet internally removing it. We will further discuss the implementation for 5.3


metacosm (approved):

xstefank (approved):
not a big fan of removing public APIs in minor releases though :)

csviri:
> not a big fan of removing public APIs in minor releases though :)

This was not released yet.

PR #3063, opened 2025-11-25
docs: blog about 5.2 release

csviri:


docs/content/en/blog/releases/v5-2-release.md
```diff
@@ -0,0 +1,230 @@
...
+
+If you cannot modify the external resource class (e.g., it's generated or final), you can provide a custom
+`ResourceIDMapper` to the components above.
+
+See the [migration guide](/docs/migration/v5-2-migration) for detailed migration instructions.
+
+### Trigger Reconciliation on All Events
+
+Version 5.2 introduces a new execution mode that provides finer control over when reconciliation occurs. By setting
+[`triggerReconcilerOnAllEvent`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java#L81)
```
xstefank: This line number is taken from where? Not matching even on the `next` branch. This is why I don't think having line numbers in docs is a good idea and we should introduce a different mechanism.

xstefank (approved):

docs/content/en/blog/releases/v5-2-release.md
csviri: Yep, lets discuss how we could handle it, will remove this line number.

PR #3075, opened 2025-12-01
feat: ReconcileUtils for strongly consistent updates

csviri:
- The PR adds utility that provides methods to update resources using comparable resource versions
- Integrates this utility to the core of the framework (thus simplifying `ReconciliationDispatcher`)
  -  note that this introduces a change in the behavior for the `UpdateControl.patchStatus` (and others), since it won't trigger the reconiliation for the event in that update.
-  Renames former `ReconcilerUtils` to `ReconcilerUtilsInternal`, this is breaking but that utils was never advertised for non-internal usage 
- Includes also fixes for ControllerEventSource event filtering
- Improves TemporaryResourceCache so it explicitly supports non filtering updates

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
@@ -96,22 +96,6 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
     genericFilter = informerConfig.getGenericFilter();
   }
 
-  public R updateAndCacheResource(
-      R resourceToUpdate, Context<?> context, UnaryOperator<R> updateMethod) {
-    ResourceID id = ResourceID.fromResource(resourceToUpdate);
-    if (log.isDebugEnabled()) {
-      log.debug("Update and cache: {}", id);
-    }
-    try {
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -0,0 +1,983 @@
...
+import io.fabric8.kubernetes.client.dsl.base.PatchContext;
+import io.fabric8.kubernetes.client.dsl.base.PatchType;
+import io.javaoperatorsdk.operator.OperatorException;
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.informer.ManagedInformerEventSource;
+
+import static io.javaoperatorsdk.operator.processing.KubernetesResourceUtils.getUID;
+import static io.javaoperatorsdk.operator.processing.KubernetesResourceUtils.getVersion;
+
+public class ReconcileUtils {
```
csviri, line 38: As alternative names was thinking about `ReconcilerUtils` and `ReconciliationUtils`. What do you think @metacosm @shawkins @xstefank

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3106

PR #3076, opened 2025-12-01
fix(javadoc): invalid method ref blocks snapshot release

csviri:
simplified the javadoc, since quite hard to maintain proper description and we will document the public api rather in subsequent PRs


csviri:
> I have no idea how this is wrong though.

` {@link #canSkipEvent(HasMetadata, HasMetadata, ResourceID)} and related usage.` Was a wrong method reference, that method does not exists anymore.

PR #3077, opened 2025-12-01
fix: Make AbstractOperatorExtension use the kubernetesClient if the infrastructureKubernetesClient is not set

xstefank:
Should help in cases like https://github.com/operator-framework/java-operator-sdk/discussions/3070


csviri (changes requested):
would be nice to somehow verify this in an automated way, maybe using it in a test?

operator-framework-junit5/src/test/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtensionTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class LocallyRunOperatorExtensionTest {
+
+  @Test
+  void getAdditionalCRDsFromFiles() {
+    System.out.println(Path.of("").toAbsolutePath());
```
csviri, line 31: pls use a logger instead

operator-framework-junit5/src/test/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtensionTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class LocallyRunOperatorExtensionTest {
```
csviri, line 27: If I remember correctly this was on purpose an integration tests since we use the client there that does some initial calls on an API server

xstefank: well, it passes as unit test too so I saw no reason why to keep as IT

csviri: yes, but it slows down the local build, especially if there is no local k8s server running

csviri: see: https://github.com/operator-framework/java-operator-sdk/pull/3028

xstefank: yes, but this doesn't run long and also it seems that failsafe is not configured at all, so I think this as IT doesn't run at all

PR #3081, opened 2025-12-02
improve: rename junit5 module to junit

csviri:
Since we already using junit 6 this is already kind correct naming, and in the future will more junit major releases, would just name it as junit


xstefank (approved):
+1

PR #3082, opened 2025-12-02
feat: record desired state in Context

metacosm:

### DIFF
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResourceTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResourceTest.java
@@ -21,6 +21,7 @@ import java.util.Set;
 import org.junit.jupiter.api.Test;
 
 import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
```diff
@@ -157,4 +162,12 @@ public DefaultContext<P> setRetryInfo(RetryInfo retryInfo) {
     this.retryInfo = retryInfo;
     return this;
   }
+
+  @SuppressWarnings("unchecked")
+  public <R> R getOrComputeDesiredStateFor(
```
csviri, line 167: IMO this should be addded under `defaultManagedDependentResourceContext` , logically relates to that.

metacosm: This is used for non-managed dependent resources as well: it's used everywhere `desired` is used and this isn't part of the public API anyway. It's only public because it's used in different packages but otherwise it would have been a package-only method.

csviri (approved):

csviri:
Not sure if the test coverage is good enough this way, but let it be up to you to judge.

metacosm:
Hmm, looks like the expectation tests are flaky…

PR #3083, opened 2025-12-02
fix: grammar issue

metacosm:

### DIFF
--- a/docs/content/en/blog/releases/v5-2-release.md
+++ b/docs/content/en/blog/releases/v5-2-release.md
@@ -44,7 +44,7 @@ See the [migration guide](/docs/migration/v5-2-migration) for detailed migration
 ### Trigger Reconciliation on All Events
 
 Version 5.2 introduces a new execution mode that provides finer control over when reconciliation occurs. By setting
-[`triggerReconcilerOnAllEvent`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java)
+[`triggerReconcilerOnAllEvents`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ControllerConfiguration.java)
 to `true`, your `reconcile` method will be called for **every** event, including `Delete` events.
 
 This is particularly useful when:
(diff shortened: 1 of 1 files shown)


csviri:
There is one more here: https://github.com/operator-framework/java-operator-sdk/blob/684eda30f935e715ee6dd0155232eb0895a84ee0/docs/content/en/blog/releases/v5-2-release.md#L64

metacosm:
> There is one more here:
> 
> https://github.com/operator-framework/java-operator-sdk/blob/684eda30f935e715ee6dd0155232eb0895a84ee0/docs/content/en/blog/releases/v5-2-release.md#L64

I was about to change these as well (there's another one in that file as well, iirc) but since it's part of the blog, I decided against changing it but we should indeed mention it in the release notes and/or blog for 5.2.1.

csviri:
> > There is one more here:
> > https://github.com/operator-framework/java-operator-sdk/blob/684eda30f935e715ee6dd0155232eb0895a84ee0/docs/content/en/blog/releases/v5-2-release.md#L64
> 
> I was about to change these as well (there's another one in that file as well, iirc) but since it's part of the blog, I decided against changing it but we should indeed mention it in the release notes and/or blog for 5.2.1.

I think we should do it also, but I can do it as separate PR if you see it better that way, just to don't make confusion around it if some reads the blog.

csviri (approved):

metacosm:
> > > There is one more here:
> > > https://github.com/operator-framework/java-operator-sdk/blob/684eda30f935e715ee6dd0155232eb0895a84ee0/docs/content/en/blog/releases/v5-2-release.md#L64
> > 
> > 
> > I was about to change these as well (there's another one in that file as well, iirc) but since it's part of the blog, I decided against changing it but we should indeed mention it in the release notes and/or blog for 5.2.1.
> 
> I think we should do it also, but I can do it as separate PR if you see it better that way, just to don't make confusion around it if some reads the blog.

If we agree that the blog should be modified, I can do it here.

csviri:
> > > > There is one more here:
> > > > https://github.com/operator-framework/java-operator-sdk/blob/684eda30f935e715ee6dd0155232eb0895a84ee0/docs/content/en/blog/releases/v5-2-release.md#L64
> > > 
> > > 
> > > I was about to change these as well (there's another one in that file as well, iirc) but since it's part of the blog, I decided against changing it but we should indeed mention it in the release notes and/or blog for 5.2.1.
> > 
> > 
> > I think we should do it also, but I can do it as separate PR if you see it better that way, just to don't make confusion around it if some reads the blog.
> 
> If we agree that the blog should be modified, I can do it here.

I would vote for that, thx!

PR #3084, opened 2025-12-03
Improve Expectation integration test

Donnerbart:
From my understanding it's not optimal to use the `context` from the current reconciliation in `Expectation.createExpectation()`. This will hog on that specific instance, including the (cloned) primary resource instance that is internally stored in `DefaultContext`. This could have a bad impact on GC, and I think it could even lead to bugs when `context.getPrimaryResource()` is used in the expectation (since that primary resource might have outdated data).

The `Expectation.createExpectation(name, predicate)` provides the primary and context, that should be preferred to be used.

Fixed the integration test and references in the documentation and blog post. Also fixed all warnings in the test and re-ordered the methods by their usage in the code. Fixed the time unit for the timeout (from seconds to millis).

### DIFF
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/expectation/ExpectationIT.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/expectation/ExpectationIT.java
@@ -15,6 +15,8 @@
  */
 package io.javaoperatorsdk.operator.baseapi.expectation;
 
+import java.time.Duration;
+
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


csviri:
> From my understanding it's not optimal to use the context from the current reconciliation in Expectation.createExpectation(). This will hog on that specific instance, including the (cloned) primary resource instance that is internally stored in DefaultContext. This could have a bad impact on GC, and I think it could even lead to bugs when context.getPrimaryResource() is used in the expectation (since that primary resource might have outdated data).
The Expectation.createExpectation(name, predicate) provides the primary and context, that should be preferred to be used.

Absolutely, nice catch, thank you!

PR #3102, opened 2025-12-22
improve: log error in NodeExecutor

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/NodeExecutor.java
```diff
@@ -37,6 +42,10 @@ public void run() {
     } catch (Exception e) {
       // Exception is required because of Kotlin
       workflowExecutor.handleExceptionInExecutor(dependentResourceNode, e);
+    } catch (Error e) {
+      // without this user would see no sign about the error
+      log.error("java.lang.Error during execution", e);
```
metacosm, line 47: Should we also add it to the incremental recording of error messages that the workflow keeps?

csviri: I think we should not, this is just for a corner case, where we receive Errors. Those should not be caught in general in Java, and indicate a fundamental issue. Think of receving here an `OutOfMemoryError`. But also from the original reported issue the `NoSuchFieldError` indicates an issue with the bytecode. So we simply should not handle it.

PR #3107, opened 2025-12-30
fix: stop thread pools when an error occurs during startup

lion7:
Fixes https://github.com/operator-framework/java-operator-sdk/issues/3099
Tested by running with a locally built snapshot, it fixed the issue.

### DIFF
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/StopWithoutStartIT.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/StopWithoutStartIT.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright Java Operator SDK Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


csviri:
@lion7 thank you for the PR. The tests unfortunatelly fail, could you pls take a look, in case I can check too a bit later.

lion7:
@csviri I altered the order a bit so the `controllerManager` is stopped before the executor, so the graceful shutdown keeps working properly.

csviri:
Thank you @lion7 , could you pls also add an integration test for the base API:

https://github.com/operator-framework/java-operator-sdk/blob/589b7a6c8654cad13ff2317fddb51693fcd6d4c3/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi

Where without the junit extension you just create an `Operator` with a reconiler (dummy) and stop it without start, just to check if this don't throw an exception. I think that is a bare minimum for this (beside some manual testing)

lion7:
I added a unit test per your suggestion

csviri:
The CI failures are related to some github networking issue (occured also before),  I will re-run it bit later.

PR #3109, opened 2026-01-06
possible refinements

shawkins:
A few ideas for simplifying the logic.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
@@ -99,7 +99,7 @@ public class TemporaryResourceCache<T extends HasMetadata> {
   }
 
   /**
-   * @return true if the resourceVersion was already known and not skipped for event filtering
+   * @return true if the resourceVersion was obsolete
    */
   public boolean onAddOrUpdateEvent(T resource) {
     return onEvent(resource, false, false);
@@ -121,7 +121,7 @@ public class TemporaryResourceCache<T extends HasMetadata> {
       latestResourceVersion = resource.getMetadata().getResourceVersion();
     }
     var cached = cache.get(resourceId);
-    boolean filterEvent = false;
+    boolean obsoleteEvent = false;
     int comp = 0;
     if (cached != null) {
       comp = ReconcileUtils.validateAndCompareResourceVersions(resource, cached);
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterDetails.java
```diff
@@ -17,48 +17,36 @@
...
 import java.util.Optional;
 
+import io.javaoperatorsdk.operator.api.reconciler.ReconcileUtils;
 import io.javaoperatorsdk.operator.processing.event.source.controller.ResourceEvent;
 
 class EventFilterDetails {
 
   private int activeUpdates = 0;
   private ResourceEvent lastEvent;
-  private int lastUpdatedResourceVersion = -1;
```
shawkins, line 26: I don't think we can generally guarentee that the resource versions will fall in the int range.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -125,7 +119,7 @@ private synchronized boolean onEvent(T resource, boolean unknownState, boolean d
     int comp = 0;
     if (cached != null) {
       comp = ReconcileUtils.validateAndCompareResourceVersions(resource, cached);
-      if (comp >= 0 || unknownState) {
```
shawkins, line 128: @csviri can you explain why this allows for equality?

csviri: Since we can remove the resource in that case from the cache.

shawkins: Right, so this needs refined a little further - it should remove, but still be seen as a known event.

csviri: Note that in the implementation we strictly define two cases: when we want to filter the event and when we don't, based on user's intention.
So there is no grey area now.

shawkins: The wording may be confusing across a couple of methods. 

The boolean used to mean "known event" - so if I put a version into the temporary resource cache, then see the event with the same version, it's a known event.

Does "filter event" have a subtely different meaning from "known event"?

csviri: Hmm good question, the filter event mean to express that the event won't be propagated further (if true). This is bit more descriptive since we might actually not know all the events (watch connection issues).

PR #3114, opened 2026-01-08
Reconcile utils new alg

shawkins:


shawkins:
@csviri sorry about the pr confusion. Starting from a gh co put things in a different state that I'm used to. This should be in a better state now. 

To continue from https://github.com/operator-framework/java-operator-sdk/pull/3109#discussion_r2672593486 - I'm using the term obsolete in the TemporaryResourceCache. We previously had logic in onEvent that distinguished these cases, so I think we want to continue with that.

csviri:
> @csviri sorry about the pr confusion. Starting from a gh co put things in a different state that I'm used to. This should be in a better state now.
> 
> To continue from [#3109 (comment)](https://github.com/operator-framework/java-operator-sdk/pull/3109#discussion_r2672593486) - I'm using the term obsolete in the TemporaryResourceCache. We previously had logic in onEvent that distinguished these cases, so I think we want to continue with that.

sure fine with that, thx!

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryPrimaryResourceCacheTest.java
```diff
@@ -151,9 +151,9 @@ void putBeforeEvent() {
     nextResource.getMetadata().setResourceVersion("3");
     temporaryResourceCache.putResource(nextResource);
 
-    // the result is false since the put was not part of event filtering update
+    // now expect an event with the matching resourceVersion to be known after the put
     result = temporaryResourceCache.onAddOrUpdateEvent(nextResource);
-    assertThat(result).isFalse();
+    assertThat(result).isTrue();
```
csviri: This is a foundamental question with the new design, that if there is no explicit event filtering, we should propagate the new event for our update or more recent (larger resource version). So this should be false based on that.

PR #3115, opened 2026-01-08
Reconcile utils new alg

shawkins:
My continued appologies, I'm not having a good day. Via the IDE I still ended up targeting upstream, instead of origin.

That should be cleanup up now. Here's even further refinements to make clearer why I want to differentate the behavior.

### DIFF
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryPrimaryResourceCacheTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryPrimaryResourceCacheTest.java
@@ -31,6 +31,7 @@ import io.javaoperatorsdk.operator.processing.event.ResourceID;
 import io.javaoperatorsdk.operator.processing.event.source.informer.TemporaryResourceCache.EventHandling;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class TemporaryPrimaryResourceCacheTest {
 
@@ -179,7 +180,7 @@ class TemporaryPrimaryResourceCacheTest {
   }
 
   @Test
-  void putAfterEventWithEventFiltering() {
+  void putAfterEventWithEventFilteringNoPost() {
     var testResource = testResource();
 
     // first ensure an event is not known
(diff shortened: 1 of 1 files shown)


shawkins:
> This is a foundamental question with the new design, that if there is no explicit event filtering, we should propagate the new event for our update or more recent (larger resource version). So this should be false based on that.

@csviri regardless of event filtering my expectation is that is something is added to the cache, it is something that we don't then care to get an event about.

I've updated things here to leave it more up to the caller to decide what to do in each of the cases.

Could you provide an example where someone adds a version to the temporary resource cache, but still wants that event?

csviri:
> > This is a foundamental question with the new design, that if there is no explicit event filtering, we should propagate the new event for our update or more recent (larger resource version). So this should be false based on that.
> 
> @csviri regardless of event filtering my expectation is that is something is added to the cache, it is something that we don't then care to get an event about.
> 
> I've updated things here to leave it more up to the caller to decide what to do in each of the cases.
> 

Hi @shawkins thx for the PR, tests are failing, can take a look a bit later more in depth into this PR.

> Could you provide an example where someone adds a version to the temporary resource cache, but still wants that event?

Yeah, this is up to the discussion, my original idea was to provide a backwards compatible way for the case if some implemented a controller, that at the end of the reconciliation call `UpdateControl.patchResource` and the event resulted from that patch is triggered the reconiliation - as a I framework I think we cannot make an assumption in general that no one is expecting that it will trigger the reconiliation. So wanted to have a way to have still the guarantee that up to date resources is in cache always, but we want to trigger an event. Therefore I added this option:

https://github.com/operator-framework/java-operator-sdk/blob/793a56d9d00f1989f5369b22233a113356a3bd2a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/AbstractUpdateControl.java#L20-L31

And we pass that flag also to in the reconciler for the utils, like here:

https://github.com/operator-framework/java-operator-sdk/blob/793a56d9d00f1989f5369b22233a113356a3bd2a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ReconciliationDispatcher.java#L395

However, now I'm thinking - and especially if this makes the implementation simlper - that we could rather just simply have a `reSchedule()` method without time delay here:

https://github.com/operator-framework/java-operator-sdk/blob/793a56d9d00f1989f5369b22233a113356a3bd2a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/BaseControl.java#L26

So we will produce that event explicitly in the event processor after reconciliation. I would acutally go rather with this option (will update the PR), and let you optimize filtering / caching part as in this PR, if that works for you.

thank you!

csviri:
see: https://github.com/operator-framework/java-operator-sdk/pull/3106/commits/00f3245ff394435cb933bb5e331289a0f1b145d0

csviri:
> Could you provide an example where someone adds a version to the temporary resource cache, but still wants that event?

Ahh there is one more place, when we adding a finalizer we want to receive an event (and having an up-to-date resource) but we can also have there an explicit additional event.

shawkins:
> Hi @shawkins thx for the PR, tests are failing, can take a look a bit later more in depth into this PR.

Sure, I'll have a look.

> I would acutally go rather with this option (will update the PR), and let you optimize filtering / caching part as in this PR, if that works for you.

An immediate re-reconciliation makes sense, and gives more control to the calling logic.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryPrimaryResourceCacheTest.java
```diff
@@ -172,9 +173,29 @@ void putBeforeEventWithEventFiltering() {
...
+    var nextResource = testResource();
+    nextResource.getMetadata().setResourceVersion("3");
+    var resourceId = ResourceID.fromResource(testResource);
+
+    temporaryResourceCache.startEventFilteringModify(resourceId);
+    result = temporaryResourceCache.onAddOrUpdateEvent(nextResource);
+    // the result is deferred
+    assertThat(result).isEqualTo(EventHandling.DEFER);
+    temporaryResourceCache.putResource(nextResource);
+    temporaryResourceCache.doneEventFilterModify(resourceId, "3");
```
csviri: Would be nice to assert the result here

shawkins: Ok, updated that test and added one more.

shawkins:
> Otherwise I think we can merge this if you agree

Sounds good to me.

csviri:
Awesome, thank you @shawkins !!

shawkins:
Thanks @csviri

PR #3119, opened 2026-01-13
improve: moving some integration test into packages, some @Sample cleanup

csviri:
Having those tests in packages gives better visibility on webpage.
Removing sample from two tests, which does not demonstrate functionlity, rather just tests some cases.


csviri:
> Why remove Sample annotations?

The intention is to have samples which are user facing, describe some use case, from integration tests which are really just testing some aspect of the framework. For example concurrency tests are not really useful for users.

PR #3127, opened 2026-01-15
Event filtering now records resource action and previous resource

csviri:
This is important to have a correct further event propagation, thus to have an event propagation equivalent to a real scenario.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
@@ -127,9 +127,8 @@ public class TemporaryResourceCache<T extends HasMetadata> {
     }
     var cached = cache.get(resourceId);
     EventHandling result = EventHandling.NEW;
-    int comp = 0;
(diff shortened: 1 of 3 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -251,11 +251,11 @@ public boolean allowsNamespaceChanges() {
     return configuration().followControllerNamespaceChanges();
   }
 
-  private boolean eventAcceptedByFilter(Operation operation, R newObject, R oldObject) {
+  private boolean eventAcceptedByFilter(ResourceAction operation, R newObject, R oldObject) {
```
xstefank: maybe also rename the param to "action"

csviri: fixed, thank you!

shawkins (approved):
LGTM

csviri:
Note that InformerEventSource and ControllerEventSource logic would be possible to uninfy even more, also the tests, but will stop at this point, and will portentially improve that later.

PR #3130, opened 2026-01-21
feat: distinct latest resource collector

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -364,13 +372,13 @@ public static <R extends HasMetadata> R resourcePatch(
...
       throw new IllegalStateException("No event source found for type: " + resource.getClass());
     }
+    var es = esList.get(0);
     if (esList.size() > 1) {
-      throw new IllegalStateException(
-          "Multiple event sources found for: "
-              + resource.getClass()
-              + " please provide the target event source");
+      log.warn(
+          "Multiple event source found for type: {}, selecting first with name {}",
```
xstefank:
```suggestion
          "Multiple event sources found for type: {}, selecting first with name {}",
```

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -364,13 +372,13 @@ public static <R extends HasMetadata> R resourcePatch(
...
+          resource.getClass(),
+          log.isWarnEnabled() ? es.name() : null);
```
xstefank: is this check necessary since you are already logging warn?

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
```diff
@@ -440,6 +441,271 @@ void resourcePatchThrowsWhenEventSourceIsNotManagedInformer() {
...
+                    .withName("pod2")
+                    .withNamespace("default")
+                    .withResourceVersion("100")
+                    .build())
+            .build();
+
+    List<Pod> result =
+        Stream.of(pod1v1, pod1v2, pod2v1).collect(ReconcileUtils.latestDistinctList());
+
+    assertThat(result).isInstanceOf(List.class);
```
xstefank: unnecessary, result is already List<Pod>

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
```diff
@@ -440,6 +441,271 @@ void resourcePatchThrowsWhenEventSourceIsNotManagedInformer() {
...
+                new ObjectMetaBuilder()
+                    .withName("pod2")
+                    .withNamespace("default")
+                    .withResourceVersion("100")
+                    .build())
+            .build();
+
+    Set<Pod> result = Stream.of(pod1v1, pod1v2, pod2v1).collect(ReconcileUtils.latestDistinctSet());
+
+    assertThat(result).isInstanceOf(java.util.Set.class);
```
xstefank: same as above

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
csviri: well, sonar flags this for the unnecessary method call, but yeah, will remove it since it does not makes too much sense.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
csviri: Thank you, fixed

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
csviri: Thank you, fixed

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -714,4 +722,56 @@ private static int validateResourceVersion(String v1) {
...
+   * Returns a collector that deduplicates Kubernetes objects by keeping only the one with the
+   * latest metadata.resourceVersion for each unique name and namespace combination. The intended
+   * use case is for the rather rare setup when there are overlapping {@link
+   * io.javaoperatorsdk.operator.processing.event.source.informer.InformerEventSource}s for a
+   * resource type.
+   *
+   * @param <T> the type of HasMetadata objects
+   * @return a collector that produces a List of deduplicated Kubernetes objects
+   */
+  public static <T extends HasMetadata> Collector<T, ?, List<T>> latestDistinctList() {
```
metacosm, line 632: I don't get why this method or the set variant are needed or even useful. The semantic should be a `Set` since instances should be unique, no?

csviri: The `Set` variant is needed since we are covering the use case whern there are multiple event sources for same type without distinct set of resource watched, so we returning the latest of those. 

I added List just for convinience. If somebody would prefer working above Lists, that is quite common I beleive, but you are right that the semantically `Set` is the correct one.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -714,4 +722,56 @@ private static int validateResourceVersion(String v1) {
...
+    return Collectors.collectingAndThen(latestDistinctToMap(), map -> new HashSet<>(map.values()));
+  }
+
+  private static <T extends HasMetadata> Collector<T, ?, Map<ResourceID, T>> latestDistinctToMap() {
+    return Collectors.toMap(
+        resource ->
+            new ResourceID(resource.getMetadata().getName(), resource.getMetadata().getNamespace()),
+        resource -> resource,
+        (existing, replacement) ->
+            compareResourceVersions(existing, replacement) >= 0 ? existing : replacement);
```
csviri, line 657: cc @metacosm this is the point to keep of this PR, just to keep the latest

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
```diff
@@ -364,13 +373,13 @@ public static <R extends HasMetadata> R resourcePatch(
     if (esList.isEmpty()) {
       throw new IllegalStateException("No event source found for type: " + resource.getClass());
     }
+    var es = esList.get(0);
     if (esList.size() > 1) {
-      throw new IllegalStateException(
-          "Multiple event sources found for: "
-              + resource.getClass()
-              + " please provide the target event source");
+      log.warn(
```
metacosm, line 378: This would technically be an API breaking change…

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
metacosm: This isn't resolved, the assertion is still present.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
metacosm: Wouldn't it be better if that was done transparently by the SDK? Also, wouldn't it be better to provide a filtered stream on `Context` rather than a `Collector` implementation?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
csviri: What do you mean, could you pls elaborate?  We don't have this released yet.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtilsTest.java
csviri: ahhh, sorry, fixed, thx!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
csviri: Ahh additional issue with adding it as default, is that this applies only for kubernetes resources, while in `getSecondaryResources` handles also external resources.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ReconcileUtils.java
metacosm: Ah yes, my bad!

csviri:
replacced by: https://github.com/operator-framework/java-operator-sdk/pull/3141

PR #3137, opened 2026-01-27
improve: move compare resource version methods to internal utils

csviri:
This should not be user facing. At least not in any obvious scenerio.


metacosm:
Split on this. On the one hand, I agree that there's probably not much need to expose this. On the other hand, I could see users wanting to perform these comparisons as well and we might as well let them use our implementation. That said, they could call `ReconcileUtilsInternal` anyway?

csviri:
> Split on this. On the one hand, I agree that there's probably not much need to expose this. On the other hand, I could see users wanting to perform these comparisons as well and we might as well let them use our implementation. That said, they could call `ReconcileUtilsInternal` anyway?

Yeah, that is the idea, so basically I don't see any use case when users should call it, and therefore hiding it, on the other hand if they are doing some very specific operation nothing stops them to using it from the internal. This approach felt a bit more correct, but up to debate.

PR #3142, opened 2026-01-30
feat: move ReconcileUtils methods to ResourceOperations accessible from Context

csviri:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
@@ -23,7 +23,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
-import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.dsl.base.PatchContext;
 import io.fabric8.kubernetes.client.dsl.base.PatchType;
@@ -34,6 +33,13 @@ import io.javaoperatorsdk.operator.processing.event.source.informer.ManagedInfor
 import static io.javaoperatorsdk.operator.processing.KubernetesResourceUtils.getUID;
 import static io.javaoperatorsdk.operator.processing.KubernetesResourceUtils.getVersion;
 
+/**
+ * Provides useful operations to manipulate resources (server-side apply, patch, etc.) in an
+ * idiomatic way, in particular to make sure that the latest version of the resource is present in
+ * the caches for the next reconciliation.
+ *
+ * @param <P> the resource type on which this object operates
+ */
 public class ResourceOperations<P extends HasMetadata> {
 
   public static final int DEFAULT_MAX_RETRY = 10;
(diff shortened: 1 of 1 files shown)


metacosm:
I was thinking that maybe instead of `KubernetesClientFacade`, we could name it `ExtendedKubernetesClient` and return it from the `getClient` method of the `Context` but that would mean either:
- have `ExtendedKubernetesClient` implement `KubernetesClient`, hold an instance of the original client and delegate all method calls to it but that would mean having to track changes in the interface in this implementation, which might be annoying and easy to miss
- have a `getUnderlyingClient` (or whatever name makes more sense) on the `ExtendedKubernetesClient` class but that would mean that the client would be even "further" away (and in this case, `KubernetesClientFacade` would probably be a better name

Actually, the second solution is rather ugly and would be breaking the API anyway…
Also, for the first solution, the methods wouldn't really be at the same semantic level so both are probably a bad idea 😅

metacosm:
Based on my last comment, I think we could keep the `ReconcileUtils` name or maybe `ResourceModifier` or `ResourceChangeUtils` (something to mean that it has methods to modify resources)?

xstefank (commented):
For some reason, I'm allergic tothe  word "facade" :D

I think sth like `context.operatorUtils()`, `context.operations()`, `context.sdkOperations()` would be better

csviri:
> Based on my last comment, I think we could keep the `ReconcileUtils` name or maybe `ResourceModifier` or `ResourceChangeUtils` (something to mean that it has methods to modify resources)?

Naming is always a good question :) 

Maybe `ResourceUpdateFacade`, `ResourceOperationsFacade`, at the end it is a facade on underlying APIs, or?

csviri:
> For some reason, I'm allergic tothe word "facade" :D

just saw your comment now, why are you allergic? it is actually quite expressing notion

csviri:
For me utils suffix always impies some static functions...

metacosm:
I'd vote for `ResourceOperations`, nice & simple. I'm not allergic to `Facade` but I also don't think it adds anything here :)

csviri:
> I'd vote for `ResourceOperations`, nice & simple. I'm not allergic to `Facade` but I also don't think it adds anything here :)

Sounds good, will change it, and finalize the PR. Thank you!

xstefank:
I personally always find facade as an indication of an unnecessary abstraction layer.

sound good to me too, now only misses an implementation

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -357,7 +438,7 @@ public static <R extends HasMetadata> R jsonMergePatchPrimaryStatus(
    * @param <R> resource type
    * @throws IllegalStateException if no event source or multiple event sources are found
    */
-  public static <R extends HasMetadata> R resourcePatch(
+  public <R extends HasMetadata> R resourcePatch(
       Context<?> context, R resource, UnaryOperator<R> updateOperation) {
```
metacosm: `context` parameter should be removed here.

csviri: good catch, thx, fixed

csviri:
> I've made a couple of changes to make things clearer and, in some instances, more correct. I will open a subsequent issue regarding the handling of the primary resource, actually.

Sure, will check those Monday, we can interate on this more.  I also want to create few additional issue. Thx!

PR #3146, opened 2026-02-02
improve: KubernetesDependentResource uses resource operations directly

csviri:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
```diff
@@ -104,35 +85,38 @@ public R create(R desired, P primary, Context<P> context) {
...
-            .serverSideApply()
-        : resource.create();
+    log.debug(
+        "Creating target resource with type: {}, with id: {} use ssa: {}",
+        desired.getClass(),
+        ResourceID.fromResource(desired),
+        ssa);
+
+    return ssa
+        ? context.resourceOperations().serverSideApply(desired)
```
csviri: Probably we should have a variant for these methods where we can specify event source. Will add it.

xstefank (approved):

PR #3151, opened 2026-02-04
improve: naming on ResourceID API

csviri:

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
@@ -62,17 +62,9 @@ public class ResourceID implements Serializable {
     return Objects.equals(name, that.name) && Objects.equals(namespace, that.namespace);
   }
 
-  public boolean isSameResource(HasMetadata hasMetadata) {
-    if (hasMetadata == null) {
-      return false;
-    }
-    final var metadata = hasMetadata.getMetadata();
-    return isSameResource(metadata.getName(), metadata.getNamespace());
-  }
-
   /**
-   * Whether this ResourceID points to the same resource as the one identified by the specified name
-   * and namespace.
+   * Whether this ResourceID points to the same resource as the one identified only by the specified
+   * name and namespace.
    *
    * <p>Note that this doesn't take API version or Kind into account so this should only be used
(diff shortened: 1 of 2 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
```diff
@@ -83,7 +75,7 @@ public boolean isSameResource(HasMetadata hasMetadata) {
    *     specified name and namespace, {@code false} otherwise
    * @since 5.3.0
    */
-  public boolean isSameResource(String name, String namespace) {
+  public boolean equals(String name, String namespace) {
```
metacosm, line 78: I don't think that `equals` is a better name (because `equals` has a special meaning in Java and it doesn't quite match here), though I agree that `isSameResource` is not great either… maybe `equilaventTo`? This way, it would be somewhat discoverable if someone is looking for equality-like methods?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
```diff
@@ -62,17 +62,9 @@ public boolean equals(Object o) {
     return Objects.equals(name, that.name) && Objects.equals(namespace, that.namespace);
   }
 
-  public boolean isSameResource(HasMetadata hasMetadata) {
```
metacosm, line 65: This would technically be an API break.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
csviri: Yes, I though overloaded would be too much an issue, but you might be right that naming wise this could raise some eyebrows.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/ResourceID.java
csviri: That is true, so not sure to how extent we consider this as an internal API, but since users use this with customt `SecondaryToPrimaryMapper` it might be an issue.

csviri:
Ok, considering this we probably don't want to break this API will close this.

PR #3153, opened 2026-02-04
docs: improve dependent resource mermaid visuals

csviri:


csviri:
should look like something like this:

image

csviri:
> fix the commit message please

final commit message will be set when squashing, will take care of it, thx

PR #3154, opened 2026-02-04
feat: metricsV2 + oTel + prometheus sample and Grafana dashboard

csviri:
Goal of this PR is to provide a OTel + Prometheus + Grafana setup. So we:
1. verify integration with OTel
2. Provide a default Grafana dashboard for metrics.
3. Adds a new operator and E2E test for h OTel + Prometheus + Grafana handling verification. This should be easy to done by a users, so dey can check and validate the dashboard.
4. new metrics implementations

Notes on new metrics implementation:

- only static gauges (lives until the operator lives) - that makes much easir to manage them
- no manual counter removal (should not be done)
- sanitized names and labels according to best practices
- added some metrics counter for last attempt retries.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GroupVersionKindPlural.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GroupVersionKindPlural.java
@@ -119,7 +119,7 @@ public class GroupVersionKindPlural extends GroupVersionKind {
    * @return the default plural form for the specified kind
    */
   public static String getDefaultPluralFor(String kind) {
-    // replace by Fabric8 version when available, see
+    // TODO replace by Fabric8 version when available, see
     // https://github.com/fabric8io/kubernetes-client/pull/6314
     return kind != null ? Pluralize.toPlural(kind.toLowerCase()) : null;
   }


csviri:
JVM metrics:

image

image

csviri:
JODSK metrics:

image

Added TODOs to improve those, like we should not have controller name as suffix, rather as a tag:

`operator_sdk_reconciliations_executions_webpagestandalonedependentsreconciler`

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,336 @@
...
+          Tag.of(EVENT, event.getClass().getSimpleName()),
+          Tag.of(ACTION, UNKNOWN_ACTION));
+    }
+  }
+
+  @Override
+  public void cleanupDoneFor(ResourceID resourceID, Map<String, Object> metadata) {
+    gauges.get(numberOfResourcesRefName(getControllerName(metadata))).decrementAndGet();
+    incrementCounter(EVENTS_DELETE, resourceID.getNamespace().orElse(null), metadata);
+  }
```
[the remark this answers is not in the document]
csviri: thank you fixed

csviri:
@metacosm @xstefank this should be super easy to try out, just run the e2e test against minikube, and check the grafana dashboard (port forward grafana (port 3000) from the cluster). Pls let me know what do you think.

docs/content/en/docs/documentation/observability.md
```diff
@@ -77,30 +77,108 @@ Metrics metrics; // initialize your metrics implementation
...
-resource is deleted and how the clean-up is performed. See the relevant classes documentation for more details.
+#### MicrometerMetricsV2 metrics
+
+All meters use `controller.name` as their primary tag. Counters optionally carry a `namespace` tag when
+`withNamespaceAsTag()` is enabled.
+
+| Meter name (Micrometer)                  | Type    | Tags                              | Description                                                          |
+|------------------------------------------|---------|-----------------------------------|----------------------------------------------------------------------|
+| `reconciliations.executions`             | gauge   | `controller.name`                 | Number of reconciler executions currently in progress                |
+| `reconciliations.active`                 | gauge   | `controller.name`                 | Number of resources currently queued for reconciliation              |
```
xstefank: is "active" really a good name for this?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,331 @@
...
+  private static final String RECONCILIATIONS_FAILED =
+      RECONCILIATIONS + FAILURE_SUFFIX + TOTAL_SUFFIX;
+  private static final String RECONCILIATIONS_SUCCESS =
+      RECONCILIATIONS + SUCCESS_SUFFIX + TOTAL_SUFFIX;
+  private static final String RECONCILIATIONS_RETRIES_NUMBER =
+      RECONCILIATIONS + "retries" + TOTAL_SUFFIX;
+  private static final String RECONCILIATIONS_STARTED = RECONCILIATIONS + "started" + TOTAL_SUFFIX;
+
+  private static final String RECONCILIATIONS_EXECUTIONS_GAUGE = RECONCILIATIONS + "executions";
+  private static final String RECONCILIATIONS_QUEUE_SIZE_GAUGE = RECONCILIATIONS + "active";
```
xstefank: queue would be better name

csviri: The thing is that queue is a bit misleading, because this is actually the queue + ongoing reconiliations, that is the reason I added active. (ExecutorService has a queue). Can expand the docs on this

docs/content/en/docs/documentation/observability.md
csviri: see below.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: Actually made an adjustment so we have now `active` and `queue` where queue is that are submitted to executor service but not started yet, the active is the number of actually running threads with reconiliation logic. This seems to be more intuitive and useful.

docs/content/en/docs/migration/v5-3-migration.md
```diff
@@ -26,4 +26,26 @@ to
       <version>5.3.0<version>
       <scope>test</scope>
 </dependency>
-```
\ No newline at end of file
+```
+
+## Metrics interface changes
```
metacosm, line 31: This would technically be an API break and would require a new major version.

observability/install-observability.sh
```diff
@@ -0,0 +1,312 @@
...
+NC='\033[0m' # No Color
+
+echo -e "${GREEN}========================================${NC}"
+echo -e "${GREEN}Installing Observability Stack${NC}"
+echo -e "${GREEN}OpenTelemetry + Prometheus + Grafana${NC}"
+echo -e "${GREEN}========================================${NC}"
+
+# Check if helm is installed, download locally if not
+echo -e "\n${YELLOW}Checking helm installation...${NC}"
+if ! command -v helm &> /dev/null; then
```
metacosm, line 33: Installing helm should be made optional. People might not want to have helm automatically installed or want to install it themselves some other way.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,328 @@
...
+package io.javaoperatorsdk.operator.monitoring.micrometer;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.jspecify.annotations.NonNull;
```
metacosm: This should not have been introduced here. Nullability should be reviewed at the SDK level, not simply here for 2 methods.

docs/content/en/docs/documentation/observability.md
```diff
@@ -77,30 +77,108 @@ Metrics metrics; // initialize your metrics implementation
...
+| `reconciliations.started.total`      | counter | `controller.name`, `namespace`*                   | Number of reconciliations started (including retries)                |
+| `reconciliations.success.total`      | counter | `controller.name`, `namespace`*                   | Number of successfully finished reconciliations                      |
+| `reconciliations.failure.total`      | counter | `controller.name`, `namespace`*                   | Number of failed reconciliations                                     |
+| `reconciliations.retries.total`      | counter | `controller.name`, `namespace`*                   | Number of reconciliation retries                                     |
+| `events.received`                    | counter | `controller.name`, `event`, `action`, `namespace` | Number of Kubernetes events received by the controller               |
+
+\* `namespace` tag is only included when `withNamespaceAsTag()` is enabled.
+
+The execution timer uses explicit SLO boundaries (10ms, 50ms, 100ms, 250ms, 500ms, 1s, 2s, 5s, 10s, 30s) to ensure
```
metacosm: SLO acronym should be used expanded first before.

docs/content/en/docs/documentation/observability.md
```diff
@@ -77,30 +77,108 @@ Metrics metrics; // initialize your metrics implementation
...
+compatibility with `histogram_quantile()` queries in Prometheus. This is important when using the OTLP registry, where
```
metacosm: OTLP acronym should be used expanded first.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,328 @@
...
+        });
+  }
+
+  @Override
+  public void eventReceived(Event event, Map<String, Object> metadata) {
+    if (event instanceof ResourceEvent resourceEvent) {
+      if (resourceEvent.getAction() == ResourceAction.ADDED) {
+        gauges.get(numberOfResourcesRefName(getControllerName(metadata))).incrementAndGet();
+      }
+      var namespace = resourceEvent.getRelatedCustomResourceID().getNamespace().orElse(null);
```
metacosm, line 170: Shouldn't that be in sync with what's done for MDC, i.e. use a default value instead of `null` for clustered resources?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,328 @@
...
+
+  private void addNamespaceTag(String namespace, List<Tag> tags) {
+    if (includeNamespaceTag && namespace != null && !namespace.isBlank()) {
+      tags.add(Tag.of(NAMESPACE, namespace));
+    }
+  }
+
+  private void incrementCounter(
+      String counterName, String namespace, Map<String, Object> metadata, Tag... additionalTags) {
+    final var tags = new ArrayList<Tag>(2 + additionalTags.length);
```
metacosm, line 254: This shouldn't be a List but a Set.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,328 @@
...
+          metadata,
+          Tag.of(EVENT, event.getClass().getSimpleName()),
+          Tag.of(ACTION, resourceEvent.getAction().toString()));
+    } else {
+      incrementCounter(
+          EVENTS_RECEIVED,
+          null,
+          metadata,
+          Tag.of(EVENT, event.getClass().getSimpleName()),
+          Tag.of(ACTION, UNKNOWN_ACTION));
```
metacosm: Why unknown action here?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
```diff
@@ -0,0 +1,328 @@
...
+  public void reconciliationSubmitted(
+      HasMetadata resource, RetryInfo retryInfoNullable, Map<String, Object> metadata) {
+    Optional<RetryInfo> retryInfo = Optional.ofNullable(retryInfoNullable);
+
+    var namespace = resource.getMetadata().getNamespace();
+    incrementCounter(RECONCILIATIONS_STARTED, namespace, metadata);
+
+    int retryNumber = retryInfo.map(RetryInfo::getAttemptCount).orElse(0);
+    if (retryNumber > 0) {
+      incrementCounter(RECONCILIATIONS_RETRIES_NUMBER, namespace, metadata);
```
metacosm, line 198: How does the retry number get propagated? Seems like the counter that is incremented is completely decorrelated from the actual retry number?

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: Removed this, supposedly if we cannot add meaningful value it should added, thank you pointing out

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: This way at the end we don't add the namespace tag (if null), that is according the guidelines.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: removed this, not sure how it got here, good catch!

docs/content/en/docs/migration/v5-3-migration.md
csviri: Strinctly sepaking yes, but such minor API changes we do some times, see the migration document. As other frameworks sometimes. It is basically I think a better choice in terms of tradeoff, since because we don't really want to increase the major verion that often and we on the other hand we have quite an amount of APIs,  that sometimes better to evolve this way IMO.

I also was trying to do backwards compatible, we still could. But at the end it looked like that it would be more confusing, that just having a table to be able to easily migrate from current impl. If that makes sense.

docs/content/en/docs/documentation/observability.md
csviri: added thx

docs/content/en/docs/documentation/observability.md
csviri: SLO I consider a well known term, but actually we can just remote that from here. done.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: That is really an implementation detail IMO

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: The idea is that we measure the number of retries, so we increase this number if the actual reconiliation is a retry. What is interesting at the end in the metrics is the rate of the retries.

observability/install-observability.sh
csviri: What this does, it checks if helm command works (so already installed), if not, downloads it as a binary, but just locally in the project dir (so does not install it), and uses that binary directly. So this is not very invasive at the end IMO, usually it is done this way in such scripts.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
csviri: Hmm, was thinking about this, actually, should not be a Set: basically, what we really want is just a list of Tags in memory, that we later iterate through. While Set might create some secondary data structures like indexes for fast access, what we don't want in this case. So using list is a better both memory and mental model IMO :)

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
metacosm: Then I'd argue that MDC should follow the same guidelines, otherwise it's just confusing to have telemetry report one thing while MDC does something else, especially if you're trying to correlate logs and telemetry.

csviri: I don't see why should we unify those completely unrelated systems.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
metacosm: We don't want duplicated tags, nor do we care about their order, that's what set are for and that matches what people would expect. As a user, I don't want duplicated tags to show up, which is a possibility with the current implementation.

csviri: If you take the look on Micrometer API:

```java
 public Counter counter(String name, Iterable<Tag> tags) {
        return this.counter(name, Tags.of(tags));
    }

    public Counter counter(String name, String... tags) {
        return this.counter(name, Tags.of(tags));
    }

    public Counter counter(String name, Tags tags) {
        return this.counter(new Meter.Id(name, tags, (String)null, (String)null, Type.COUNTER));
    }
```

Does not work with Sets either.

Set's usually maintains an addiitonal index in memory, we don't want to have that memory overhead. While adding an additional tag should not be our responsibility filter out, that is an usage error.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GroupVersionKindPlural.java
```diff
@@ -119,7 +119,7 @@ public static GroupVersionKindPlural gvkFor(Class<? extends HasMetadata> resourc
    * @return the default plural form for the specified kind
    */
   public static String getDefaultPluralFor(String kind) {
-    // todo: replace by Fabric8 version when available, see
+    // replace by Fabric8 version when available, see
```
metacosm: Why was the todo removed? This still needs to be done.

csviri: Ahh should not be removed in this PR, will put it back.

Wanted to remove these TODOs for which we already have issues, so that when go through TODOs I see just relevant, in general it is a good practice to remove all the TODO's from the code before we merge and rather create issue, since it pollutes the code.

micrometer-support/src/main/java/io/javaoperatorsdk/operator/monitoring/micrometer/MicrometerMetricsV2.java
metacosm: I'm fine with using `Iterable`. The semantics should not be a list, though, because this implies order, which does not exist.

csviri: They have also an array of tags api, that implies order...  I don't think that is a bad approach; the other way to look at this, is that we just want to have a list of elemnts in memory (not a linked list, not a set, since those have memory overhead).

Is there order? yes
Do we care about the order? no

csviri: But will take a look how it would look like with iterable.

csviri: Hmm, checking, but those are 2 liner methods, internal APIs, there is no memory overhead. This is completely  fine this way.

csviri:
@metacosm @xstefank I think we clarified all the points, here if not objections will merge it today.

PR #3155, opened 2026-02-04
improve: event filtering algorithm for multiple parallel updates

csviri:


csviri:
> This makes sense. To drive home the usage pattern you could consider collapsing decreaseActiveUpdates, handleLastOwnUpdatedResourceVersion, and getActiveUpdates to a single call -
> 
> int decreaseActiveUpdates(String resourceVersion)

 did some simplification; feel free to do additional (in this or separate pr),

and mainly thank you for feedback!

PR #3156, opened 2026-02-05
refactor: avoid creating intermediate collections when unneeded

metacosm:
Also use constant filters and collectors


csviri:
@metacosm the test fails seems to be relevant

metacosm:
> @metacosm the test fails seems to be relevant

Yes, looking at it, thanks!

PR #3161, opened 2026-02-06
improve: prepare for removal of exitOnStopLeading from public API

metacosm:


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/LeaderElectionConfiguration.java
```diff
@@ -82,6 +82,9 @@ public LeaderElectionConfiguration(
     this(leaseName, leaseNamespace, leaseDuration, renewDeadline, retryPeriod, null, null, true);
   }
 
+  /**
+   * @deprecated Use {@link LeaderElectionConfigurationBuilder} instead
+   */
```
csviri, line 87: Please add proper annotation not just javadoc

csviri:
The test failures are not releated, but we have to check on those later

PR #3162, opened 2026-02-06
docs: moving the start days of the invite

csviri:
had a issue to test this, that might be caused the first date is
in 2024- what anyway is unnecessary


metacosm:
No need to run CI on these changes

PR #3163, opened 2026-02-08
feat: support additional java CRDs in test extension

### LINKED ISSUE #3004: LocallyRunOperatorExtension issue with loading CRD from file
Hi, I found something a bit unexpected while integration testing.

### The use case:

We want to integrate with a third party, optionally setting up an Informer Event Source if the third party CRD is present in kubernetes. The third party (Strimzi) offers an API module that contains fabric8 `CustomResource`s and `CustomResourceDefinitions`s, but not any CRD yaml manifests.

### The problem

To integration test this I want to use `LocallyRunOperatorExtension` to additionally install one of these Custom Resource Definitions.

I cannot use `LocallyRunOperatorExtension.builder().withAdditionalCustomResourceDefinition(Kafka.class)` as their module does not contain CRD files in META-INF, and JOSDK looks like it would want them in a specific naming convention too.

I can take the relevant `CustomResourceDefinition` and serialize it to a temporary file and attempt to apply it with ` LocallyRunOperatorExtension.builder().withAdditionalCRD("/path/to/crd)`. But it appears that the `LocallyRunOperatorExtension` applies the CRD after it registers the Reconciler, so the CRD is not applied at the point we are trying to prepare the Event Sources.

Here's a [minimal reproducer ](https://github.com/robobario/josdk-reproducer) based off of `io.javaoperatorsdk:bootstrapper:5.1.4:create`

The reproducer also demonstates a workaround, using a fabric8 KubernetesClient from an `@BeforeAll` and `@AfterAll` to directly apply the `CustomResourceDefinition` from the module.

### Ideas/Questions

1. Would you be open to supporting the fabric8 `CustomResourceDefinition` directly? Like adding a `withAdditionalCustomResourceDefinition(CustomResourceDefinition definition)` type method and applying them at the same time as the Class<CustomResource> based one.
2. Why are the CRDs passed with `withAdditionalCustomResourceDefinition` and `withAdditionalCRD` applied at different times? Could they all be applied before the Reconciler is registered?

Thanks!
I'm happy to take a crack at making those changes

robobario:
Adds `withAdditionalCustomResourceDefinition(CustomResourceDefinition definition)` to `LocallyRunOperatorExtension#Builder`. These definitions are applied at the same time as the existing `additionalCustomResourceDefinitions`. The definitions are deleted according to the existing lifecycle.

Resolves #3004

Why:
In some cases we may wish to apply a CustomResourceDefinition that has no corresponding CRD file. For example, some frameworks publish fabric8-based api modules which contain only the java classes for their CRDs, and no CRD files.

### DIFF
--- a/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
+++ b/operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
@@ -67,6 +67,7 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
   private final List<PortForwardSpec> portForwards;
   private final List<LocalPortForward> localPortForwards;
   private final List<Class<? extends CustomResource>> additionalCustomResourceDefinitions;
+  private final List<CustomResourceDefinition> additionalCustomResourceDefinitionInstances;
   private final Map<Reconciler, RegisteredController> registeredControllers;
   private final Map<String, String> crdMappings;
   private final Consumer<LocallyRunOperatorExtension> beforeStartHook;
@@ -76,6 +77,7 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
       List<HasMetadata> infrastructure,
       List<PortForwardSpec> portForwards,
       List<Class<? extends CustomResource>> additionalCustomResourceDefinitions,
+      List<CustomResourceDefinition> additionalCustomResourceDefinitionInstances,
       Duration infrastructureTimeout,
       boolean preserveNamespaceOnError,
       boolean waitForNamespaceDeletion,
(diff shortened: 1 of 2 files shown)


operator-framework-junit5/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -195,6 +198,33 @@ public void applyCrd(Class<? extends CustomResource> crClass) {
     applyCrd(ReconcilerUtils.getResourceTypeName(crClass));
   }
 
+  public void applyCrd(CustomResourceDefinition customResourceDefinition) {
+    try {
+      String resourceTypeName = customResourceDefinition.getMetadata().getName();
+      final var pathAsString = crdMappings.get(resourceTypeName);
```
robobario, line 204: I am unsure whether this logic should be applied.

For the other mechanism ``` withAdditionalCustomResourceDefinition(Class<? extends CustomResource> customResource) {``` it checks the custom `crdMappings` first and prefers to apply a CRD supplied there, else it looks in the standard location.

What should we do if the `crdMappings` contains a CRD with the same name as this `CustomResourceDefinition`?

csviri: IMHO, we should check if we already applied CRD with same name, if yes throw an exception, since probably that is not what user intends. This can be a separate PR.

csviri:
@xstefank pls take a look also

csviri:
@robobario @xstefank if you have the bandwidth and could do the duplicate detection would be great, but mainly thank you for this PR.

csviri:
@robobario if that helps we can have a patch release this week for you

robobario:
thanks @csviri we have a workaround in place so there's no urgency to release, whenever suits you

PR #3250, opened 2026-03-24
fix: register event sources when dependents are marked for deletion

SamBarker:
Adds a failing integration test to reproduce the bug described in #3249.

## Workflow under test

```
ConfigMapDependentResource  (reconcilePrecondition = ALWAYS_FALSE)
  └── SecretBulkDependentResource  (activationCondition = ALWAYS_TRUE)
```

On first reconciliation the ConfigMap precondition fails → `markDependentsForDelete` cascades to `SecretBulkDependentResource` → `NodeDeleteExecutor` fires → the Secret event source was never registered (because `NodeReconcileExecutor` never ran) → `NoEventSourceForClassException`.

The test asserts this exception does **not** occur, and currently **fails**, demonstrating the bug.

I'm quite happy to attempt a fix, but didn't wanted to make sure it was confirmed as a bug first

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileExecutor.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileExecutor.java
@@ -246,6 +246,10 @@ class WorkflowReconcileExecutor<P extends HasMetadata> extends AbstractWorkflowE
     // so if the activation condition was false, this node is not meant to be deleted.
     var dependents = dependentResourceNode.getParents();
     if (activationConditionMet) {
+      // make sure we register the dependent's event source if it hasn't been added already
+      // this might be needed in corner cases such as
+      // https://github.com/operator-framework/java-operator-sdk/issues/3249
+      registerOrDeregisterEventSourceBasedOnActivation(true, dependentResourceNode);
       createOrGetResultFor(dependentResourceNode).markForDelete();
       if (dependents.isEmpty()) {
         bottomNodes.add(dependentResourceNode);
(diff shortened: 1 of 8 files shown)


metacosm:
@SamBarker can you confirm that this fixes your issue?

SamBarker:
> @SamBarker can you confirm that this fixes your issue?

Yes I've built 5.3.1-SNAPSHOT locally and tested the kroxylicious PR with it. I no longer see the `NoEventSourceForClassException` exception.

SamBarker (commented):
LGTM

Thanks for the quick turn around @metacosm

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileExecutor.java
```diff
@@ -246,6 +246,10 @@ private void markDependentsForDelete(
     // so if the activation condition was false, this node is not meant to be deleted.
     var dependents = dependentResourceNode.getParents();
     if (activationConditionMet) {
+      // make sure we register the dependent's event source if it hasn't been added already
+      // this might be needed in corner cases such as
+      // https://github.com/operator-framework/java-operator-sdk/issues/3249
```
csviri, line 251: just nit: instead of linking the issue we might want to describe the case when it is needed. (Actually this is quite an obvious why it should be there so might not even need a description, but maybe it just me).

csviri (approved):
Awesome, LGTM. 

just added a nit, but looks great. thank you all!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/workflow/WorkflowReconcileExecutor.java
metacosm: I think the issue explains the issue way better than I could do in a short comment and that provides historical context for the change that might be useful later on because that method call here is quite unexpected, imo.

PR #3279, opened 2026-04-08
fix: delete CRD only after all tests in the class pass

### LINKED ISSUE #2928: [Junit extension] Delete CRD only after All test not each
Deleting CRD after each test case does not make sense, and causes all subsequent tests to fail after one test case failed since CRD stays in "deleting" state.

xstefank:
Fixes #2928

### DIFF
--- a/operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
+++ b/operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
@@ -28,7 +28,9 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


xstefank:
@csviri I wasn't sure if you mean after all tests or after test class, so I did a after a test class first, but I can move it to after testsuite.

xstefank:
@csviri this is ready for review

csviri:
thx @xstefank , will take a look tomorrow; but was thinking if we should target `next` branch, since it is change in the behavior (not super important though)

csviri (approved):
nice! LGTM

operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -387,12 +392,27 @@ protected void after(ExtensionContext context) {
...
 
-  private void deleteCrd(AppliedCRD appliedCRD, KubernetesClient client) {
-    if (!deleteCRDs) {
-      LOGGER.debug("Skipping deleting CRD because of configuration: {}", appliedCRD);
-      return;
+  private static class CrdCleanup implements ExtensionContext.Store.CloseableResource {
+    @Override
+    public void close() {
+      // Create a fresh client for cleanup since operator clients may already be closed.
+      try (var client = new KubernetesClientBuilder().build()) {
```
metacosm, line 399: Isn't it the role of the infrastructure client to perform these operations?

xstefank: no, because this method runs after `@After` method that closes the infrastructure client.

PR #3302, opened 2026-04-20
improve: owner reference mappers checks only group not apiVersion

csviri:
In case of a resource is has a new version, but owner reference
still points to the old version we should still propapage the event.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternal.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternal.java
@@ -361,4 +361,16 @@ public class ReconcilerUtilsInternal {
     }
     return v1Length;
   }
+
+  public static String getGroup(String apiVersion) {
+    if (apiVersion.startsWith("/")) {
+      apiVersion = apiVersion.substring(1);
+    }
+    var index = apiVersion.indexOf("/");
+    if (index < 0) {
+      return "";
+    } else {
+      return apiVersion.substring(0, index);
+    }
+  }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternal.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternal.java
@@ -45,6 +45,13 @@ public class ReconcilerUtilsInternal {
   private static final String GET_STATUS = "getStatus";
   private static final Pattern API_URI_PATTERN =
       Pattern.compile(".*http(s?)://[^/]*/api(s?)/(\\S*).*"); // NOSONAR: input is controlled
+  private static final String DOT = ".";
+  private static final char DOT_CHAR = '.';
+  private static final String SLASH = "/";
+  private static final char SLASH_CHAR = '/';
+  private static final String EMPTY = "";
+  private static final char ZERO = '0';
+  private static final char NINE = '9';
 
   // prevent instantiation of util class
   private ReconcilerUtilsInternal() {}
@@ -55,7 +62,7 @@ public class ReconcilerUtilsInternal {
 
   public static String getResourceTypeNameWithVersion(Class<? extends HasMetadata> resourceClass) {
     final var version = HasMetadata.getVersion(resourceClass);
-    return getResourceTypeName(resourceClass) + "/" + version;
+    return getResourceTypeName(resourceClass) + SLASH + version;
   }
 
   public static String getResourceTypeName(Class<? extends HasMetadata> resourceClass) {
@@ -69,7 +76,7 @@ public class ReconcilerUtilsInternal {
   public static String getDefaultFinalizerName(String resourceName) {
     // resource names for historic resources such as Pods are missing periods and therefore do not
     // constitute valid domain names as mandated by Kubernetes so generate one that does
-    if (resourceName.indexOf('.') < 0) {
+    if (resourceName.indexOf(DOT_CHAR) < 0) {
       resourceName = resourceName + MISSING_GROUP_SUFFIX;
     }
     return resourceName + FINALIZER_NAME_SUFFIX;
@@ -102,7 +109,7 @@ public class ReconcilerUtilsInternal {
 
   public static String getDefaultReconcilerName(String reconcilerClassName) {
     // if the name is fully qualified, extract the simple class name
-    final var lastDot = reconcilerClassName.lastIndexOf('.');
+    final var lastDot = reconcilerClassName.lastIndexOf(DOT_CHAR);
     if (lastDot > 0) {
       reconcilerClassName = reconcilerClassName.substring(lastDot + 1);
     }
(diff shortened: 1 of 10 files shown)


csviri:
cc @shawkins

csviri:
see also: https://github.com/keycloak/keycloak/issues/48030#issuecomment-4270016428

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternalTest.java
```diff
@@ -308,6 +309,14 @@ void compareResourceVersionsWithHasMetadata() {
     assertThat(ReconcilerUtilsInternal.compareResourceVersions(resource2, resource1)).isPositive();
   }
 
+  @Test
+  void extractsGroupFromApiVersion() {
+    assertThat(getGroup("v1")).isEqualTo("");
+    assertThat(getGroup("/v1")).isEqualTo("");
```
metacosm: This isn't a valid apiVersion String

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternalTest.java
```diff
@@ -308,6 +309,14 @@ void compareResourceVersionsWithHasMetadata() {
...
+    assertThat(getGroup("apps/v1")).isEqualTo("apps");
+    assertThat(getGroup("/apps/v1")).isEqualTo("apps");
```
metacosm: This isn't a valid apiVersion so I don't think we should cover this case, apart maybe to throw an IllegalArgumentException

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
```diff
@@ -98,10 +100,12 @@ public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerRefer
...
   public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerReferences(
       String apiVersion, String kind, boolean clusterScope) {
-    String correctApiVersion = apiVersion.startsWith("/") ? apiVersion.substring(1) : apiVersion;
     return resource ->
         resource.getMetadata().getOwnerReferences().stream()
-            .filter(r -> r.getKind().equals(kind) && r.getApiVersion().equals(correctApiVersion))
+            .filter(
+                r ->
+                    r.getKind().equals(kind)
+                        && getGroup(r.getApiVersion()).equals(getGroup(apiVersion)))
```
metacosm, line 108: I'm not sure this is the correct behavior: shouldn't there be a conversion hook instead? It seems dangerous to potentially match resources with differing versions…

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternal.java
```diff
@@ -361,4 +361,16 @@ private static int validateResourceVersion(String v1) {
     }
     return v1Length;
   }
+
+  public static String getGroup(String apiVersion) {
```
metacosm, line 372: This method should probably eventually move to the client code itself

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
csviri: Conversion hooks, does not update the owner references of the resource. Maybe we should have an integration test also this, to see how exactly and what happens. I will add one.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
```diff
@@ -98,10 +100,12 @@ public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerRefer
 
   public static <T extends HasMetadata> SecondaryToPrimaryMapper<T> fromOwnerReferences(
       String apiVersion, String kind, boolean clusterScope) {
-    String correctApiVersion = apiVersion.startsWith("/") ? apiVersion.substring(1) : apiVersion;
```
csviri, line 101: cc @metacosm see this part. Also if you check the related commit, that explains the issue here.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternalTest.java
csviri: I made this so we are backwards compatible with the current implementation marked you on the related code part. But I tend to agree that we should make it throw such exception in next minor version.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternalTest.java
csviri: see conversation here: https://github.com/operator-framework/java-operator-sdk/pull/3302#discussion_r3110463820

csviri:
Maybe would be more pc to target `next` with this PR.

xstefank (approved):
I don't have anything else to add to what was already mentioned.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
metacosm: Hmm, that's weird, I wonder why we have that code here to start with… Let me look into it.

metacosm: Indeed, this was a workaround for a bug in the client which has been addressed in https://github.com/fabric8io/kubernetes-client/commit/84f46da12cee6128a421930706a6f268eeeb729c so we should actually remove this workaround as well now.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/ReconcilerUtilsInternalTest.java
metacosm: Since the client does not have the underlying issue anymore, I don't think that we should implement these backwards compatible checks since they shouldn't be needed anymore.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/Mappers.java
csviri: great, thank you for checking! Will update the PR.

csviri: updated.

csviri:
@xstefank @metacosm I added also an E2E test, it seems that even if the CR is updated to new apiVersion the owner ref stays the same. So the implementation is correct this way.

metacosm (approved):

shawkins:
Thank you everyone.

PR #3315, opened 2026-04-27
Flag for opting out delete CRDs for LocallyRunOperatorExtension

csviri:

### DIFF
--- a/operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
+++ b/operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
@@ -65,8 +65,8 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
   private static final Set<AppliedCRD> appliedCRDs = new HashSet<>();
   private static final boolean DELETE_CRDS_DEFAULT =
       Boolean.parseBoolean(System.getProperty("testsuite.deleteCRDs", "true"));
-  private static volatile boolean deleteCRDs = DELETE_CRDS_DEFAULT;
 
+  private volatile boolean deleteCRDs = DELETE_CRDS_DEFAULT;
   private final Operator operator;
   private final List<ReconcilerSpec> reconcilers;
   private final List<PortForwardSpec> portForwards;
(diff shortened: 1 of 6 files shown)


xstefank (approved):
I still don't understand why it would be useful for somebody though.

csviri:
> I still don't understand why it would be useful for somebody though.

Let's say you have a test cluster that runs the operator, and have pre installled CRDs. But you want to run also integrationt tests on that cluster, you don't want to delete CRDs after tests are already run.

operator-framework-junit/src/main/java/io/javaoperatorsdk/operator/junit/LocallyRunOperatorExtension.java
```diff
@@ -63,8 +63,9 @@ public class LocallyRunOperatorExtension extends AbstractOperatorExtension {
   private static final int CRD_DELETE_TIMEOUT = 5000;
   private static final int CRD_DELETE_WAIT_TIMEOUT = 60000;
   private static final Set<AppliedCRD> appliedCRDs = new HashSet<>();
-  private static final boolean deleteCRDs =
+  private static final boolean DELETE_CRDS_DEFAULT =
       Boolean.parseBoolean(System.getProperty("testsuite.deleteCRDs", "true"));
+  private static volatile boolean deleteCRDs = DELETE_CRDS_DEFAULT;
```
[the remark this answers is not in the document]
csviri: this was good point, fixed.

sample-operators/metrics-processing/src/test/java/io/javaoperatorsdk/operator/sample/metrics/MetricsHandlingE2E.java
```diff
@@ -74,6 +74,7 @@ class MetricsHandlingE2E {
           ? LocallyRunOperatorExtension.builder()
               .withReconciler(new MetricsHandlingReconciler1())
               .withReconciler(new MetricsHandlingReconciler2())
+              .withDeleteCRDs(false)
```
metacosm: If this isn't required for the tests to pass, we should keep the previous behavior of deleting the CRD so that the tests leave the target cluster clean once done.

csviri: Hmm something went wrong with cherry pick it seems, will remove it. thx!

csviri:
But I was thinking maybe we could do this more intelligent, and apply the CRD only if not exists, and delete it if we applied in the test. WDYT? @metacosm @xstefank  (that would be without a flag)

csviri:
Ok found why the tomcat is flaky on tests: basically we should not delete the CRD at the end of the test of local run operator since that causes race condition with the namespace cleanup and CRD delete. So we need this flag. Will update the E2E tests as a separate PR. cc @xstefank

PR #3318, opened 2026-04-27
fix: remove MultiOwnerDependentTriggeringIT

### LINKED ISSUE #3124: Enhance Dependent Resources sample / impl to how to manage resource with multiple owners
Note that there is now an issue that when one of the CR's is deleted it actually not cleanup it's own data in `MultiOwnerDependentTriggeringIT`.

One solution is to not have the DR garbage collected, and override the delete to remove the data and owner reference label (it's label if not GC-d). But that we have to customize also the owber label key naming.

Note sure how or if we want to support this use case more explicitly since it is quite rare, so for now will just show in the test how to do it with current setup.

See discussion here:
https://github.com/operator-framework/java-operator-sdk/discussions/3122
https://github.com/operator-framework/java-operator-sdk/discussions/3123

csviri:
Note that although the test was passing it had multiple issues.
Since this is not properly supported removing the IT so it does not make confusion.


xstefank (commented):
I don't understand the reasoning why removing and not fixing it?

csviri:
> I don't understand the reasoning why removing and not fixing it?

It would be quite an effort to fix it, that would be feature in dependent resources. Feel free to take a look at it, we can have a call to explore.

metacosm (approved):

PR #3325, opened 2026-04-29
Informer pools

### LINKED ISSUE #2415: Informer Pools
In some cases, informers might be shared especially between controllers. So if multiple controllers watch same resource (with same label selectors, namespaces, item store ) sharing it would spare resources

csviri:
## Summary

Introduces an **informer pool** so that informers backing `InformerEventSource`s (and the internal controller event source) can be **shared across controllers** instead of each controller/event source always creating its own `SharedIndexInformer`. This reduces the number of watch connections opened against the API server and memory overhead in operators where multiple controllers (or multiple dynamically-registered event sources) watch the same secondary resource type (e.g. `ConfigMap`, `Secret`).

This is marked `@Experimental`: the pooling behavior itself is production ready, but the configuration API may still evolve in a non-backwards-compatible way.

### `InformerPool` abstraction (`processing/event/source/informer/pool`)

- **`InformerPool`** — interface for getting/releasing informers, keyed by an `InformerClassifier`, plus reference counting via `numberOfInformersForResource`.
- **`InformerClassifier`** — identifies "equivalent" informer requests: API server URL, label/field/shard selectors, namespace, resource type (or GVK for generic resources), and item store. `informerListLimit` and indexers are intentionally excluded from identity (a warning is logged if an existing informer is reused with a different list limit; indexers can be added independently to a shared informer as long as names don't collide).
- **`AbstractInformerPool`** — shared logic for building the underlying `SharedIndexInformer` from a classifier (selectors, field selectors, item store, list limit, exception/stopped handlers) and starting it with the configured `cacheSyncTimeout`.
- **`DefaultInformerPool`** (new default) — reference-counted, sharing pool: creates an informer on first request for a given classifier and reuses it for subsequent requests; only stops it once the last user releases it.
- **`NonSharingInformerPool`** — opt-out pool that creates a dedicated informer per event source, preserving pre-pooling behavior.

### Configuration

- `ConfigurationService.informerPool()` — new (non-default) method; implementations must return the same instance on every call so controllers using the same `ConfigurationService` share the pool. `AbstractConfigurationService` caches a `DefaultInformerPool` lazily (synchronized to avoid creating more than one instance under concurrent first access).
- `ConfigurationServiceOverrider.withInformerPool(InformerPool)` — lets users opt into `AlwaysNewInformerPool` (or a custom implementation) to disable/customize sharing.

### Integration

- `InformerManager` and `InformerWrapper` were reworked to go through the pool instead of creating/starting/stopping the `SharedIndexInformer` directly:
  - `InformerManager` now builds an `InformerClassifier` per namespace and calls `informerPool.getInformer(...)` / `releaseInformer(...)` instead of constructing the informer from the Fabric8 client itself.
  - Starting/stopping is now delegated to the pool: `getInformer` only registers/reference-counts, `InformerPool#start` blocks (idempotently) until cache sync, and `releaseInformer` decrements the reference count, stopping the informer only when it drops to zero. The event handler is always removed from the returned informer, even when it stays running for other users.
  - `InformerWrapper` no longer owns lifecycle (`start`/`stop` removed); it now just exposes the informer/classifier and cache-facing behavior.
- `ControllerEventSource` and `ManagedInformerEventSource` updated to go through the same pooled path.
- Dynamic registration/deregistration of event sources correctly reuses/releases pooled informers, replaying existing cache state to newly added handlers when attaching to an already-running shared informer.

## Docs

Added a new "Sharing Informers Between Controllers (Informer Pool)" section to `docs/content/en/docs/documentation/eventing.md` describing sharing semantics, what counts toward informer identity, and how to select/override the pooling strategy.

## Tests

- Unit tests for `NonSharingInformerPool`, `DefaultInformerPool`, and `InformerClassifier` (equality/identity semantics).
- New integration tests under `informerpool`:
  - `basic` — two reconcilers sharing an informer for the same secondary resource type, parameterized to also run against `AlwaysNewInformerPool` for regression coverage.
  - `deregister` — verifies informer lifecycle when an event source is dynamically deregistered while shared.
  - `dynamic` — verifies dynamically-registered/static shared informers, including replay of existing cache state to a newly attached handler.
- Existing tests (`ControllerTest`, `EventSourceManagerTest`, `ReconciliationDispatcherTest`, `ControllerEventSourceTest`, `InformerEventSourceTest`, `MockKubernetesClient`, `StandaloneDependentResourceIT`, `WorkflowMultipleActivationIT`) updated for the new construction/wiring paths.

### DIFF
--- a/docs/content/en/docs/documentation/eventing.md
+++ b/docs/content/en/docs/documentation/eventing.md
@@ -347,3 +347,69 @@ for [primary resources](https://github.com/operator-framework/java-operator-sdk/
 See
 also [CaffeineBoundedItemStores](https://github.com/operator-framework/java-operator-sdk/blob/main/caffeine-bounded-cache-support/src/main/java/io/javaoperatorsdk/operator/processing/event/source/cache/CaffeineBoundedItemStores.java)
 for more details.
+
+### Sharing Informers Between Controllers (Informer Pool)
+
+{{% alert title="Experimental" color="warning" %}}
+Informer pooling is marked `@Experimental`: the feature itself is production ready, but its
+configuration API may still change in a non-backwards-compatible way.
+{{% /alert %}}
+
+By default JOSDK maintains an *informer pool* so that informers are **shared** across controllers
+and event sources. When several `InformerEventSource`s (whether belonging to different controllers,
+or dynamically registered at runtime) watch the same resource type with an equivalent configuration,
+they are all backed by a single underlying `SharedIndexInformer` instead of one informer each. This
+reduces memory usage and the number of watch connections opened against the API server — which
+matters in operators where many controllers watch the same secondary resource type (for example
+`ConfigMap` or `Secret`).
+
+Two event sources share an informer when their effective informer configuration matches on all of:
+
+- the `KubernetesClient` they watch through, compared by instance: normally every event source
+  resolves the operator's own client, but an event source watching another cluster brings its own
+  (see [multi-cluster](#informereventsource-multi-cluster-support)). Two separate client instances
+  never share an informer, not even when they connect to the same API server — they may differ in
+  credentials, impersonation or TLS material, and the informer keeps using the client it was created
+  from,
+- the resource type (or the group/version/kind for generic resources),
+- the watched namespace,
+- the label, field and shard selectors,
+- the configured [item store](#bounded-caches-for-informers).
+
+The `informerListLimit` is intentionally *not* part of this identity: if two otherwise-equivalent
+event sources request a different list limit, the existing informer is reused (a warning is logged
+and the first-configured limit is kept). Indexers are also not part of the identity: they are
+registered on the shared informer under a name qualified with the controller and event source that
+added them, so index names are private to an event source and cannot collide with those of another
+one. You keep looking indexes up by the name you registered, and the indexers of an event source are
+removed from the shared informer when it stops using it.
+
+The pool is reference-counted: the shared informer is created on first use and only stopped once the
+last event source using it is de-registered (or its controller stops). Dynamically registering an
+event source for a resource that is already backed by a running informer reuses that informer, and
+the initial state already in its cache is replayed to the newly added handler.
+
+#### Selecting the pooling strategy
+
+The strategy is provided by the
+[`InformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/InformerPool.java)
+configured on the `ConfigurationService`. Two implementations are available:
+
+- [`DefaultInformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/DefaultInformerPool.java)
+  (the default): shares informers as described above.
+- [`NonSharingInformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/NonSharingInformerPool.java):
+  never shares informers, creating a dedicated informer for every event source. Use this to opt out
+  of pooling and restore the pre-pooling behavior.
+
+You can override the strategy through the `ConfigurationService`:
+
+```java
+Operator operator = new Operator(overrider ->
+    overrider.withInformerPool(new NonSharingInformerPool()));
+```
+
+A custom strategy has to extend
+[`AbstractInformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/AbstractInformerPool.java),
+which is what `withInformerPool` accepts: it already creates the informers from an
+`InformerClassifier` and starts them, leaving the subclass to decide only whether and how they are
+shared. `InformerPool` itself is just the narrower contract that the event sources consume.
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -36,6 +36,7 @@ import io.fabric8.kubernetes.client.KubernetesClientBuilder;
 import io.fabric8.kubernetes.client.utils.KubernetesSerialization;
 import io.javaoperatorsdk.operator.api.monitoring.Metrics;
 import io.javaoperatorsdk.operator.api.reconciler.Context;
+import io.javaoperatorsdk.operator.api.reconciler.Experimental;
 import io.javaoperatorsdk.operator.api.reconciler.Reconciler;
 import io.javaoperatorsdk.operator.api.reconciler.dependent.DependentResourceFactory;
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependent;
@@ -43,6 +44,7 @@ import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDep
 import io.javaoperatorsdk.operator.processing.dependent.kubernetes.KubernetesDependentResourceConfig;
 import io.javaoperatorsdk.operator.processing.dependent.workflow.ManagedWorkflowFactory;
 import io.javaoperatorsdk.operator.processing.event.source.controller.ControllerEventSource;
+import io.javaoperatorsdk.operator.processing.event.source.informer.pool.InformerPool;
 
 /** An interface from which to retrieve configuration information. */
 public interface ConfigurationService {
@@ -476,4 +478,23 @@ public interface ConfigurationService {
   default boolean cloneSecondaryResourcesWhenGettingFromCache() {
     return false;
   }
+
+  /**
+   * The informer pool used to create and (when using the default, sharing pool) share the informers
+   * backing the event sources of all controllers managed by this {@code ConfigurationService}.
+   *
+   * <p><strong>Implementations must return the same instance on every call.</strong> The pool is
+   * effectively a per-{@code ConfigurationService} singleton: controllers share informers only if
+   * they resolve the same pool, and reference counting / informer shutdown are only correct if
+   * {@code getInformer} and {@code releaseInformer} operate on that same instance. This is
+   * intentionally not a {@code default} method, since a {@code default} could not cache the result
+   * and would hand out a fresh (unshared) pool on each call; {@link AbstractConfigurationService}
+   * provides a cached implementation backed by the default sharing pool.
+   *
+   * @return the informer pool for this configuration service
+   */
+  @Experimental(
+      "Only the configuration API around informer pooling could still change in a"
+          + " non-backwards-compatible way, the pooling itself is prod ready.")
+  InformerPool informerPool();
 }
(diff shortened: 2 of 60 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/NonSharingInformerPool.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public <R extends HasMetadata> SharedIndexInformer<R> getInformer(
+      String controllerName,
+      String name,
+      InformerClassifier<R> classifier,
+      KubernetesClient client) {
+    var informer = createInformer(classifier, client);
+    informers.put(new ClassifierWithName(controllerName, name, classifier), informer);
+    return informer;
+  }
```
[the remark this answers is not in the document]
csviri: This is not true, strange.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -56,18 +54,12 @@ public class InformerEventSource<R extends HasMetadata, P extends HasMetadata>
 
   public InformerEventSource(
       InformerEventSourceConfiguration<R> configuration, EventSourceContext<P> context) {
-    this(configuration, configuration.getKubernetesClient().orElse(context.getClient()));
+    this(configuration);
   }
```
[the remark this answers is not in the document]
csviri: I added deprecation

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/NonSharingInformerPool.java
```diff
@@ -0,0 +1,78 @@
...
```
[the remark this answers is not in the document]
csviri: Now we throw exception if this happens

csviri:
@SamBarker @k-wall I saw that you added explicit support in kroxy operator to share informers, in case you are interested and have the bandwidth PTAL if this makes sense for you; this should solve that problem out of the box. thank you!

k-wall:
> @SamBarker @k-wall I saw that you added explicit support in kroxy operator to share informers, in case you are interested and have the bandwidth PTAL if this makes sense for you; this should solve that problem out of the box. thank you!

Thank you for the heads up.  We'll take a look and feed back.

docs/content/en/docs/documentation/eventing.md
```diff
@@ -346,4 +346,57 @@ for [primary resources](https://github.com/operator-framework/java-operator-sdk/
...
+
+#### Selecting the pooling strategy
+
+The strategy is provided by the
+[`InformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/InformerPool.java)
+configured on the `ConfigurationService`. Two implementations are available:
+
+- [`DefaultInformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/DefaultInformerPool.java)
+  (the default): shares informers as described above.
+- [`AlwaysNewInformerPool`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/AlwaysNewInformerPool.java):
```
xstefank: Maybe `SeparateInformerPool`, I kind of find the name `AlwaysNewInformerPool` confusing.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/DefaultInformerPool.java
```diff
@@ -0,0 +1,136 @@
...
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
+
+public class DefaultInformerPool extends AbstractInformerPool {
+
+  private static final Logger log = LoggerFactory.getLogger(DefaultInformerPool.class);
+
+  private final Map<InformerClassifier<?>, SharedIndexInformer<?>> informers = new HashMap<>();
+  private final Map<InformerClassifier<?>, AtomicInteger> counters = new HashMap<>();
```
xstefank: Maybe create a record for values so we don't need to have duplicated maps with the same key.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/NonSharingInformerPoolTest.java
```diff
@@ -0,0 +1,151 @@
...
+    assertThat(pool.size()).isEqualTo(1);
+    verify(client, times(2)).resources(TestCustomResource.class);
+  }
+
+  @Test
+  void releaseStopsAndRemovesTheInformer() {
+    var classifier = classifier("default");
+    var informer = pool.getInformer(CONTROLLER, ES_NAME, classifier, client);
+
+    var released = pool.releaseInformer(CONTROLLER, ES_NAME, classifier);
```
xstefank, line 111: maybe we could also have `pool.releaseInformer(informer)`

xstefank (approved):
approving, I have only opinioned enhancements. Feel free to comment on each and we can either do it in this PR or as followups.

docs/content/en/docs/documentation/eventing.md
csviri: What about `UnpooledInformerPool` ? :) `Separate` is vague imho

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/DefaultInformerPool.java
csviri: changed good point

docs/content/en/docs/documentation/eventing.md
csviri: Or `NonSharingInformerPool` ?!

csviri: I changed it to `NonSharingInformerPool`.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/NonSharingInformerPoolTest.java
csviri: I was thinking about this a lot that time, I would rather keep the contract simpler with just the current method; I marked it @Experimental ( well `@PublicEvolving` might be a better annotation) so we can enhance this in the future.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -476,4 +478,23 @@ default boolean useSSAToPatchPrimaryResource() {
...
+   * backing the event sources of all controllers managed by this {@code ConfigurationService}.
+   *
+   * <p><strong>Implementations must return the same instance on every call.</strong> The pool is
+   * effectively a per-{@code ConfigurationService} singleton: controllers share informers only if
+   * they resolve the same pool, and reference counting / informer shutdown are only correct if
+   * {@code getInformer} and {@code releaseInformer} operate on that same instance. This is
+   * intentionally not a {@code default} method, since a {@code default} could not cache the result
+   * and would hand out a fresh (unshared) pool on each call; {@link AbstractConfigurationService}
+   * provides a cached implementation backed by the default sharing pool.
+   *
```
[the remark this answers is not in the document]
csviri: I added default impl.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -90,9 +93,14 @@ protected InformerConfiguration(
     this.comparableResourceVersions = comparableResourceVersions;
   }
 
-  private InformerConfiguration(Class<R> resourceClass) {
+  private InformerConfiguration(Class<R> resourceClass, GroupVersionKind resourceGroupVersionKind) {
     this.resourceClass = resourceClass;
+    this.resourceGroupVersionKind = resourceGroupVersionKind;
```
metacosm, line 98: This seems like an either/or type of situation where either we have a resource class or we pass a GVK, no? The GVK should always be present and inferred (I think it should be possible to do so) from the resource class if that is what is provided.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
```diff
@@ -305,6 +320,10 @@ public Long getInformerListLimit() {
     return informerListLimit;
   }
 
+  public GroupVersionKind getResourceGroupVersionKind() {
```
metacosm, line 323: This can be `null` with the current implementation, which it probably shouldn't. Either make sure that the GVK is always set or return an `Optional` to convey the fact that it can be unset.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
```diff
@@ -176,6 +180,21 @@ public ConfigurationServiceOverrider withCloneSecondaryResourcesWhenGettingFromC
...
+   * Overrides the informer pool strategy used to create/share the informers backing the event
+   * sources. When not set, the default (informer-sharing) pool is used.
+   *
+   * <p>Custom strategies extend {@link AbstractInformerPool}, which already takes care of creating
+   * and starting the informers.
+   */
+  @Experimental(
+      "Only the configuration API around informer pooling could still change in a"
+          + " non-backwards-compatible way, the pooling itself is prod ready.")
+  public ConfigurationServiceOverrider withInformerPool(AbstractInformerPool informerPool) {
```
metacosm, line 193: The parameter should be `InformerPool`, not `AbstractInformerPool`.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerEventSourceConfiguration.java
```diff
@@ -76,6 +76,7 @@ default boolean followControllerNamespaceChanges() {
 
   <P extends HasMetadata> PrimaryToSecondaryMapper<P> getPrimaryToSecondaryMapper();
 
+  // todo deprecate
```
metacosm, line 79: Why?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationServiceOverrider.java
csviri: Good catch, will fix this in a subsequent PR.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerEventSourceConfiguration.java
csviri: This same as ResourceClass, should be part of InformerConfiguration not the event source configuraiton

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerConfiguration.java
csviri: Since this was an afterthought (I mean the whole GenericKubernetesResource support), I wanted make this consistent with the resourceClass provided in this class too. 

Maybe we could have both optional and deprecate the old methods. Well just remove this on next.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/InformerEventSourceConfiguration.java
csviri: Hmm, checking, but we don't use Optional on other values there either, so would just stick with this for now, maybe fix those in v6?

csviri:
@metacosm addressed comments in this PR: https://github.com/operator-framework/java-operator-sdk/pull/3541

PR #3333, opened 2026-05-05
docs: improve metrics and health probes docs

csviri:
- making grafana dashboard, Exploring metrics end-to-end section heading to lvl 3 (###), so it is more visible
- removing runtime info from metrics page 


xstefank (approved):
3333!

PR #3336, opened 2026-05-07
improve: use the default list methods with read-cache-after-write consistency

csviri:
Remove the index and list related methods "strong consistency variant".

The rational behind this change is that the user either uses read-cache-after-write consistency feature,
than the "strong read" should be used. If on the other hand tha feature,
is not used, the current implementation does not impose any overhead.

In addition to that strong consistency is not a good name, we call this feature
as read-cache-after-write consistency, that might be too long though.

Adds ghost resource handling for `keys()`.

Add also missing javadocs.


xstefank (approved):
Do we have a policy that public methods should be first deprecated before they are removed?

csviri:
> Do we have a policy that public methods should be first deprecated before they are removed?

This is against `next` branch and code that is not released yet. Otherwise we deprecate methods, and remove them in the next major version.

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3345

PR #3343, opened 2026-05-12
chore: fabric8 client 7.7.0

csviri:


csviri:
@metacosm @xstefank pls let me know if you would rather target main and add this as a patch release

metacosm (approved):
Should be OK for the next patch release, imo.

xstefank (approved):

csviri:
> Should be OK for the next patch release, imo.

kk will target `main` in that case. thx!!

PR #3345, opened 2026-05-12
read-cache-after-write aware list and byIndex

csviri:
Now `list()` and `byIndex()` of ManagedInformerEventSource (so also subsclasses) are aware of cached resources for read-cache-after-write consistency. 

Adds ghost resource handling for `keys()` and missing javadocs.

Signed-off-by: Attila Mészáros [a_meszaros@apple.com](mailto:a_meszaros@apple.com)

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -353,7 +353,10 @@ public abstract class ManagedInformerEventSource<
       String namespace, Predicate<R> predicate) {
     return r -> {
       if (namespace != null) {
-        var res = Optional.of(r).map(ns -> ns.equals(namespace)).orElse(false);
+        var res =
+            Optional.of(r)
+                .map(rr -> namespace.equals(rr.getMetadata().getNamespace()))
+                .orElse(false);
         if (!res) return false;
       }
       if (predicate != null) {
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -387,7 +387,8 @@ public abstract class ManagedInformerEventSource<
     }
     var managerKeys = manager().keys().collect(Collectors.toSet());
     var tempKeys = temporaryResourceCache.getResources().keySet();
-    return Stream.concat(managerKeys.stream(), tempKeys.stream().filter(k -> !managerKeys.contains(k)));
+    return Stream.concat(
+        managerKeys.stream(), tempKeys.stream().filter(k -> !managerKeys.contains(k)));
   }
 
   @Override
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -385,8 +385,9 @@ public abstract class ManagedInformerEventSource<
     if (!comparableResourceVersions || temporaryResourceCache.isEmpty()) {
       return manager().keys();
     }
+    var managerKeys = manager().keys().collect(Collectors.toSet());
     var tempKeys = temporaryResourceCache.getResources().keySet();
-    return Stream.concat(manager().keys(), tempKeys.stream().filter(k -> !manager().contains(k)));
+    return Stream.concat(managerKeys.stream(), tempKeys.stream().filter(k -> !managerKeys.contains(k)));
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSourceTest.java
```diff
@@ -533,6 +541,222 @@ void informerStoppedHandlerShouldBeCalledWhenInformerStops() {
...
+    var original = testDeployment();
+    var newer = testDeployment();
+    newer.getMetadata().setResourceVersion("5");
+
+    when(temporaryResourceCache.getResources())
+        .thenReturn(new HashMap<>(Map.of(ResourceID.fromResource(original), newer)));
+
+    var mim = mock(InformerManager.class);
+    when(mim.list(nullable(String.class), any())).thenReturn(Stream.of(original));
+    doReturn(mim).when(informerEventSource).manager();
```
metacosm: Why the different mocking style?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -275,4 +275,12 @@ private void checkGhostResources() {
   public synchronized Optional<T> getResourceFromCache(ResourceID resourceID) {
     return Optional.ofNullable(cache.get(resourceID));
   }
+
+  synchronized boolean isEmpty() {
+    return cache.isEmpty();
+  }
+
+  synchronized Map<ResourceID, T> getResources() {
```
metacosm, line 274: Shouldn't we return an immutable version here instead of a mutable copy?

xstefank (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -239,19 +235,161 @@ public void addIndexers(Map<String, Function<R, List<String>>> indexers) {
...
+   * {@inheritDoc}
+   *
+   * <p>This implementation is read-cache-after-write consistent. Results are merged with the
+   * temporary resource cache to ensure recently written resources are reflected in the output.
+   */
   @Override
-  public List<R> byIndex(String indexName, String indexKey) {
-    return manager().byIndex(indexName, indexKey);
+  public Stream<R> list(String namespace, Predicate<R> predicate) {
+    return mergeWithTempCacheForList(manager().list(namespace), namespace, predicate);
```
metacosm, line 243: This looks a little complex… wouldn't there be a way to simplify the `mergeWithTempCacheForList` method in such a way that the other `list` method could call this one?
Another aspect is: shouldn't the predicate be used in the call to `manager().list` as well here? This might simplify the rest of the code and make it go through less items?

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: This is intentional since later we remove elements from that map.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
csviri: > Another aspect is: shouldn't the predicate be used in the call to manager().list as well here? This might simplify the rest of the code and make it go through less items?

Unfortunatelly that is not possible, since we want to apply the predicate on the up-to-date list.

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSourceTest.java
csviri: fixed, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
csviri: > This looks a little complex… wouldn't there be a way to simplify the mergeWithTempCacheForList method in such a way that the other list method could call this one?

Not sure if that would be cleaner. Would rather rely on underlying APIs from manager.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
metacosm: I think it'd be better to let the client code make that copy, then so that it's explicit that it's working out of a copy, without having to look at the `getResources` code. Otherwise, we might think that we're directly modifying the cache itself.

csviri: In that case I would return it as unmodifiable map, and wrap it on reader, this would be a nicer API for sure. I just did not bother since it is an internal API (and unittested). But will do it, if this caught your eye.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
metacosm: OK, was just wondering. Thanks!

csviri:
@metacosm can we merge this or you need more time to review?

metacosm (commented):
LGTM

PR #3361, opened 2026-05-18
fix: revert breaking constructor change in InformerConfiguration

xstefank:
-


csviri:
Do we really need this, also on protected constructor? 
what is the main reason behind this?

xstefank:
It breaks QOSDK. And it is a public API. Users can also extend it and call the constructor. Not possible to do this in micro release.

csviri (approved):

PR #3373, opened 2026-05-22
feat(core): add by-name secondary resource lookup on Context

### LINKED ISSUE #3372: Typed by-name secondary resource accessors on Context
## Summary

Revisit #1410 ("Access secondary resources by ResourceID") now that `ResourceDiscriminators` are gone. Add typed by-name accessors on `Context<P>` so reconcilers can look up secondary resources scoped to a specific event source without dropping to `eventSourceRetriever().getEventSourceFor(...)` and casting.

## Motivation

When a reconciler watches external secondary resources via a named event source (e.g. `ConfigMap`s referenced from a custom resource's spec), today's options on `Context<P>` are:

1. `getSecondaryResource(Class, String eventSourceName)` returns the single linked resource. No by-name lookup.
2. `getSecondaryResourcesAsStream(Class, boolean deduplicate)` returns all secondary resources of the type across all event sources. No event-source scoping.
3. Drop to `context.eventSourceRetriever().getEventSourceFor(...)`, cast to `InformerEventSource`, then call `.get(ResourceID)`. Works, but ad-hoc.

#1410 acknowledged this back in 2022 and the chosen mechanism (`ResourceDiscriminators`) was later removed in #2253, leaving the gap unfilled.

With PR #3345 having landed (`Cache<T>` / `ResourceCache<T>` interfaces, read-cache-after-write consistent), the natural fix is to expose typed accessors on `Context<P>` that dispatch to those interfaces.

## Proposed API additions on `Context<P>`

```java
<R> Optional<R> getSecondaryResource(
        Class<R> expectedType, String eventSourceName, String name, String namespace);

default <R> Optional<R> getSecondaryResource(
        Class<R> expectedType, String eventSourceName, String name) {
    return getSecondaryResource(
            expectedType, eventSourceName, name,
            getPrimaryResource().getMetadata().getNamespace());
}

<R> Stream<R> getSecondaryResourcesAsStream(Class<R> expectedType, String eventSourceName);
```

Semantics:

- **Single-resource by name and namespace**: dispatch to `Cache<R>.get(ResourceID)` when the event source implements `Cache<R>` (direct cache lookup, read-cache-after-write consistent). Otherwise fall back to `EventSource.getSecondaryResources(primary)` filtered by `ResourceID`. `ResourceID` stays an implementation detail.
- **Single-resource by name** (default method): convenience that uses the primary resource's namespace. Documented caveat: for cluster-scoped primaries (no namespace) the lookup is cluster-scoped; use the four-arg overload to target a specific namespace.
- **Stream by event source**: dispatch to `ResourceCache<R>.list(namespace)` when the event source implements `ResourceCache<R>` (read-cache-after-write consistent). For cluster-scoped primaries (null namespace) falls back to `Cache.list()` to avoid an NPE in `InformerWrapper.list`. Otherwise falls back to `eventSource.getSecondaryResources(primary).stream()`.
- `NoEventSourceForClassException` handling mirrors the existing `getSecondaryResource(Class, String)`: rethrow unless `eventSourceName == null` and the type is part of a managed workflow, in which case return an empty result.

Validated end-to-end against a real consumer; PR includes the API + tests, downstream migration verified separately (119 ITs passing).

afalhambra-hivemq:
Adds three typed by-name secondary-resource accessors to `Context<P>`:

- `<R extends HasMetadata> Optional<R> getSecondaryResource(Class<R>, String eventSourceName, String name, String namespace)`
- `<R extends HasMetadata> Optional<R> getSecondaryResource(Class<R>, String eventSourceName, String name)` (default; uses the primary's namespace)
- `<R extends HasMetadata> Stream<R> getSecondaryResourcesAsStream(Class<R>, String eventSourceName)`

Closes #3372

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
@@ -114,6 +114,89 @@ public interface Context<P extends HasMetadata> {
 
   <R> Optional<R> getSecondaryResource(Class<R> expectedType, String eventSourceName);
 
+  /**
+   * Retrieves a specific secondary resource by name and namespace from the event source identified
+   * by the given name.
+   *
+   * <p>This is a typed convenience over manually retrieving the {@link
+   * io.javaoperatorsdk.operator.processing.event.source.EventSource} and calling its cache. When
+   * the underlying event source implements {@link
+   * io.javaoperatorsdk.operator.processing.event.source.Cache}, the lookup is a direct cache lookup
+   * and read-cache-after-write consistent.
+   *
+   * <p>{@code eventSourceName} may be {@code null}. When {@code null} and {@code expectedType} is
+   * part of a managed workflow whose activation condition may not have registered the event source,
+   * an empty {@link Optional} is returned instead of throwing {@link
+   * io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException}.
+   *
+   * @param expectedType the class representing the type of secondary resource to retrieve
+   * @param eventSourceName the name of the event source to look in (may be {@code null})
+   * @param name the name of the secondary resource
+   * @param namespace the namespace of the secondary resource (may be {@code null} for
+   *     cluster-scoped resources)
+   * @param <R> the type of secondary resource to retrieve
+   * @return an {@link Optional} containing the matching secondary resource, or {@link
+   *     Optional#empty()} if none matches
+   * @throws io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException if no event
+   *     source is registered for the given type and name (and no workflow activation condition
+   *     accounts for it)
+   * @since 5.4.0
+   */
+  <R extends HasMetadata> Optional<R> getSecondaryResource(
+      Class<R> expectedType, String eventSourceName, String name, String namespace);
+
+  /**
+   * Convenience overload of {@link #getSecondaryResource(Class, String, String, String)} that uses
+   * the primary resource's namespace.
+   *
+   * <p>If the primary resource is cluster-scoped (no namespace), the lookup is performed against
+   * the cluster scope. To target a specific namespace from a cluster-scoped primary, use {@link
+   * #getSecondaryResource(Class, String, String, String)} directly.
+   *
+   * <p>{@code eventSourceName} may be {@code null} with the same semantics as in {@link
+   * #getSecondaryResource(Class, String, String, String)}.
+   *
+   * @param expectedType the class representing the type of secondary resource to retrieve
+   * @param eventSourceName the name of the event source to look in (may be {@code null})
+   * @param name the name of the secondary resource (namespace inferred from the primary)
+   * @param <R> the type of secondary resource to retrieve
+   * @return an {@link Optional} containing the matching secondary resource, or {@link
+   *     Optional#empty()} if none matches
+   * @since 5.4.0
+   */
+  default <R extends HasMetadata> Optional<R> getSecondaryResource(
+      Class<R> expectedType, String eventSourceName, String name) {
+    return getSecondaryResource(
+        expectedType, eventSourceName, name, getPrimaryResource().getMetadata().getNamespace());
+  }
+
+  /**
+   * Retrieves a {@link Stream} of the secondary resources of the specified type from the event
+   * source identified by the given name. Useful when several event sources are registered for the
+   * same type and you need to scope retrieval to one of them, or when you want to apply a custom
+   * filter at the call site.
+   *
+   * <p>When the underlying event source implements {@link ResourceCache}, the stream is
+   * read-cache-after-write consistent.
+   *
+   * <p>{@code eventSourceName} may be {@code null} with the same semantics as in {@link
+   * #getSecondaryResource(Class, String, String, String)}: when {@code null} and {@code
+   * expectedType} is part of a managed workflow whose activation condition may not have registered
+   * the event source, an empty {@link Stream} is returned instead of throwing {@link
+   * io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException}.
+   *
+   * @param expectedType the class representing the type of secondary resources to retrieve
+   * @param eventSourceName the name of the event source to look in (may be {@code null})
+   * @param <R> the type of secondary resources to retrieve
+   * @return a {@link Stream} of secondary resources of the specified type
+   * @throws io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException if no event
+   *     source is registered for the given type and name (and no workflow activation condition
+   *     accounts for it)
+   * @since 5.4.0
+   */
+  <R extends HasMetadata> Stream<R> getSecondaryResourcesAsStream(
+      Class<R> expectedType, String eventSourceName);
+
   ControllerConfiguration<P> getControllerConfiguration();
 
   /**
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
@@ -36,6 +36,7 @@ import io.javaoperatorsdk.operator.processing.Controller;
 import io.javaoperatorsdk.operator.processing.event.EventSourceRetriever;
 import io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException;
 import io.javaoperatorsdk.operator.processing.event.ResourceID;
+import io.javaoperatorsdk.operator.processing.event.source.Cache;
 
 public class DefaultContext<P extends HasMetadata> implements Context<P> {
   private RetryInfo retryInfo;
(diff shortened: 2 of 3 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -114,6 +115,74 @@ default <R> Stream<R> getSecondaryResourcesAsStream(Class<R> expectedType) {
...
+   * @param expectedType the class representing the type of secondary resource to retrieve
+   * @param eventSourceName the name of the event source to look in
+   * @param resourceID the {@link ResourceID} identifying the secondary resource
+   * @param <R> the type of secondary resource to retrieve
+   * @return an {@link Optional} containing the matching secondary resource, or {@link
+   *     Optional#empty()} if none matches
+   * @throws io.javaoperatorsdk.operator.processing.event.NoEventSourceForClassException if no event
+   *     source is registered for the given type and name (and no workflow activation condition
+   *     accounts for it)
+   * @since 5.3.5
```
afalhambra-hivemq: Since this would be a new feature, I'm not sure whether this is something it will be released as part of a patch release or as a minor one.
So depending on it, this might change.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
```diff
@@ -114,6 +115,74 @@ default <R> Stream<R> getSecondaryResourcesAsStream(Class<R> expectedType) {
...
+   */
+  <R> Optional<R> getSecondaryResource(
+      Class<R> expectedType, String eventSourceName, ResourceID resourceID);
```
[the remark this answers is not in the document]
afalhambra-hivemq: I kept it as is to be in sync with the rest of the API contract in this class. The fallback already filters by `r instanceof HasMetadata` in the default implementation, so misuse silently returns Optional.empty() rather than crashing.

metacosm (approved):
LGTM

csviri (changes requested):
Thank you for the PR @afalhambra-hivemq !

Could you pls target 'next' branch so this ends up in next minor release rather than in next patch release. 

I was also thinking if we should use ResourceID in the method or simply name and namespace directly it might be less user boilerplate code at the end,or? What do you think?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
xstefank: As @csviri said, better for this to go in minor. So, please change this to 5.4.0

xstefank (approved):
LGTM, just please change the target branch and the `@since` tags. Thanks!

afalhambra-hivemq:
> Thank you for the PR @afalhambra-hivemq !
> 
> Could you pls target 'next' branch so this ends up in next minor release rather than in next patch release.
> 
> I was also thinking if we should use ResourceID in the method or simply name and namespace directly it might be less user boilerplate code at the end,or? What do you think?

Agree, cleaner and less boilerplate for customers.

Also targeting `next` branch, and changed `@since` to `5.4.0` as suggested.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
afalhambra-hivemq: Done

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
```diff
@@ -120,19 +135,54 @@ public <T> Optional<T> getSecondaryResource(Class<T> expectedType, String eventS
...
+      Class<R> expectedType, String eventSourceName, String name, String namespace) {
+    try {
+      final var eventSource =
+          controller.getEventSourceManager().getEventSourceFor(expectedType, eventSourceName);
+      final var resourceID = new ResourceID(name, namespace);
+      if (eventSource instanceof Cache<?> cache) {
+        return cache.get(resourceID).map(expectedType::cast);
+      }
+      return eventSource.getSecondaryResources(primaryResource).stream()
+          .filter(HasMetadata.class::isInstance)
```
csviri: Note that not all resources are Kubernetes resources, so here we might get a `ClassCastException`. Please put this into javadoc that this method is limited to Kubernetes resources and/or we might want to enforce that `R` extends `HasMetadata`.
Would it be worth to introduce a dedicated exception for that situation?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
```diff
@@ -95,6 +96,20 @@ public <R> Stream<R> getSecondaryResourcesAsStream(Class<R> expectedType, boolea
...
+   * should return an empty result instead of propagating the {@link
+   * NoEventSourceForClassException}.
+   *
+   * <p>If a workflow has an activation condition there can be event sources which are only
+   * registered if the activation condition holds, but to provide a consistent API we return an
+   * empty result instead of throwing an exception. Note that not only the resource which has an
+   * activation condition might not be registered but dependents which depend on it.
+   */
+  private boolean isMissingEventSourceExpected(String eventSourceName, Class<?> expectedType) {
+    return eventSourceName == null && controller.workflowContainsDependentForType(expectedType);
```
csviri, line 110: I'm not sure if this is the right thing to do, if there is no event source for a type, and user askes for that type, wouldn't be rather throw an exception? It is probably a issue with the user's code and would be nice to indicate that explicitly with an Exception. WDYT?

csviri (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
```diff
@@ -120,19 +135,54 @@ public <T> Optional<T> getSecondaryResource(Class<T> expectedType, String eventS
...
         return Optional.empty();
-      } else {
-        throw e;
       }
+      throw e;
+    }
+  }
+
+  @Override
+  public <R> Optional<R> getSecondaryResource(
```
csviri: Alternatively we could have this as `<R extends HasMetadata>`. See my comment below.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
afalhambra-hivemq: This is pre-existing behavior, inherited from [`getSecondaryResource(Class, String)`](https://github.com/operator-framework/java-operator-sdk/blob/492744ef3d8d4793f8bb2b1c6ed71ac62b6d59cb/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java#L115-L137):

```java
  } catch (NoEventSourceForClassException e) {                        
    /*                                                                                                                                                                                       
     * If a workflow has an activation condition there can be event sources which are only
     * registered if the activation condition holds, but to provide a consistent API we return an                                                                                            
     * Optional instead of throwing an exception.                     
     *                                                                                                                                                                                       
     * Note that not only the resource which has an activation condition might not be registered
     * but dependents which depend on it.                                                                                                                                                    
     */                                                               
    if (eventSourceName == null && controller.workflowContainsDependentForType(expectedType)) {
      return Optional.empty();                                                                                                                                                               
    } else {
      throw e;                                                                                                                                                                               
    }                                                                 
  }
```

The original commit comment explains the rationale: workflows with activation conditions may not have the event source registered yet, so the swallow keeps the API consistent.

What I did here was just extract that exact logic into `isMissingEventSourceExpected(...)` so the three try/catch blocks (existing method + two new ones) wouldn't duplicate the same condition and comment.
Nothing else.

I don't think we should change behavior here. It would either break existing callers or create inconsistency.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
afalhambra-hivemq: Yeah, I know, I just wanted to keep the contract in the Context interface aligned with the rest of the methods signature here where there is no such constraint in the returned type.

We would be breaking this pattern if we limit the resources to be `<R extends HasMetadata>` here (some will have this and some not). For example, there are more getSecondaryResources methods which do not have this constraint.

Otherwise, I can handle it with a dedicated `IllegalArgumentException` exception?

Let me know which one.

csviri: I think type safety is better choice here, since this method is always for Kubernetes resources

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
csviri: Ahh, I see, makes sense. Ok!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/DefaultContext.java
afalhambra-hivemq: Done - all three new methods are type safe now.

csviri (approved):
LGTM thank you @afalhambra-hivemq

PR #3377, opened 2026-05-26
improve: getSecondaryResourcesAsStream(expectedType,eventSourceName) does not have to be Kubernetes resources

csviri:
In a previous PR we added overloaded functions for accessing resources from event sources. Some of those were limited to Kubernetes resources. I've just realized that this one should not be one of them.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/Context.java
@@ -194,8 +194,7 @@ public interface Context<P extends HasMetadata> {
    *     accounts for it)
    * @since 5.4.0
    */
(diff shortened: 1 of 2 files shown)


csviri:
@afalhambra-hivemq FYI just a small adjustment for this one method. Pls let us know if it makes sense also for you.

afalhambra-hivemq:
> @afalhambra-hivemq FYI just a small adjustment for this one method. Pls let us know if it makes sense also for you.

@csviri, yes, makes sense to me. The stream method's fallback doesn't do anything HasMetadata-specific, and the fast path's constraint comes from ResourceCache itself anyway. No impact on us.

afalhambra-hivemq (approved):
LGTM! thanks!

metacosm (approved):

PR #3380, opened 2026-05-26
improve: reconciliation counts as retry attempt only if close to retry deadline

csviri:
Currently there is already a retry happening, and we receive an event that triggers the reconciliation counts as a retry. But this way we can fairly easily exhaust retries. So algorithm should is changed a reconciliation counts as a retry only if it is close enough to a scheduled retry.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
@@ -49,6 +49,13 @@ public class EventProcessor<P extends HasMetadata> implements EventHandler, Life
   private static final Logger log = LoggerFactory.getLogger(EventProcessor.class);
   private static final long MINIMAL_RATE_LIMIT_RESCHEDULE_DURATION = 50;
 
+  /**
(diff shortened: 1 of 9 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -49,6 +49,13 @@ public class EventProcessor<P extends HasMetadata> implements EventHandler, Life
   private static final Logger log = LoggerFactory.getLogger(EventProcessor.class);
   private static final long MINIMAL_RATE_LIMIT_RESCHEDULE_DURATION = 50;
 
+  /**
+   * Threshold below which an event-driven failed reconciliation that lands inside the current retry
+   * window is allowed to consume a retry attempt (i.e. advance the retry counter). Above this
+   * threshold the existing retry deadline is preserved instead.
+   */
+  private static final long RETRY_DEADLINE_PRESERVE_THRESHOLD_MILLIS = 5_000;
```
csviri, line 57: It is up to debate if this should be configurable, but maybe we can for now stick KISS principle

xstefank (approved):

PR #3388, opened 2026-05-28
improve: extend SecondaryToPrimaryMapper so it also gets the old resource

csviri:
This might needed in some corner cases where might help with optimizations to which resource to trigger.

TODO adaptor

### DIFF
--- a/docs/content/en/docs/documentation/eventing.md
+++ b/docs/content/en/docs/documentation/eventing.md
@@ -139,6 +139,25 @@ rare corner cases. Returning an empty set means that the mapper considered the s
 resource event as irrelevant and the SDK will thus not trigger a reconciliation of the primary
 resource in that situation.
 
+`SecondaryToPrimaryMapper` exposes two methods:
+
+- `toPrimaryResourceIDs(R resource)` — the original mapper. Implementing it is sufficient for
+  the vast majority of use cases.
+- `toPrimaryResourceIDs(R newResource, R oldResource)` — a variant that is the one actually
+  invoked by the SDK on every secondary event. Its default implementation delegates to the
+  single-argument method, so existing mappers keep working unchanged.
+
+Override the two-argument variant only in edge cases where the set of primary resources to
+reconcile depends on what changed between the previous and the new version of the secondary
+resource (e.g. a reference that moved from one primary to another, where both primaries need
+to be reconciled). **Use it with caution:** `oldResource` is sourced from the informer cache and
+is only populated for genuine update events observed while the controller is already running.
+On controller startup the cache is empty, so the initial events received for resources that
+already exist in the cluster are delivered as adds with `oldResource == null` — even if those
+resources had been updated before the operator came up. `oldResource` is also `null` for delete
+events and for events triggered through the primary-to-secondary index. Implementations must
+therefore handle a `null` `oldResource` gracefully.
+
 Adding a `SecondaryToPrimaryMapper` is typically sufficient when there is a one-to-many relationship
 between primary and secondary resources. The secondary resources can be mapped to its primary
 owner, and this is enough information to also get these secondary resources from the `Context`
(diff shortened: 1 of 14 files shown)


csviri:
see also discussio here: https://discord.com/channels/723455000604573736/723455000604573739/1508830486913482913

csviri:
as discussed with @dvob , he will created a detailed issue for this.

dvob:
@csviri thank you for implementing this so quickly. I also created issue #3401 for reference which we can more or less close right away if this is merged.

I updated my reproducer to use two different types (`MyThing` and `MyThingConfig`), in the hope that it is more understandable. I also updated it to use this PR and it did work (test does no longer fail) https://github.com/dvob/java-operator-sdk-repo-example/tree/csviri-secondary-to-primary-improve

However, when updating my reproducer, I ran into a different challenge. In my [initial reproducer](https://github.com/dvob/java-operator-sdk-repo-example/tree/791beb6f23c6e17a38a2ff7da92f5ecd786df4a0) I used only one CRD (`Repo`) which referenced itself. But now I have to two different CRDs and in `withPrimaryToSecondaryMapper` I basically want to access the `MyThingConfig` informer which I'm building which is not easily possible. For now I did some dirty hack. But you might have a better idea to achieve this: https://github.com/dvob/java-operator-sdk-repo-example/blob/4b9e01c3099ae580036aa81cbe1558c0d422f0b0/src/main/java/com/example/MyThingReconciler.java#L53 ?

One thing which also came to my mind is, that if you not specify `withPrimaryToSecondaryMapper` it will use the DefaultPrimaryToSecondaryIndex which most likely will behave wrong in such a scenario. This is not really a problem, as you just can set `withPrimaryToSecondaryMapper`, but I'm just no sure how obvious this is for users of the API.

csviri:
Yes, not sure if we can do better than to document it,see:

https://javaoperatorsdk.io/docs/documentation/working-with-es-caches/#the-use-case-for-primarytosecondarymapper

https://javaoperatorsdk.io/docs/documentation/eventing/#managing-relation-between-primary-and-secondary-resources

dvob:
Yes, I think docs are fine. 

For a long time i didn't even knew about the default reverse index behavior. I did set the primaryToSecondaryMapper every time explicitly and I still do this because I think its easier to understand if things are a little more explicit. I don't know how other users rely on this.

csviri:
>  I did set the primaryToSecondaryMapper every time explicitly and I still do this because I think its easier to understand if things are a little more explicit. I don't know how other users rely on this.

Yes, unfortunately this is not a that nice and round concept, but currently the only way to be able to use `getSecondaryResource` uniformly. We time to time get questions around it so it is used.

csviri:
>  basically want to access the MyThingConfig informer which I'm building which is not easily possible. For now I did some dirty hack. But you might have a better idea to achieve this: https://github.com/dvob/java-operator-sdk-repo-example/blob/4b9e01c3099ae580036aa81cbe1558c0d422f0b0/src/main/java/com/example/MyThingReconciler.java#L53 ?

Please take a look on indexers, setting a custom indexer for that would solve the problem as far I can see. InformerEventSource has API to add indexer:

https://github.com/operator-framework/java-operator-sdk/blob/064799e8c80717b70b0ec52a6875d8f8288486b4/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java#L227-L233

then access the resource by index:

https://github.com/operator-framework/java-operator-sdk/blob/064799e8c80717b70b0ec52a6875d8f8288486b4/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java#L276

csviri:
I will add an Integration test to showcase your sample handling

csviri:
replaced by: https://github.com/operator-framework/java-operator-sdk/pull/3425

csviri:
@dvob I implemented an alternative version with an integration test to cover your use case, if everything goes well, this will be in next minor release. Please take a look, and let me know if it looks good to you.

PR #3390, opened 2026-05-29
build: add SpotBugs plugin and fix all reported findings

csviri:
Add the SpotBugs Maven plugin (effort=Max, threshold=Medium) with a
`check` execution bound to the `verify` phase across all modules, plus a
documented `spotbugs-exclude.xml` for intentional/false-positive findings.

Real bug fixes:

- ExecutorServiceManager: synchronize start()/stop() so `started` and
  `configurationService` are accessed consistently with the already-locked
  lazyInitWorkflowExecutorService() (IS2_INCONSISTENT_SYNC,
  AT_STALE_THREAD_WRITE_OF_PRIMITIVE).
- AbstractEventSourceHolderDependentResource: make `isCacheFillerEventSource`
  volatile (written under lock, read from reconcile threads).
- ManagedInformerEventSource: make `cache` and `controllerConfiguration`
  volatile (assigned under lock in start(), read lock-free).
- AbstractOperatorExtension: the builder's `namespaceDeleteTimeout` was
  configurable but never plumbed into the extension, so the namespace
  deletion timeout always used the default. Threaded it through the
  constructors of both LocallyRunOperatorExtension and
  ClusterDeployedOperatorExtension.
- MicrometerMetrics.addMetadataTags(): guard against null `metadata`, which
  incrementCounter() already tolerates (latent NPE, NP_NULL_PARAM_DEREF).
- Bootstrapper.addTemplatedFile(): close the Writer via try-with-resources
  and write as UTF-8 (OBL_UNSATISFIED_OBLIGATION, DM_DEFAULT_ENCODING).

Default-encoding fixes (DM_DEFAULT_ENCODING), all pinned to UTF-8:

- LocallyRunOperatorExtension CRD apply/delete byte conversions.
- AccumulativeMappingWriter reader and PrintWriter.
- ClassMappingProvider InputStreamReader.
- sample mysql-schema Secret/Schema Base64 encode/decode.
- sample tomcat-operator WebappReconciler ByteArrayOutputStream.toString().

Sample cleanups:

- TomcatReconciler: avoid unbox-then-rebox in a log argument
  (BX_UNBOXING_IMMEDIATELY_REBOXED).
- OperationsSampleOperator: make the config path overridable via the
  CONFIG_PATH env var instead of a hardcoded absolute path
  (DMI_HARDCODED_ABSOLUTE_FILENAME).

Suppressed in spotbugs-exclude.xml (intentional / false positives):
EI_EXPOSE_REP(2) and CT_CONSTRUCTOR_THROW project-wide, plus DM_EXIT in
LeaderElectionManager, NP_BOOLEAN_RETURN_NULL in BooleanWithUndefined,
EQ_DOESNT_OVERRIDE_EQUALS in GroupVersionKindPlural, and
SING_SINGLETON_HAS_NONPRIVATE_CONSTRUCTOR in
SSABasedGenericKubernetesResourceMatcher and ConfigLoader.


sample-operators/mysql-schema/src/main/java/io/javaoperatorsdk/operator/sample/dependent/SchemaDependentResource.java
```diff
@@ -121,7 +122,8 @@ public void delete(MySQLSchema primary, Context<MySQLSchema> context) {
   }
 
   public static String decode(String value) {
-    return new String(Base64.getDecoder().decode(value.getBytes()));
+    return new String(
+        Base64.getDecoder().decode(value.getBytes(StandardCharsets.UTF_8)), StandardCharsets.UTF_8);
   }
```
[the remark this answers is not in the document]
csviri: fixed

metacosm (approved):

sample-operators/operations/src/main/java/io/javaoperatorsdk/operator/sample/operations/OperationsSampleOperator.java
```diff
@@ -66,7 +66,8 @@ public static void main(String[] args) throws Exception {
 
     var configProviders = new ArrayList<ConfigProvider>();
     configProviders.add(new EnvVarConfigProvider());
-    configProviders.add(new YamlConfigProvider(Path.of("/config/config.yaml")));
+    var configPath = System.getenv().getOrDefault("CONFIG_PATH", "/config/config.yaml");
```
xstefank: this can be a constant and also if it is an environment variable, maybe we should call it JOSDK_CONFIG_PATH

xstefank (approved):

sample-operators/operations/src/main/java/io/javaoperatorsdk/operator/sample/operations/OperationsSampleOperator.java
csviri: This is just local for this test implementation, but I will change it. thx!!

PR #3394, opened 2026-06-04
fix: deletion not reconciled when merged into eventFilteringUpdate response

### LINKED ISSUE #1353: Operator intermittently never processes resource deletion — informer misses events, finalizer stays forever
## Environment

- Java Operator SDK: 5.3.2
- Quarkus Operator SDK (quarkus-operator-sdk): 7.7.3
- Quarkus: 3.33.1
- fabric8 kubernetes-client: 7.5.2 (also reproduced with 7.7-SNAPSHOT built from main branch, and the newest JOSDK version)
- Kubernetes: K3s via TestContainers
- Java: 17

## Description

We have a Kubernetes operator managing custom resources with a finalizer. Intermittently (~30-40% of test runs), when a resource is deleted, the operator never processes the deletion. The resource gets stuck indefinitely with `deletionTimestamp` set and the finalizer still present. The reconciler is never called for the deletion.

Our test suite runs ~170 tests, each creating resources in random namespaces. The informer creates namespace-scoped watches dynamically.

This started to happen once we upgraded JOSDK:

```
JOSDK:             5.2.3  -> 5.3.2 
fabric8 :  7.4.0  -> 7.5.2  (updated alongside JOSDK)
```

## What happens

1. Resource is created and reconciled successfully (reaches READY status)
2. Resource is deleted via the Kubernetes API
3. Kubernetes sets `deletionTimestamp` on the resource (confirmed via API server)
4. The operator **never receives the delete event** — no reconciliation with `deletionTimestamp` appears in logs
5. Since the reconciler never runs for the deletion, the finalizer is never removed
6. The resource stays stuck forever with `deletionTimestamp` + finalizer

## Reproduction rate

- 30-40% of test runs (4/10 in one batch, 3/10 in another)
- Tested with fabric8 7.5.2 and 7.7-SNAPSHOT (main branch) — same behavior on both
- Tends to correlate with tests that trigger informer reconnection (e.g., tests that delete and recreate dependent resources, causing the Reflector to re-list)

## Key observations from logs

**When deletion works correctly**, the informer delivers the updated resource (with `deletionTimestamp` set) and JOSDK handles the finalizer removal:

```
DEBUG [EventProcessor] Event processing finished.
  Scope: ExecutionScope{resource=ResourceID{name='my-resource', namespace='abc123'},
    deleteEvent=false, isDeleteFinalStateUnknown=false},
  PostExecutionControl: PostExecutionControl{
    onlyFinalizerHandled=true,
    updatedCustomResource=CustomResource{
      kind='MyResource',
      deletionTimestamp=2026-04-24T14:13:32Z,
      finalizers=[],           <-- finalizer removed successfully
      resourceVersion=795
    }
  }
```

**When it fails**, the last thing the informer does for the affected namespace is a `Listing items` (re-list), after which no further events are ever received for the resource despite it being deleted via the API server:

```
DEBUG [Reflector] Listing items (1) for example.com/v1alpha1/namespaces/xyz789/myresources at v1197
```

The last reconciliation seen is a normal (non-delete) reconciliation where the resource has no `deletionTimestamp`:

```
INFO [MyReconciler] Reconcile executed for resource:
  kind='MyResource', name='my-resource', namespace='xyz789',
  deletionTimestamp=null,
  finalizers=[myresources.example.com/finalizer],
  status=READY
```

After the delete is issued, the `EventProcessor` never logs any event for this resource again. No `deletionTimestamp` is ever seen by the reconciler. The resource remains stuck with the finalizer until it is forcibly removed.

## Workaround

We work around this in test cleanup by catching stuck resources and forcibly removing finalizers via direct API patch:

```kotlin
// If resources are still present after 5s, strip finalizers and retry delete
val stuckResources = client.resources(kClass.java)
    .inAnyNamespace().list().items
    .filter { it.metadata.finalizers?.isNotEmpty() == true }
stuckResources.forEach { resource ->
    resource.metadata.finalizers = emptyList()
    client.resource(resource).patch()
}
```

## Related fabric8 issues

- https://github.com/fabric8io/kubernetes-client/issues/7435 — "Informer rsync functionality intermittently fails, no event re-entry" (same symptoms)
- https://github.com/fabric8io/kubernetes-client/pull/7537 — "generalizes onNothing to onList" (merged to main, but issue persists with 7.7-SNAPSHOT)

xstefank:
Should fix https://github.com/quarkiverse/quarkus-operator-sdk/issues/1353.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
@@ -97,6 +97,7 @@ public abstract class ManagedInformerEventSource<
   public R eventFilteringUpdateAndCacheResource(R resourceToUpdate, UnaryOperator<R> updateMethod) {
     ResourceID id = ResourceID.fromResource(resourceToUpdate);
     log.debug("Starting event filtering and caching update");
+    final boolean wasMarkedForDeletion = resourceToUpdate.isMarkedForDeletion();
     R updatedResource = null;
     try {
       temporaryResourceCache.startEventFilteringModify(id);
@@ -141,7 +142,23 @@ public abstract class ManagedInformerEventSource<
                     ? ((ResourceDeleteEvent) r).isDeletedFinalStateUnknown()
                     : null);
           },
-          () -> log.debug("No new event present after the filtering update"));
+          () -> {
+            log.debug("No new event present after the filtering update");
+            // If the resource was marked for deletion concurrently with this update, the deletion
+            // event arrived at a lower RV than the update result and was deferred then dropped by
+            // doneEventFilterModify (deferred RV < lastOwnUpdatedRV). The API server merges the
+            // deletionTimestamp into the update response, so updatedForLambda already carries it,
+            // but no event was propagated to the reconciler. Trigger one now so the finalizer can
+            // be removed.
+            if (updatedForLambda != null
+                && updatedForLambda.isMarkedForDeletion()
+                && !wasMarkedForDeletion) {
+              log.debug(
+                  "Resource {} was marked for deletion during update, triggering reconciliation",
+                  id);
+              handleEvent(ResourceAction.UPDATED, updatedForLambda, resourceToUpdate, null);
+            }
+          });
     }
   }
 
--- a/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/deletionduringstatusupdate/DeletionDuringStatusUpdateCustomResource.java
+++ b/operator-framework/src/test/java/io/javaoperatorsdk/operator/baseapi/deletionduringstatusupdate/DeletionDuringStatusUpdateCustomResource.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright Java Operator SDK Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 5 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/ManagedInformerEventSource.java
```diff
@@ -141,7 +142,23 @@ public R eventFilteringUpdateAndCacheResource(R resourceToUpdate, UnaryOperator<
...
+            // deletionTimestamp into the update response, so updatedForLambda already carries it,
+            // but no event was propagated to the reconciler. Trigger one now so the finalizer can
+            // be removed.
+            if (updatedForLambda != null
+                && updatedForLambda.isMarkedForDeletion()
+                && !wasMarkedForDeletion) {
+              log.debug(
+                  "Resource {} was marked for deletion during update, triggering reconciliation",
+                  id);
+              handleEvent(ResourceAction.UPDATED, updatedForLambda, resourceToUpdate, null);
```
metacosm, line 159: Shouldn't this be triggered with `ResourceAction.DELETED` instead?

csviri: No, this is not a delete event, for marked for deletion is an update event (setting that property in metadata)

csviri (changes requested):
@xstefank this would not work for example for server side apply:

https://github.com/operator-framework/java-operator-sdk/blob/f0f6924b0783ef69885761ea29cf1030c60cc8bf/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java#L99

where the resource always needs to a fresh resource.

I created a draft alternative:
https://github.com/operator-framework/java-operator-sdk/pull/3397

xstefank:
@csviri OK

PR #3414, opened 2026-06-11
improve: filter only own updates for read-after-write-consistency

### LINKED ISSUE #3398: Filtering ONLY own events in read-after-write-consitency
When doing updates with read-after-write-consistency we are now filtering events that happened before our update. This is causing mainly now a practical issue only for primary resource:

https://github.com/quarkiverse/quarkus-operator-sdk/issues/1353

however, for primary resource it could happen that while we update the status some changes the spec, and that we would like to reconcile. So for primary we should just filter only changes for our updates.  But similar issue(s) scenarios might happen for the secondary resources see below. 

At the end we will be on safe side to propagate intermediate events for changes that happen during our update. Thus only filter the event for our resource change, where the new resource version is the same as the resource version received after our update request.
Note that this is a bit harder than it seems, since relist might happen in informer for that we might need a small patch to fabric8 client, if that happens we need to propagate events also for our update since an event might be lost.

The problematic cases:

## Primary resource spec update during status update

When we update the status at the end of the reconciliation, we might receive an update just before for the spec, that spec change now will be filtered out, so reconciliation won't be triggered for it.

## Workaround

For now users can simply use the client to patch the status of the primary resource or update update secondary resources:

```java

context.getClient().resource(webPage).patchStatus();

```

That will not filter our the events.

Or downgrade to v `5.2.x`

##  Dependency between managed resources

Let's say we have a primary `P` that manages a resource `A` and there are other controllers on the cluster, that patches that resource A also. Assuming that in the controller we reconcile the resource A and that we should take into account properties of A that are managed by an other controller (think SSA). So in this case we have similar situation as with primary resource (above), that if we do an update on resource A and before that it just get updated (but after we calculated the target state) we don't get triggered by again for that update, however that is needed to get into desired state.

## Edge case with re-list

Note that if a a 3rd party did an update on the resource just before our update but informer was down and had to re-list we might not see that event. To fix this we would need and additional callback before re-list and propagate events for all changed resources since the web socket if informer was down until the most recent resource version from list.

## Notes

- Note that both of these scenarios might be handled currently by the user with some  (if user uses the the caching and filtering update directly not using `UpdateControl`) since user sees the up-to-date resource after updates. But for that user might require to check the resource after the update; what is quite strong requirement and it goes against the current mindset in the operator pattern.

csviri:
Reworks how the framework filters watch events caused by the controller's own writes. The previous "expect this single RV" approach couldn't handle concurrent writes, foreign updates between two of our writes, mid-window deletes,
  or relist gaps. It is replaced with a **per-resource event-filter window** driven by an explicit state machine.

  ### Components

  - **`TemporaryResourceCache` (TRC)** — caches the response of our own writes so the next read serves the up-to-date resource; also gates event delivery while a write is in flight.
  - **`EventFilterSupport`** — keeps one `EventFilterWindow` per `ResourceID`. Routes lifecycle calls (`startEventFilteringModify` / `doneEventFilterModify` / `processEvent`, relist start/finish, ghost cleanup) and disposes empty windows.
  - **`EventFilterWindow`** — the algorithm. Holds two RV-keyed sorted structures plus an `activeUpdates` counter and a `reListOnGoing` flag:
    - `ownResourceVersions` — RVs returned from our writes
    - `relatedEvents` — informer events for this resource, tagged `partOfReList` if relist is ongoing
  - **`GenericResourceEvent`** — richer event carrier (action, current, previous, `lastStateUnknown`, `partOfReList`) so the framework can synthesize realistic ADD/UPDATE/DELETE events.
  - **`ManagedInformerEventSource.eventFilteringUpdateAndCacheResource`** — the single funnel for caching writes (status patch, internal SSA, `context.resourceOperations()`). Opens the window, runs the update, records the response RV, asks the
  window what to surface on close.
  - **`InformerEventSource.handleEvent`** — the synthesized-event delivery point. Now also updates the secondary→primary index on `DELETED` so synthesized deletes don't leave stale tombstones.

  <details>
  <summary><b>Window decision rules (<code>EventFilterWindow.check</code>)</b></summary>

  1. No related events → nothing to surface.
  2. No active writes & no own RVs → flush every buffered event (no filtering contribution).
  3. DELETE arrived but no own RV yet → surface the DELETE; resource is gone.
  4. Own RVs ≤ last event RV → echoes for some prefix of own writes. If `events.keys() == ownRVs` and no relist tag, the window is a *pure own echo* and is filtered. Exception: a trailing `DELETED` is still surfaced (we wrote, then someone
  deleted).
  5. Foreign mixed in → synthesize a single `UPDATED` whose `previousResource` is the pre-window snapshot and `resource` is the latest known state. One reconciliation, faithful before/after.
  6. Relist in progress → events tagged `partOfReList` are never silently filtered; relist may hide events, so we err on the side of surfacing.
  </details>

  <details>
  <summary><b>Ghost cleanup</b></summary>

  Resources can sit in the TRC for which the informer never delivers any watch event (created then immediately deleted by a third party while the informer's watch was disconnected). `TemporaryResourceCache.checkGhostResources()`, triggered
  from `onList`, sweeps the TRC: anything stale relative to `lastSyncResourceVersion` and absent from the informer cache is removed and surfaced as a synthetic `DELETED` via `handleEvent` — which is precisely the path that needed
  `primaryToSecondaryIndex.onDelete`.
  </details>

  ### Sequence — own write + foreign update interleave

   ```mermaid
  sequenceDiagram
      autonumber
      participant R as Reconciler
      participant MES as ManagedInformerEventSource
      participant TRC as TemporaryResourceCache
      participant EFS as EventFilterSupport
      participant W as EventFilterWindow
      participant INF as Informer
      participant K8S as K8s API
      participant H as Reconcile queue

      R->>MES: serverSideApply secondary
      MES->>TRC: startEventFilteringModify id
      TRC->>EFS: startEventFilteringModify id
      EFS->>W: getOrCreate id, activeUpdates++
      MES->>K8S: SSA
      K8S-->>MES: response RV=100
      MES->>TRC: putResource rv=100
      TRC->>EFS: addToOwnResourceVersions id, 100

      Note over K8S: third party PATCH, RV=101
      K8S-->>INF: watch rv=101
      INF->>MES: onUpdate old, rv=101
      MES->>TRC: onAddOrUpdateEvent UPDATED, rv=101
      TRC->>EFS: processEvent id, ev
      EFS->>W: addRelatedEvent rv=101, check returns empty

      K8S-->>INF: watch rv=100 own echo
      INF->>MES: onUpdate old, rv=100
      MES->>TRC: onAddOrUpdateEvent UPDATED, rv=100
      TRC->>EFS: processEvent id, ev
      EFS->>W: addRelatedEvent rv=100, check returns empty

      MES->>TRC: doneEventFilterModify id
      TRC->>EFS: doneEventFilterModify id
      EFS->>W: activeUpdates--, check
      W-->>EFS: synthesized UPDATED, prev=pre-window, res=rv=101
      EFS-->>MES: GenericResourceEvent
      MES->>H: handleEvent UPDATED, rv=101, prev, null
      H->>R: reconcile, foreign change surfaces once
  ```

  The window collapses an own write (RV 100) and a foreign write (RV 101) into one synthesized `UPDATED` — instead of silencing the foreign change or firing twice (self-loop).

  ### Notes

  - Relist hooks (`setStartingReList` / `setRelistFinished`) are wired but currently inert in `ManagedInformerEventSource.onList` pending [fabric8/kubernetes-client#7899](https://github.com/fabric8io/kubernetes-client/pull/7899); that's why
  `OnRelistFilterIT` is `@Disabled`.
  - New ITs pinning the rules: `externalsecondaryupdate`, `externalupdateduringownupdate`, `deletionduringstatusupdate`, `filterpatchevent`, `readownupdates`, `ownsecondaryupdate`, `onrelistfilter`.
  - New unit tests: `EventFilterWindowTest`, `EventFilterSupportTest`, plus expansions to `InformerEventSourceTest` and `TemporaryResourceCacheTest`.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java
```diff
@@ -0,0 +1,214 @@
...
+
+  private GenericResourceEvent getLastRelatedEvent() {
+    return getLastRelatedEvent(relatedEvents);
+  }
+
+  public synchronized boolean canRemoved() {
+    if (activeUpdates == 0 && ownResourceVersions.isEmpty() && relatedEvents.isEmpty()) {
+      return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
csviri: fixed

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerEventSource.java
```diff
@@ -123,8 +122,15 @@ public synchronized void onDelete(R resource, boolean deletedFinalStateUnknown)
             log.debug(
                 "On delete event received. deletedFinalStateUnknown: {}", deletedFinalStateUnknown);
           }
+          var resultEvent =
+              temporaryResourceCache.onDeleteEvent(resource, deletedFinalStateUnknown);
+          if (resultEvent.isEmpty()) {
+            return;
+          }
```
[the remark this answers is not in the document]
csviri: this is intentional here.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -206,9 +176,6 @@ public synchronized void putResource(T newResource) {
       return;
```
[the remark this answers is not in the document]
csviri: not relevant

docs/content/en/blog/news/read-after-write-consistency.md
```diff
@@ -221,13 +226,38 @@ sequenceDiagram
...
-Note that the implementation of this is relatively complex, since while performing the update we want to record all the
-events received in the meantime and decide whether to propagate them further once the update request is complete.
 
-However, this way we significantly reduce the number of reconciliations, making the whole process much more efficient.  
+The framework runs a per-resource *event filter window* around each in-flight
+write: it records the resource version returned by our update, buffers any
+related events that arrive in the meantime, and at the end of the window
+decides what (if anything) to surface to the reconciler. The rules:
+
+- **Pure own echo**: if the only events in the window are watch events whose
```
metacosm, line 235: It would help to include the sequence diagram that is in the PR here as well… Actually, a sequence diagram for each of these cases would be nice but might be too much work.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/ControllerEventSource.java
```diff
@@ -84,7 +84,7 @@ protected synchronized void handleEvent(
     try {
       if (log.isDebugEnabled()) {
         log.debug("Event received with action: {}", action);
-        log.trace("Event Old resource: {},\n new resource: {}", oldResource, resource);
+        log.debug("Event Old resource: {},\n new resource: {}", oldResource, resource);
```
metacosm: Why was this moved to debug? Outputting resources is quite verbose and is probably better only for trace-level logging.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/GenericResourceEvent.java
```diff
@@ -24,26 +24,33 @@
 import io.javaoperatorsdk.operator.processing.event.source.controller.ResourceEvent;
 
 /** Used only for resource event filtering. */
-public class ExtendedResourceEvent extends ResourceEvent {
+public class GenericResourceEvent extends ResourceEvent {
```
metacosm: The name is not great… The previous one wasn't great either but renaming now doesn't make sense if the name is not more explicit.

metacosm: The class should also probably be package-only

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterSupport.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.javaoperatorsdk.operator.processing.event.ResourceID;
+
+public class EventFilterSupport {
```
metacosm: Class should be package-only.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/GenericResourceEvent.java
```diff
@@ -24,26 +24,33 @@
...
+    super(action, ResourceID.fromResource(latestResource), latestResource);
     this.previousResource = previousResource;
+    this.lastStateUnknow = lastStateUnknow;
   }
 
   public Optional<HasMetadata> getPreviousResource() {
     return Optional.ofNullable(previousResource);
   }
 
+  public Boolean getLastStateUnknow() {
```
metacosm: isLastStateUnknow would be better, imo.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -206,16 +186,19 @@ public synchronized void putResource(T newResource) {
...
       log.debug(
           "Temporarily moving ahead to target version {} for resource id: {}",
           newResource.getMetadata().getResourceVersion(),
           resourceId);
       cache.put(resourceId, newResource);
+    } else {
+      log.debug(
+          "Skipping temp cache put; new rv {} is not later than cached rv {}. id={}",
+          newResource.getMetadata().getResourceVersion(),
+          cachedResource.getMetadata().getResourceVersion(),
```
metacosm, line 200: This might cause an NPE because cachedResource might be null

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -231,7 +214,7 @@ private String getLastSyncResourceVersion(String namespace) {
    * explicitly add resources to this cache. Those are cleaned up by this check, which is triggered
    * by the informer's onList callback.
    */
-  public void checkGhostResources() {
+  public synchronized void checkGhostResources() {
```
metacosm, line 217: Won't this impact concurrency if the cache is big?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -272,6 +254,19 @@ synchronized boolean isEmpty() {
   }
 
   synchronized Map<ResourceID, T> getResources() {
-    return Collections.unmodifiableMap(cache);
+    return Map.copyOf(cache);
+  }
+
+  // for testing purposes
+  synchronized EventFilterSupport getEventFilterSupport() {
```
metacosm: There's no need for this method to be synchronized.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -272,6 +254,19 @@ synchronized boolean isEmpty() {
...
+    return eventFilteringSupport;
+  }
+
+  public synchronized void setOngoingRelist(String lastKnownSyncVersion) {
```
metacosm: No need to synchronize this method

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
```diff
@@ -272,6 +254,19 @@ synchronized boolean isEmpty() {
...
+    eventFilteringSupport.setStartingReList();
+  }
+
+  public synchronized void setRelistFinished(String syncResourceVersions) {
```
metacosm: No need to synchronize this method

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java
```diff
@@ -0,0 +1,233 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.javaoperatorsdk.operator.processing.event.source.ResourceAction;
+
+/**
+ * Contains all the relevant information around the eventing and algorithms of a single resources.
+ */
+class EventFilterWindow {
```
metacosm, line 56: This class needs documentation and comments. I'm also concerned about the impact of having a class with heavy fields per processed event… This might going to increase memory requirements and GC churn.

metacosm (changes requested):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/ControllerEventSource.java
csviri: I just reverted this, but nice catch thx!

docs/content/en/blog/news/read-after-write-consistency.md
csviri: Yeah, also that blog post I would like to keep short, so folks actually read it. The unit tests basically cover all the sequences, and there  is >20 of them, what is quite a lot.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/GenericResourceEvent.java
csviri: yeah, if you insist I can rename it back, wanted to make it package private but it is used also in `ControllerEventSource` - what is in a separate package, but for next minor version I can actually fix that. Just did not want to do that on a patch release. (Although both are internal classes, but did not feel right)

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: Not in practice, this happens on re-list only, what should be a rare occasion to my understanding if Kubernetes API server is configured right, that should rarely happen. Also we operate only on top of resources that we cache explicitly, which are should be present only for a very short time.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java
csviri: The thing is that in almost all cases, this will just live while we get the event of our update or we finished the update operation. And we are caching object that is already in memory. So those cases what we test here can happen, just probably won't, and will be very short lived. Note that the algorithm will clean cache also in case of ongoing parallel updates, so not relevant information is discarded as soon as possible.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: true, removed, thx!!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: removed, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/TemporaryResourceCache.java
csviri: removed, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/GenericResourceEvent.java
csviri: changed, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterSupport.java
csviri: changed, than you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java
csviri: Moved the javadoc from check method, pls let me know if you think still needs to be extended.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/GenericResourceEvent.java
csviri: renamed back, thank you!

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java
metacosm: Yes, I know that the objects will be short-lived, hence my concern regarding GC churn. My concern regarding memory is more about the TreeMap and TreeSet fields, which are fairly heavy objects by themselves, and we'll be creating / discarding lots of them, not what we put in them.

metacosm: Regarding the doc, I think it's OK for the javadoc, but I'd like to see comments on the algorithm implementation to clarify what's going on.

csviri: I added comprehensive description to the algorithm, with examples for main paths. @metacosm if it makes sense this way.

csviri: Added some optimizations, AFAIK those short lived objects shot not result in some big performance issues, but will mark this in performance tests issues, and we can examine / optimize based on the numbers see there later.

metacosm (approved):

PR #3433, opened 2026-06-19
test: de-flake DynamicGenericEventSourceRegistrationIT

csviri:
The test captured the reconcile execution count while the create-triggered
reconcile cascade was still running (the first await only checked that the
ConfigMap and Secret existed, not that the system had quiesced), and asserted
an exact +2 executions after the external replace.

The reconciler re-writes the ConfigMap/Secret on every execution, so the number
of self-induced reconciles is timing-dependent, and after the read-after-write
filtering rework (#3414) the reconciler's own follow-up write is filtered, so a
deterministic +2 no longer holds.

Wait for the execution count to stabilize before capturing the baseline, and
assert only a lower bound (at least one more execution) after the replace.


xstefank (commented):
LGTM

PR #3434, opened 2026-06-19
test: fix flaky additionalFinalizer in TriggerReconcilerOnAllEventIT

csviri:
The phase-1 await could pass prematurely: right after create() the resource
already has exactly [ADDITIONAL_FINALIZER], before the operator adds its own
FINALIZER. This snapshotted a stale event count, so the phase-2
isEqualTo(eventCount + 1) assertion against a monotonic counter would overshoot
and time out.

Wait for the operator to add its own finalizer before deleting, so phase 1
reflects the completed add/mark-for-deletion/remove cycle and the baseline is
deterministic.


xstefank (commented):
LGTM

PR #3436, opened 2026-06-22
fix: add onUpdateFilterOr option to the Informer

### LINKED ISSUE #3413: Support SSA in InformerEventSource, listen to specific status updates
I am working on a multi-reconciler setup (Main and Secondary) managing the same Custom Resource. The Secondary reconciler updates a specific field within the resource's .status using Server-Side Apply (SSA). The Main reconciler needs to be notified and triggered only when the resource's generation changes OR when this specific status field is updated by the Secondary reconciler.

Currently, achieving this cleanly is not possible:

- If `generationAwareEventProcessing` is set to `true`, the Main reconciler completely misses the status updates.
- If `generationAwareEventProcessing` is set to `false`, the Main reconciler is triggered far too often by unrelated status updates.
- Attempting to pass a custom `OnUpdateFilter` via the `ControllerConfiguration` fails because custom filters are combined with JOSDK's internal filters using a strict AND operation. This prevents implementing a logical OR condition (e.g., `DefaultFiltering OR SpecificStatusFieldChanged`).

I would like a way to configure the controller's event filtering behavior to allow logical OR operations between the internal filters (like generation-aware filtering) and user-defined custom filters.

Alternatively, a feature allowing developers to keep `generationAwareEventProcessing = true` but explicitly define a list of observed status paths (e.g., `status.mySpecificField`) that should bypass the generation-aware restriction and still trigger an event would elegantly solve this use case. Or another way to achieve this.

1. Setting `generationAwareEventProcessing` to `false`: This causes performance degradation due to a flood of redundant reconciliation loops from other status updates.
2. Creating a separate `EventSource` on our own resource with an `OnUpdateFilter`: While this successfully isolates the specific status change, it introduces two major downsides:
- It forces the creation and maintenance of an unnecessary secondary cache.
- In JOSDK `v5.2.0`, it introduces read-after-write inconsistencies where the resource instance passed to the Main reconciler is outdated at the moment the event source triggers it.

**Additional context**
- Environment: Java, Micronaut, JOSDK `v5.2.0`.
- This issue is particularly relevant in advanced operator patterns where responsibilities are split across a Main and a Secondary controller for the same CRD, relying on status fields to pass state back to the primary loop.

xstefank:
Fixes #3413

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Informer.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/informer/Informer.java
@@ -80,12 +80,26 @@ public @interface Informer {
   Class<? extends OnAddFilter> onAddFilter() default OnAddFilter.class;
 
   /**
-   * Optional {@link OnUpdateFilter} to filter update events sent to the associated informer
+   * Optional {@link OnUpdateFilter} to filter update events sent to the associated informer.
+   * Combined with JOSDK's internal filters using AND logic — the event is only accepted when both
+   * this filter and JOSDK's internal filters accept it.
    *
    * @return the {@link OnUpdateFilter} filter implementation to use, defaulting to the interface
    *     itself if no value is set
    */
   Class<? extends OnUpdateFilter> onUpdateFilter() default OnUpdateFilter.class;
+
+  /**
+   * Optional {@link OnUpdateFilter} combined with JOSDK's internal filters using OR logic — the
+   * event is accepted when either this filter or JOSDK's internal filters accept it. Use this to
+   * expand the set of events that trigger reconciliation beyond what JOSDK's internal filters (e.g.
+   * generation-aware filtering) would normally allow, for instance to also reconcile on specific
+   * status field updates.
+   *
+   * @return the {@link OnUpdateFilter} filter implementation to use, defaulting to the interface
+   *     itself if no value is set
+   */
+  Class<? extends OnUpdateFilter> onUpdateFilterOr() default OnUpdateFilter.class;
 
   /**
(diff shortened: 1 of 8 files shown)


metacosm:
Maybe it would make more sense to have a switch to change the composition behavior from AND to OR instead of defining another filter with a different behavior?

csviri:
@xstefank I see more conceptual issues with this approach. One is what @metacosm mentioned:

> Maybe it would make more sense to have a switch to change the composition behavior from AND to OR instead of defining another filter with a different behavior?

and yes I think too, that would be one way to solve this.

Another approach would be to have a flag to "withoutDefaultFilters", so default filters don't interfere with the users, so user would have a full control of the filters, that would be future-proof, also might impose a simpler mental model.

In addition to that this should not be on Informer level, since it only make sense in context of `ControllerEventSource`.

metacosm:
Another approach would be to expose the internal filters and provide ways to let users combine them as they want. That would be future-proof but maybe a little overkill.

csviri:
> Another approach would be to expose the internal filters and provide ways to let users combine them as they want. That would be future-proof but maybe a little overkill.

Exactly, this is how I meant also, so if users set the flag to not add the default filter, they will set a custom filter(s), but there they could just combine those filters with the existing one in any way they want. But I think they can do this even now, we just don't have the way to "not add" the default ones.

xstefank:
Replaced with https://github.com/operator-framework/java-operator-sdk/pull/3437

xstefank:
do you think that withoutDefaultFilters + verifying that users can still set up the same filter as they have by default would still be a valid use case?

Now with https://github.com/operator-framework/java-operator-sdk/pull/3437, they can switch defaultFilter OR userFilters.

It is a slightly different scenario.

PR #3438, opened 2026-06-23
feat: add defaultFilters option to skip JOSDK internal update filters

### LINKED ISSUE #3413: Support SSA in InformerEventSource, listen to specific status updates

xstefank:
Fixes https://github.com/operator-framework/java-operator-sdk/issues/3413

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
@@ -121,4 +121,8 @@ public interface ControllerConfiguration<P extends HasMetadata> extends Informab
   default boolean triggerReconcilerOnAllEvents() {
     return false;
   }
+
+  default boolean isDefaultFilters() {
+    return true;
+  }
 }
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/InternalEventFilters.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/InternalEventFilters.java
@@ -22,7 +22,7 @@ public class InternalEventFilters {
 
   private InternalEventFilters() {}
 
-  static <T extends HasMetadata> OnUpdateFilter<T> onUpdateMarkedForDeletion() {
+  public static <T extends HasMetadata> OnUpdateFilter<T> onUpdateMarkedForDeletion() {
     // the old resource is checked since in corner cases users might still want to update the status
     // for a resource that is marked for deletion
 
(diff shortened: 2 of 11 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ControllerConfiguration.java
```diff
@@ -121,4 +121,8 @@ default boolean triggerReconcilerOnAllEvent() {
   default boolean triggerReconcilerOnAllEvents() {
     return false;
   }
+
+  default boolean isWithoutDefaultFilters() {
```
csviri: Wonder if we could improve the naming of this flag. Like `useDefaultFilters` / `addDefaultFilters` or just `defaultFilters` ?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/InternalEventFilters.java
```diff
@@ -22,15 +22,15 @@ public class InternalEventFilters {
 
   private InternalEventFilters() {}
```
csviri, line 24: What we could do is to add here a static method "defaultFilters" that returns a composition of the default filters - later also we could use in `ControllerEventSource`.

csviri (changes requested):

csviri (approved):
LGTM

csviri:
Thank you @xstefank !

PR #3451, opened 2026-06-30
improve: generic cache-aware secondary resource lookup for dependent resources

mohit24694:
## Summary

Introduces a generic O(1) cache-aware secondary resource lookup path in `AbstractEventSourceHolderDependentResource`, extending the optimisation beyond `KubernetesDependentResource` to any event-source-backed dependent resource whose event source implements `Cache<R>`.

Also adds a direct cache lookup shortcut in `AbstractExternalDependentResource` for dependents backed by `ExternalResourceCachingEventSource`.

---

## Motivation / Problem

`AbstractDependentResource.getSecondaryResource()` currently calls `Context.getSecondaryResources(resourceType())`, which:

1. Iterates the event source's internal index/cache to build a `Set<R>` of **all** secondary resources associated with the primary.
2. Passes that set to `selectTargetSecondaryResource()` to filter down to the single expected resource.

For **Kubernetes dependent resources** (`KubernetesDependentResource`), `selectTargetSecondaryResource()` additionally calls `getOrComputeDesired()` — just to derive the target name/namespace — then streams and filters the full set by name and namespace. This is wasteful since `InformerEventSource` already provides an O(1) `cache.get(ResourceID)` API.

For **external dependent resources** (`AbstractExternalDependentResource`), `selectTargetSecondaryResource()` calls `desired(primary, context)` equally unnecessarily for what is essentially a lookup operation.

A Kubernetes-specific optimisation existed in the `optimize-kube-dependent` branch but was never merged and applied only to the Kubernetes case. This PR generalises the pattern to the shared base class so any `Cache<R>`-backed dependent resource benefits without extra implementation work.

---

### `AbstractEventSourceHolderDependentResource`

- Overrides `getSecondaryResource(P primary, Context<P> context)` to check whether the associated event source implements `Cache<R>` (e.g. `InformerEventSource` via `ManagedInformerEventSource`). When it does **and** `targetSecondaryResourceID()` returns a non-null `ResourceID`, the framework calls `cache.get(resourceID)` directly — a single O(1) hash-map lookup.
- Adds `targetSecondaryResourceID(P primary, Context<P> context)` as a protected hook method returning `null` by default. The `null` default preserves the existing behaviour completely, making this **fully backward compatible**: no existing subclass needs to change.

### `KubernetesDependentResource`

- The existing `targetSecondaryResourceID()` override naturally wires into the new generic parent mechanism via `@Override` — **no functional change**.
- `selectTargetSecondaryResource()` is retained as a safety fallback for edge cases (shared event sources, non-informer backing stores). In the common case (event source is `InformerEventSource`) this method is no longer called.
- Both methods re-documented to make the fast-path / fallback relationship explicit.

### `AbstractExternalDependentResource`

- Overrides `getSecondaryResource()` to call `ExternalResourceCachingEventSource.getSecondaryResource(ResourceID primaryID)` directly when the event source is of that type. This bypasses both `Context.getSecondaryResources()` and `selectTargetSecondaryResource()`, the latter of which called `desired(primary, context)` solely to identify the resource.
- Falls back to the standard path for custom or shared event source implementations.

---

## Backward Compatibility

All changes are strictly additive or override-with-fallback:

| Condition | Behaviour |
|---|---|
| Event source implements `Cache<R>` **and** `targetSecondaryResourceID()` returns non-null | ✅ New O(1) cache path taken |
| Event source implements `Cache<R>` but `targetSecondaryResourceID()` returns `null` (default) | Existing path — **no change** |
| Event source does NOT implement `Cache<R>` | Existing path — **no change** |
| `AbstractExternalDependentResource` with `ExternalResourceCachingEventSource` | Direct primary-keyed lookup, no `desired()` call |
| `AbstractExternalDependentResource` with custom event source | Existing path — **no change** |

No API is removed. No existing concrete subclass needs to change.

---

## Performance Impact

**`KubernetesDependentResource` (common case — event source is always `InformerEventSource implements Cache<R>`):**

- **Before:** `context.getSecondaryResources()` → index traversal → `HashSet<R>` allocation → `selectTargetSecondaryResource()` → `getOrComputeDesired()` → stream filter by name/namespace
- **After:** `InformerEventSource.get(ResourceID)` — single hash-map lookup

**`AbstractExternalDependentResource`:**

- **Before:** `context.getSecondaryResources()` → `HashSet<R>` → `selectTargetSecondaryResource()` → `desired()` computation → stream filter
- **After:** `ExternalResourceCachingEventSource.getSecondaryResource(primaryID)` — direct map lookup, no `desired()` computation

---

## Further optimisation (for implementors)

Override `targetSecondaryResourceID()` to return a `ResourceID` derived **statically** from the primary (e.g. from the primary's name and namespace) to avoid even calling `getOrComputeDesired()` during a lookup:

```java
@Override
protected ResourceID targetSecondaryResourceID(MyPrimary primary, Context<MyPrimary> context) {
    // No desired() call needed — derive name from primary directly
    return new ResourceID(
        primary.getMetadata().getName() + "-config",
        primary.getMetadata().getNamespace());
}
```

---

## Testing

All existing unit tests pass (`mvn test -pl operator-framework-core`). No new integration tests are added: the change is a lookup-path selection optimisation whose correctness is fully covered by the existing test suite.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/KubernetesDependentResource.java
@@ -247,6 +247,48 @@ public abstract class KubernetesDependentResource<R extends HasMetadata, P exten
     annotations.put(typeKey, GroupVersionKind.gvkFor(primary.getClass()).toGVKString());
   }
 
+  /**
+   * Returns the secondary resource using a direct O(1) lookup against the {@link
+   * InformerEventSource} cache when possible, bypassing the costlier {@link
+   * io.javaoperatorsdk.operator.api.reconciler.Context#getSecondaryResources} path.
+   *
+   * <p>The fast path is taken when the event source is present and {@link
+   * #targetSecondaryResourceID(HasMetadata, Context)} returns a non-null {@link ResourceID}. In
+   * that case {@code informerEventSource.get(resourceID)} is called directly — a single hash-map
+   * lookup — without building a {@code Set} of all secondary resources and filtering through it.
+   *
+   * <p>Falls back to the standard {@link
+   * AbstractDependentResource#getSecondaryResource(HasMetadata, Context)} path (which delegates to
+   * {@link #selectTargetSecondaryResource(Set, HasMetadata, Context)}) when the event source is not
+   * present or {@code targetSecondaryResourceID} returns {@code null}.
+   *
+   * @param primary the primary resource
+   * @param context the reconciliation context
+   * @return the secondary resource if found, or {@link Optional#empty()}
+   */
+  @Override
+  public Optional<R> getSecondaryResource(P primary, Context<P> context) {
+    return eventSource()
+        .flatMap(
+            es -> {
+              ResourceID targetID = targetSecondaryResourceID(primary, context);
+              if (targetID != null) {
+                return es.get(targetID);
+              }
+              return super.getSecondaryResource(primary, context);
+            });
+  }
+
+  /**
+   * Selects the target secondary resource from the full set of candidates. This is the fallback
+   * path taken when {@link #getSecondaryResource(HasMetadata, Context)} cannot use the direct cache
+   * lookup (e.g. when a shared event source is used and the event source reference is not directly
+   * available via {@link #eventSource()}).
+   *
+   * <p>Prefer overriding {@link #targetSecondaryResourceID(HasMetadata, Context)} for
+   * performance-critical paths to avoid both this filtering and the {@code getOrComputeDesired}
+   * call it triggers.
+   */
   @Override
   protected Optional<R> selectTargetSecondaryResource(
       Set<R> secondaryResources, P primary, Context<P> context) {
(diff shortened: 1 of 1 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractEventSourceHolderDependentResource.java
```diff
@@ -123,6 +124,60 @@ public Optional<T> eventSource() {
...
+   * ResourceID} (e.g. based on the primary's name or a fixed naming convention) to avoid computing
+   * the full desired state during secondary resource lookup. See {@code
+   * KubernetesDependentResource#targetSecondaryResourceID} for a concrete example.
+   *
+   * @param primary the primary resource associated with the secondary resource being looked up
+   * @param context the current reconciliation context
+   * @return the {@link ResourceID} identifying the expected secondary resource, or {@code null} if
+   *     not known (falls back to the default lookup path)
+   */
+  protected ResourceID targetSecondaryResourceID(P primary, Context<P> context) {
```
csviri: Secondary resource in this case is not limited to Kubernetes resources (R does not extend HasMetadata), therefore not necessarily has ResourceID as ID

csviri (changes requested):
Hi @mohit24694 , thank you for the PR.

Please target `next` branch with such PR-s which are not bug fixes.

Added also a comment, with an issue.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractEventSourceHolderDependentResource.java
mohit24694: Acknowledged, checking the comment.

mohit24694: Good catch — you are correct that R in AbstractEventSourceHolderDependentResource is not bounded to HasMetadata, making ResourceID an inappropriate secondary resource identifier at that level.

The fix moves the entire optimisation into KubernetesDependentResource (where R extends HasMetadata and ResourceID is valid) and reverts AbstractEventSourceHolderDependentResource to its original state. The new getSecondaryResource() override in KubernetesDependentResource calls InformerEventSource.get(targetSecondaryResourceID) directly, falling back to the standard AbstractDependentResource path when eventSource() is empty (e.g. shared event source). The base class is untouched.

csviri:
Hi @mohit24694 pls fix the spotbug issue. 

Regarding the change, this makes sense, but little bends the contract, what is otherwise uniform across the implementations of various dependent resources. In generic case this might be a bit more efficient, but would be nice to see the numbers how much, would it be possible to measure that @mohit24694 ?

cc @metacosm please take a look, and let us know what do you think.

xstefank (changes requested):
We cannot accept a change that breaks main. As @csviri said, this should target "next" branch and the tests need to be passing.

PR #3456, opened 2026-07-01
test: failing test for EventFilterWindow coalesced foreign spec change

### LINKED ISSUE #3455: EventFilterWindow silently drops foreign spec changes coalesced with own-write echo
Supersedes #3445, which investigated the wrong path (shared InformerEventSource). This issue identifies the actual root cause in `EventFilterWindow`.

## Summary

When a controller uses SSA (Server-Side Apply) and a foreign actor concurrently changes the same resource's spec, the `EventFilterWindow` can silently drop the foreign change. The controller never re-reconciles and is stuck until `maxReconciliationInterval` (default 10 hours).

## Mechanism

1. Controller reconciles resource at **gen=1**
2. Foreign actor changes spec → **gen=2** → new RV on API server
3. Controller's SSA `patchResource` merges on top → API returns **RV N** (gen=2, because SSA merged with the already-modified resource)
4. The fabric8 informer coalesces both watch events (same resource key) into a **single event at RV N**
5. `EventFilterWindow` records own write RV=N, sees `relatedEvents={N} == ownResourceVersions={N}` → **drops the event** ([`EventFilterWindow.java` line 182–186](https://github.com/operator-framework/java-operator-sdk/blob/f1237ba1/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindow.java#L182-L186))
6. Controller finishes gen=1 reconciliation successfully, no pending events → rescheduled for `maxReconciliationInterval`
7. Resource has gen=2 but `observedGeneration=1` — **permanently stale** until the 10-hour timer fires

## Why SSA is specifically affected

With non-SSA updates (full PUT), a concurrent foreign change causes a **409 Conflict**, forcing the controller to retry with the current resource state. With SSA, patches succeed even when the resource was concurrently modified — different field managers don't conflict. The API response RV carries both changes, but `EventFilterWindow` treats it as a pure own-write echo.

## Why this didn't happen before 5.3.x

The `EventFilterWindow` buffering mechanism was introduced in 5.3.x. It captures events that arrive **during** the API call window (`startEventFilteringModify` → `doneEventFilterModify`). When the informer coalesces the foreign change event with the own-write echo into a single event, the window sees only one event at the own-write RV and drops it.

In 5.2.x, the simpler `TemporaryResourceCache` / `canSkipEvent` approach checked events as they arrived. With different timing characteristics (no synchronized `onAddOrUpdate`, no buffering window), the foreign event was more likely to arrive and be processed before the own-write RV was recorded in the cache.

## Observed in production

This was observed in a Kroxylicious operator GHA failure when upgrading from JOSDK 5.2.5 to 5.3.5. A dependent controller sees the stale resource (`gen=2, observedGeneration=1`), exhausts retries, and also goes to sleep for `maxReconciliationInterval`.

SamBarker:
## Summary

- Adds a `@Disabled` unit test demonstrating that `EventFilterWindow` silently drops a foreign spec change when the informer coalesces it with the controller's own-write echo at the same resource version

## Related

Closes #3455

### DIFF
--- a/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindowTest.java
+++ b/operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindowTest.java
@@ -538,6 +538,38 @@ class EventFilterWindowTest {
   }
 
   @Test
+  @Disabled("EventFilterWindow drops coalesced foreign spec changes — #3455")
+  void coalescedForeignSpecChangeShouldNotBeFilteredByOwnWriteRV() {
+    // Reproduces the scenario where a controller's SSA patch merges with a concurrent
+    // foreign spec change. The informer coalesces both events into one at the own-write RV.
+    //
+    // Real-world sequence (from Kroxylicious GHA failure):
+    //   1. Controller reconciles resource at gen=1
+    //   2. Foreign actor changes spec → gen=2 → new RV on API server
+    //   3. Controller's SSA patchResource merges on top → returns RV=5 (gen=2)
+    //   4. Informer coalesces both events (same resource key) into one at RV=5
+    //   5. EventFilterWindow sees relatedEvents={5} == ownResourceVersions={5} → drops
+    //   6. Controller never learns about gen=2 → stuck until maxReconciliationInterval
+    eventFilterWindow.increaseActiveUpdates();
+    eventFilterWindow.addToOwnUpdateVersions(s(FIRST_OWN_VERSION));
+
+    var resource = testResource(FIRST_OWN_VERSION);
+    resource.getMetadata().setGeneration(2L);
+    var previousResource = testResource(FIRST_OWN_VERSION - 1);
+    previousResource.getMetadata().setGeneration(1L);
+    eventFilterWindow.addRelatedEvent(
+        new ExtendedResourceEvent(UPDATED, resource, previousResource, null));
+
+    eventFilterWindow.decreaseActiveUpdates();
+
+    assertThat(eventFilterWindow.check())
+        .as(
+            "Event at own-write RV with generation change (1→2) should propagate:"
+                + " a foreign spec change was coalesced with the own-write echo")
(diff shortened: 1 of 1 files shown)


csviri:
see comment here: https://github.com/operator-framework/java-operator-sdk/issues/3455#issuecomment-4851485763

operator-framework-core/src/test/java/io/javaoperatorsdk/operator/processing/event/source/informer/EventFilterWindowTest.java
```diff
@@ -537,6 +537,38 @@ void reListBetweenTwoUpdates() {
...
+    //   3. Controller's SSA patchResource merges on top → returns RV=5 (gen=2)
+    //   4. Informer coalesces both events (same resource key) into one at RV=5
+    //   5. EventFilterWindow sees relatedEvents={5} == ownResourceVersions={5} → drops
+    //   6. Controller never learns about gen=2 → stuck until maxReconciliationInterval
+    eventFilterWindow.increaseActiveUpdates();
+    eventFilterWindow.addToOwnUpdateVersions(s(FIRST_OWN_VERSION));
+
+    var resource = testResource(FIRST_OWN_VERSION);
+    resource.getMetadata().setGeneration(2L);
+    var previousResource = testResource(FIRST_OWN_VERSION - 1);
```
csviri, line 558: Just some more detail on this, this is not really a realistic scenario, I mean it is but in this case the behavior is correct. Since before you scenario you would receive an update event where the actual resource would have version `testResource(FIRST_OWN_VERSION - 1)` (previous resource version FIRST_OWN_VERSION - 2 ), that would trigger the reconciliation.

A more realistic, and that might cause issue with actual version is 
`var previousResource = testResource(FIRST_OWN_VERSION - 2);`

So a the version `FIRST_OWN_VERSION - 1` event would be lost. But this happens only on re-list:

This will be covered with the mentioned commit.

csviri:
Closing see, details in the related issue

PR #3467, opened 2026-07-02
docs: event filter page

### LINKED ISSUE #3466: Add event filtering page into documentation
Currently we don't describe usage of filters in documentation. We should add a dedicated page.

csviri:

### DIFF
--- a/docs/content/en/docs/documentation/event-filters.md
+++ b/docs/content/en/docs/documentation/event-filters.md
@@ -158,3 +158,13 @@ methods on
 [`InternalEventFilters`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/controller/InternalEventFilters.java)
 inside your own filter.
 {{% /alert %}}
+
+## Non-informer based event sources
+
+The filters described above apply only to informer-backed event sources, since they operate on the
+*add*, *update* and *delete* events an informer receives from the Kubernetes API. Event sources that
+are not backed by an informer — such as [`PollingEventSource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/polling/PollingEventSource.java)
+and [`PerResourcePollingEventSource`](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/polling/PerResourcePollingEventSource.java)
+— have no filters. For those, it is up to your implementation to emit only the events that are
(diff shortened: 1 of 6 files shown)


docs/content/en/docs/documentation/event-filters.md
```diff
@@ -0,0 +1,160 @@
...
+Every [event source](eventing.md) that watches Kubernetes resources is backed by an informer that
+receives *add*, *update* and *delete* events from the Kubernetes API. Not every one of those events
+is worth a reconciliation: a resource might be touched by another controller, its `.metadata` might
+change without any meaningful change to its `.spec`, or your own controller might have caused the
+change in the first place. **Event filters** let you decide, per event, whether it should be
+propagated and (potentially) trigger a reconciliation. Filtering events as early as possible keeps
+your operator efficient by avoiding needless reconciliations.
+
+Filters apply both to the **primary resource** (the resource your `Reconciler` manages) and to any
+**secondary resources** watched through an [`InformerEventSource`](eventing.md#informereventsource).
```
metacosm, line 15: Maybe it's worth pointing that for non-Kubernetes resources, there's no need for filtering because it should be the responsibility of the `EventSource` to only emit relevant events?

xstefank (approved):

docs/content/en/docs/documentation/event-filters.md
```diff
@@ -0,0 +1,160 @@
...
+`GenericFilter` is applied to every kind of event, so it is a convenient way to express a condition
+that is independent of the event type. When both a specific filter (e.g. `OnUpdateFilter`) and the
+`GenericFilter` are configured, **both** must accept the event for it to be propagated.
+
+All four interfaces are `@FunctionalInterface`s and provide `and(...)`, `or(...)` and `not()`
+default methods, so you can compose several conditions:
+
+```java
+OnUpdateFilter<MyCustomResource> filter =
+    onUpdateFilterA.and(onUpdateFilterB).not();
```
metacosm, line 40: It would be interesting to explain what the resulting "filter" would be because it's not obvious. Also, what the composition rules are, like precedence if any…

metacosm (changes requested):

docs/content/en/docs/documentation/event-filters.md
csviri: added dedicated section at the end. Thank you!

docs/content/en/docs/documentation/event-filters.md
csviri: We define there logical operations, which are trivial if somebody takes a look into source code, not sure if we should explicitly explain that.

@xstefank what do you think?

csviri: I added javadoc to filter base classes

PR #3475, opened 2026-07-03
docs: note that shard selectors are alpha in k8s 1.36

csviri:


xstefank (approved):
Do we also have this in the code if the user starts an operator using this feature? Or in other words, how do we fail if JOSDK tries to use shards and cluster doesn't support it?

csviri:
> Do we also have this in the code if the user starts an operator using this feature? Or in other words, how do we fail if JOSDK tries to use shards and cluster doesn't support it?

No, kubernetes API just ignores these flags in that case, would be nice to detect, but there is no easy and nice way to do that (what I can think of).  But also it is not our responsibility to document k8s api.

PR #3486, opened 2026-07-11
fix: event filtering edge case with no-op updates - extended

csviri:
Extended version of https://github.com/operator-framework/java-operator-sdk/pull/3484

Which adds more refined API, explicit option to cache only (not filtering) and future proofed to machers.

TODO:

update javadocs, documentation, release create post.


csviri:
@xstefank @metacosm prepared this enhanced version, would be probably nicer to go with this and release a minor version.

PR #3491, opened 2026-07-13
feat: resource operation support multiple update modes

### LINKED ISSUE #3479: Superseding event during retry does not reset retry counter — resource can permanently stall
### What did you do?

Operated a reconciler with an `InformerEventSource` watching a secondary resource. The reconciler throws when the secondary's status is stale (its informer cache hasn't propagated the latest update yet). The retry policy is the default `@GradualRetry`.

During the retry cycle, the informer delivers a superseding event for the secondary resource (status is now fresh). The superseding event triggers an immediate re-execution (`handleRetryOnException` line 382–385), but the informer cache is still stale at execution time — so that re-execution also fails.

No `maxReconciliationInterval` is configured.

### What did you expect to see?

The superseding event represents genuinely new external state. When its re-execution fails, the resource should get a **fresh retry budget** — the new event is a different failure from the original, and the retry counter should reset.

After one more retry (with the cache now fresh), the reconciler should succeed.

### What did you see instead? Under which circumstances?

The retry counter from the original failure cycle is **never reset**. When the superseding event's re-execution fails, `handleRetryOnException` falls through to `nextDelay()` (line 396), which returns `Optional.empty()` because `lastAttemptIndex >= maxAttempts` — the budget was already consumed by the original cycle.

The resource is permanently stuck. No further reconciliation is attempted.

From the operator logs (real CI failure):

```
04:38:42  Retry 5 → "another reconciliation will be attempted because a superseding event
           has been received or another retry attempt is pending"
04:38:49  Retry 6 → "no superseding event is present and this is the retry last attempt"
           → "Exhausted retries for scope: ..."
```

The superseding event was consumed during retry 5, but the informer cache was still stale. By retry 6 the event flag is cleared — the event has been "used up" without ever driving a successful reconciliation.

### Root cause

Two things combine:

1. **`handleRetryOnException` does not reset the retry counter when `eventPresent` is true** (line 382–385). It calls `submitReconciliationExecution` immediately (skipping `nextDelay()`), but the `GenericRetryExecution.lastAttemptIndex` is carried forward unchanged.

2. **`GenericRetryExecution` has no reset mechanism.** `cleanupOnSuccessfulExecution` sets the retry to `null` (line 441), but nothing resets the counter for a new failure cycle triggered by a superseding event.

So when the superseding event's re-execution fails, the *next* call to `nextDelay()` sees the counter from the *original* cycle and reports exhaustion.

### Reproducing test

PR #3478 has two `@Disabled` failing tests in `EventProcessorTest`:

1. **`newEventShouldBeRetryableAfterPriorRetryExhaustion`** — unit-level: a new event after retry exhaustion inherits the exhausted counter and gets zero retry budget.

2. **`supersedingEventConsumedDuringRetryShouldNotPermanentlyStallReconciliation`** — uses a real `TimerEventSource` and `CountDownLatch` to reproduce the end-to-end scenario:
   - Initial event → fails → timer retry starts (blocked by latch)
   - Superseding event injected while retry is executing
   - Latch released → retry fails → superseding event triggers immediate re-execution → also fails
   - `nextDelay()` returns empty (exhausted from original cycle) → resource permanently stuck
   - Expects 4 executions (the 4th being a retry with fresh budget that succeeds), but only 3 occur.

### Environment

**Kubernetes cluster type:** vanilla (KinD)

**java-operator-sdk version:** 5.2.5 (also reproduced at 5.4.0, code path unchanged on `main`)

**Java:** 21

### Possible Solution

Reset the retry state when a superseding event triggers re-execution. One approach: in `handleRetryOnException`, when `eventPresent` is true (line 382), set `state.setRetry(null)` before calling `submitReconciliationExecution`. This gives the superseding event's execution a fresh `GenericRetryExecution` if it fails.

This matches the semantics: a superseding event is new external state, not a continuation of the previous failure. It should get its own retry budget.

### Additional context

`maxReconciliationInterval` masks this bug — if configured, `scheduleExecutionForMaxReconciliationInterval` fires after exhaustion (line 404) and eventually drives a successful reconciliation. But the default is no interval, and operators that don't configure it are exposed to permanent stalls.

### LINKED ISSUE #3483: Enhanced mode(s) in ResourceOperation
Recently we changed event filtering, see details here: https://github.com/operator-framework/java-operator-sdk/pull/3484
That bit limits the capabilities of this feature.

In next minor version we could still provide improved alternatives in [ResourceOperations](https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java) that user can use to optimize their specific use cases.

## Prevent no-op updates

Things to consider regarding this enhancement:

1. Users ideally do just updates/patches on resources when the desired state of those resources are not in the desired state.
2. We discovered an issue with event filtering, it only works reliable in all the cases when the update changes the resources (so results in a bump in resource version)

Introducing an update mode for methods in `ResourceOperations` , that will match the current state with desired state and executes the patch only those cases, would handle these two situations optimally.

We already have such matchers implemented and battle tested partially (not for status updates) for dependent resources, so we can just reuse them here.

## Explicit cache-only methods

User would have an option to explicitly set to not filter the events just cache the resource.

## Other Remarks

- In `UpdateControl` we still have `.patchResource` and `.patchResourceAndStatus`. In future we might:
  1. want to deprecated this, since generally considered as a bad practice to do updates on the spec of the controller. (Users with some special need will be able to still do it using the `KubernetesClient` or `ResourceOperations`.
  2. If other than the status is updated we might want to enforce optimistic locking in `UpdateControl` since there is a risk that it would override some changes on spec from third party.

csviri:
### Summary

  This reworks `ResourceOperations` into the primary, idiomatic API for create/update/patch operations that keep the informer cache read-after-write consistent **and** control how the resulting own event is handled. Every operation comes in a
  default flavor and an `Options`-taking flavor, so users can tune caching vs. own-event filtering per call.

  Own-event filtering is only correct when the framework can distinguish an own write from a concurrent third-party write. This PR makes that requirement explicit through a small strategy model (`Options` → `Mode`) and a pluggable `Matcher`
  framework with sensible per-operation defaults.

  ### `Options` / `Mode`

  - **`matchAndFilter(matcher)` / `matchAndFilterWithDefaultMatcher(updateType)`** — compare desired vs. actual (cached) state; skip the write entirely if they already match, otherwise write and filter the own event. This is the **default** for the
  update/patch methods and the most efficient option (filters *and* avoids a needless API call).
  - **`filterWithOptimisticLocking()`** — filter the own event; the write **must** use optimistic locking (a resourceVersion set), otherwise an `IllegalArgumentException` is thrown. Guarantees a concurrent change is rejected server-side rather than
  silently filtered out.
  - **`cacheOnly()`** — only cache the response (read-after-write consistency), no own-event filtering.
  - **`forceFilterEvents()`** — always filter (mostly internal; safe only when correctness is otherwise guaranteed).

  ### Matcher framework

  - New `Matcher` interface plus default matchers for every `UpdateType` (`UPDATE`, `SSA`, `JSON_PATCH`, `JSON_MERGE_PATCH`, and their `*_STATUS` variants), backed by `MatcherUtils`.
  - `GenericKubernetesResourceMatcher.matchStatus(...)` added to match only the `/status` subtree (tolerating server-added fields by default).

  ### Framework integration

  - **Finalizer handling** (`ReconciliationDispatcher`): the finalizer is now added with `cacheOnly` (its own event is no longer filtered), replacing the previous filter + `INSTANT_RESCHEDULE` so the finalizer-add event itself drives the follow-up
  reconcile.
  - **Primary `UpdateControl` writes** now use `cacheOnly`, trading one extra self-triggered reconcile for correctness (a concurrent spec change during a status patch can no longer be absorbed by the own-event filter).
  - **`KubernetesDependentResource`** create/SSA now use `forceFilterEvents()`.
  - `AbstractExternalDependentResource` persists explicit state through `resourceOperations().create(...)`.

  ### Tests & docs
  
  - New ITs covering the concurrency and read-after-write edge cases: `ResourceOperationsIT`, `SecondaryResourceOperationsIT`, `SpecChangeDuringStatusPatchIT`, `OwnSsaStatusUpdateIT`, plus expanded `ResourceOperationsTest`, `PatchMatchersTest`,
  `StatusMatchersTest`.
  - Adjusted existing tests to the new event cadence (e.g. `SubResourceUpdateIT`, `WorkflowMultipleActivationIT` now snapshot reconcile counts only after they stabilize).
  - Docs updated: `reconciler.md`, `v5-5-migration.md`, `v5-5-release.md`, and the read-after-write blog post.

  ### Notes / follow-ups

  - The `Options` API is marked `@Experimental` (API may change).
  - Default matchers are heuristics — reconcilers relying on them should be tested against their concrete resources, or supply a custom `Matcher`.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -205,346 +317,586 @@ public P serverSideApplyPrimaryStatus(P resource) {
...
+      UnaryOperator<R> unaryOperator,
+      InformerEventSource<R, P> informerEventSource,
+      Options options) {
+
+    return resourcePatch(
+        actualResource,
+        r -> context.getClient().resource(actualResource).edit(unaryOperator),
+        informerEventSource,
+        options);
+  }
```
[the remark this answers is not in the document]
csviri: fixed

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -205,346 +317,586 @@ public P serverSideApplyPrimaryStatus(P resource) {
...
    */
-  public <R extends HasMetadata> R jsonPatchStatus(R resource, UnaryOperator<R> unaryOperator) {
-    return resourcePatch(resource, r -> context.getClient().resource(r).editStatus(unaryOperator));
+  public <R extends HasMetadata> R jsonPatch(
+      R actualResource, UnaryOperator<R> unaryOperator, Options options) {
+    return resourcePatch(
+        actualResource,
+        r -> context.getClient().resource(actualResource).edit(unaryOperator),
+        options);
   }
```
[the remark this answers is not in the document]
csviri: fixed

docs/content/en/blog/news/read-after-write-consistency.md
```diff
@@ -226,10 +220,30 @@ sequenceDiagram
...
+public UpdateControl<WebPage> reconcile(WebPage webPage, Context<WebPage> context) {
+    
+    ConfigMap managedConfigMap = prepareConfigMap(webPage);
+    
+   // resource operation in this case will resource only if
+   // it does not match the actual, and will filter our the related event
+   context.resourceOperations().serverSideApply(managedConfigMap);
+
+   // UpdateControl.patchStatus would only cache the resource to filter out events too
+   // you have to use resourceOperations.
```
xstefank: this should be better structured, it can be confusing

docs/content/en/blog/releases/v5-5-release.md
```diff
@@ -0,0 +1,120 @@
+---
+title: Version 5.5 Released!
```
xstefank: Shouldn't this be a separate PR? Won't you squash commits on merge?

docs/content/en/blog/releases/v5-5-release.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Matcher-based updates in `ResourceOperations`
+
+`ResourceOperations` (available from the reconciliation `Context` via `context.resourceOperations()`)
+now offers a complete, consistent family of update/patch/create methods for every write strategy —
+**server-side apply, update (PUT), JSON Patch (RFC 6902) and JSON Merge Patch (RFC 7386)** — each
+available for the whole resource and the `status` subresource, and with dedicated `primary`
+variants.
+
+Every method comes in two flavors: a default one, and one taking an `Options` argument that controls
```
xstefank: Maybe we should also have a global default change so the user doesn't have to pass the argument to each method call.

docs/content/en/blog/releases/v5-5-release.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Migration Notes
+
+There are **no breaking API changes**; existing code compiles and runs unchanged. However:
+
+### `UpdateControl` no longer filters own events by default
+
+Returning an `UpdateControl` (or `ErrorStatusUpdateControl`) still updates the resource and keeps the
+cache read-after-write consistent, but it **no longer filters the resulting own event by default** —
+so the write may cause an additional (idempotent) reconciliation. This change fixes correctness edge
```
xstefank:
```suggestion
so the write may cause an additional reconciliation (which should be idempotent). This change fixes correctness edge
```

docs/content/en/docs/documentation/reconciler.md
```diff
@@ -208,20 +241,26 @@ public UpdateControl<WebPage> reconcile(WebPage webPage, Context<WebPage> contex
...
+    // patches the status and caches the response (does not filter the own event by default, see below)
     return UpdateControl.patchStatus(webPage);
 }
 ```
 
-`UpdateControl` and `ErrorStatusUpdateControl` by default use this functionality, but you can also update your primary resource at any time during the reconciliation using `ResourceOperations`:
+{{% alert title="UpdateControl and event filtering" %}}
+Since v5.5, returning an `UpdateControl` (or `ErrorStatusUpdateControl`) updates the resource and
+keeps the cache read-after-write consistent, but by default it **no longer filters the own event** —
+the resulting update may cause an additional (idempotent) reconciliation. To also filter the own
```
xstefank: same as above

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -205,346 +318,589 @@ public P serverSideApplyPrimaryStatus(P resource) {
...
-   * @return updated resource
-   * @param <R> resource type
+   * @param resource the resource to create
+   * @param <R> the resource type
+   * @return the created resource as returned by the API server
    */
   public <R extends HasMetadata> R create(R resource) {
-    return resourcePatch(resource, r -> context.getClient().resource(r).create());
+    // it is safe to do event filtering for create since check if the resource already exists.
+    return create(resource, Options.forceFilterEvents());
```
xstefank, line 401: I understand, however, once something is default, I think it should be default everywhere. If you implement that changing of default as I suggested above, this might be confusing. Maybe then don't provide Options parametrized overload and just keep this as implementation detail.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -553,19 +909,96 @@ public <R extends HasMetadata> R resourcePatch(R resource, UnaryOperator<R> upda
...
+        log.debug(
+            "Resource match resource id: {}, type: {}, version: {}",
+            ResourceID.fromResource(desiredResource),
+            desiredResource.getClass().getSimpleName(),
+            desiredResource.getMetadata().getResourceVersion());
+      }
+      return actualResource;
+    }
+
+    if (options.getMode() == Mode.FILTER_WITH_OPTIMISTIC_LOCKING
```
xstefank, line 995: maybe switch?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
```diff
@@ -754,4 +1247,171 @@ public P addFinalizerWithSSA(String finalizerName) {
...
+
+  /**
+   * The strategy used to decide how the own event resulting from a write is handled. This is a
+   * low-level enum; users should not reference it directly but select the desired behavior through
+   * the {@link Options} factory methods (e.g. {@link Options#filterWithOptimisticLocking()}, {@link
+   * Options#matchAndFilter(Matcher)}, {@link Options#cacheOnly()}, {@link
+   * Options#forceFilterEvents()}), which is why each constant links to its corresponding factory.
+   */
+  @Experimental(API_MIGHT_CHANGE)
+  enum Mode {
```
xstefank: why package-private?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
```diff
@@ -195,6 +197,58 @@ public static <R extends HasMetadata, P extends HasMetadata> Matcher.Result<R> m
...
+    final var kubernetesSerialization = context.getClient().getKubernetesSerialization();
+    var desiredNode = kubernetesSerialization.convertValue(desired, JsonNode.class);
+    var actualNode = kubernetesSerialization.convertValue(actualResource, JsonNode.class);
+    var wholeDiffJsonPatch =
+        JsonDiff.asJson(desiredNode, actualNode, DiffFlags.dontNormalizeOpIntoMoveAndCopy());
+
+    boolean matched = true;
+    for (int i = 0; i < wholeDiffJsonPatch.size() && matched; i++) {
+      var node = wholeDiffJsonPatch.get(i);
+      if (nodeIsChildOf(node, List.of(STATUS))) {
```
xstefank, line 244: should this also support nested values?

docs/content/en/blog/releases/v5-5-release.md
csviri: Yes, but since this is I have only this PR related changes in here, I added as a part of the commit, want to make a separate PR for all other changes after we merged this. But if you insist can have this as a separate PR.

docs/content/en/blog/releases/v5-5-release.md
csviri: Not sure what do you mean, could you pls elaborate ? maybe with some code snippet

docs/content/en/docs/documentation/reconciler.md
csviri: fixed, thx

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
csviri: The default methods are always filtering, we achieve that with the Matchers (to cover the problematic edge case). We however want to give the user possibility to not filter (cache-only). The API might be more restrictive in that regard, but this way it is also forward compatible.

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
csviri: ?? don't follow, IMHO this is much more readable than force a swtich there

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
csviri: changed it to public, good catch

docs/content/en/blog/news/read-after-write-consistency.md
csviri: changed

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
csviri: it supports nested values, there is a recursive call there.

docs/content/en/blog/releases/v5-5-release.md
xstefank: for history purposes it would be better to split

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
xstefank: ok

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/kubernetes/GenericKubernetesResourceMatcher.java
xstefank: ok, didn't check, thanks

docs/content/en/blog/releases/v5-5-release.md
csviri: changed! created a separate PR: https://github.com/operator-framework/java-operator-sdk/pull/3497

Thank you!

xstefank (approved):

PR #3501, opened 2026-07-16
improve: set sonar sonar version explicitly

csviri:
Set sonar version to 5.7.0.6970 in GitHub Action


xstefank (approved):
however, maybe this should be in pom, not hardcoded in CI.

csviri:
Sorry just started to experiment with this, actually will close this for now, since the core of the sonar issues is somewhere else that we will have to tackle in a separate PR anyways.

Issue #3517, opened 2026-07-30
Correctness AI check: collection of possible issues and related PRs


csviri:
An audit of the codebase turned up 15 correctness bugs plus one cosmetic nit.
Each is fixed by its own PR, linked below; the reasoning, impact and (where
present) the regression test live in the individual commit messages.

SpotBugs (`effort=Max`, `threshold=Medium`) reports zero findings across all
modules, so none of these are reachable by the existing static analysis.

### Confirmed with regression tests

Each of these ships a test that fails without the fix and passes with it.

| Area | Bug |
|---|---|
| `ExternalResourceCachingEventSource` | Configuring only a generic filter throws NPE on add/delete/update — the branch guard is `\|\|` but the body dereferences the specific filter. Affects all polling and inbound event sources. |
| `ExternalResourceCachingEventSource` | `handleRecentResourceUpdate` NPEs when the primary is cached but the secondary id is not. |
| `AbstractDependentResource` / bulk | A `BulkDependentResource` that is not a `Creator`/`Updater` still gets create/update attempted, failing with `ClassCastException`. The `creatable()`/`updatable()` hooks added for this were never wired up. |
| `LinearRateLimiter` | `isLimited` returns time *elapsed* instead of time *remaining*, so rate-limited resources are rescheduled almost immediately (burst of timer events) and then wait roughly a full extra period. |
| `MicrometerMetricsV2` | `processing.started.latency` is registered against a boxed `Long` that nothing holds a strong reference to, so the gauge always reports `NaN`. |
| `PollingEventSource` | The timer is a non-daemon `Timer` created in a final field, so it keeps the JVM alive and the event source cannot be restarted (`IllegalStateException: Timer already cancelled`). |
| `ExternalResourceCachingEventSource` | Cache reads are unsynchronized while every mutator is `synchronized`, and `getSecondaryResources` looks the primary up twice (TOCTOU NPE). |
| `EventProcessor` | When the primary is gone from cache at execution time the submission is never completed: the resource stays flagged as under processing (never reconciled again) and the in-flight metrics gauges drift. |
| `ExecutorServiceManager` | The scheduled executor is never shut down; on interrupt the helper pool leaks and `started` stays `true`, leaving a "started" operator whose executors are all terminated. |

### Fix only

Covered by integration tests only, or needing clock/HTTP-layer control that the
unit tests are not set up for. Each commit says which.

| Area | Bug |
|---|---|
[...]

PR #3518, opened 2026-07-30
fix: NPE in external resource caching event source when only a generic filter is set

csviri:
`acceptedByFiler` guards each of its three filter branches with
`onXFilter != null || genericFilter != null`, but the branch body
dereferences `onXFilter` unconditionally:

    if (onAddFilter != null || genericFilter != null) {
      ... .anyMatch(r -> acceptedByGenericFiler(r) && onAddFilter.accept(r));

So configuring only a generic filter via `setGenericFilter(...)` throws a
NullPointerException as soon as a resource is added, deleted or updated.
All three branches (add / delete / update) are affected, which means
`PollingEventSource`, `PerResourcePollingEventSource` and
`CachingInboundEventSource` all break when used with a generic filter
only.

The existing `genericFilteringEvents` test missed this because it uses a
filter that returns `false`: `&&` short-circuits before the null
dereference. Only a generic filter that accepts a resource reaches the
NPE.

Each filter check is now null-safe (an absent filter accepts), which
preserves the previous behaviour whenever the specific filter is set.

Adds three regression tests, one per branch; they fail with
NullPointerException without this change.

Part of #3517


csviri (commented):
LGTM

PR #3519, opened 2026-07-30
fix: NPE in handleRecentResourceUpdate for an unknown secondary resource

csviri:
`ExternalResourceCachingEventSource.handleRecentResourceUpdate` checks
that the primary has an entry in the cache, but then dereferences the
per-secondary lookup without checking it:

    R actualResource = actualValues.get(resourceId);
    if (actualResource.equals(previousVersionOfResource)) {

`actualValues.get(resourceId)` returns null whenever the primary has a
cache entry but that particular secondary id is not in it, which throws a
NullPointerException. This is reachable from
`AbstractEventSourceHolderDependentResource.onUpdated` for any
`RecentOperationCacheFiller` event source, e.g. after an update whose
resource id is not the one currently cached for that primary.

Skips the cache update when the resource is not tracked, which matches
the intent of the surrounding "only overwrite if we still hold the
version the caller saw" check.

Adds a regression test that fails with NullPointerException without this
change.

Part of #3517


csviri (commented):
LGTM

PR #3520, opened 2026-07-30
fix: honour bulk dependent resource capabilities during reconciliation

csviri:
A `BulkDependentResource` may implement any subset of `Creator`,
`Updater` and `Deleter`. `BulkDependentResourceInstance`, the internal
per-item wrapper that reuses `AbstractDependentResource`'s reconcile
logic, implements all three and delegates each call to the bulk resource
with an unchecked cast. Its capabilities therefore have to be derived
from the wrapped bulk resource.

0b8d8dd8e introduced the overridable `creatable()` / `updatable()` hooks
for exactly this, but the wiring was never completed:

* `BulkDependentResourceInstance` overrides `isCreatable()` /
  `isUpdatable()`, which `reconcile` never consults, so both overrides
  were dead code.
* the inner create branch of `AbstractDependentResource.reconcile` still
  read the `creatable` field directly instead of calling `creatable()`.

Since the wrapper implements all three interfaces, both fields are always
true, so create and update were attempted regardless of what the bulk
resource actually supports, failing with a ClassCastException in
`BulkDependentResourceInstance.create` / `update`:

    java.lang.ClassCastException: class ...CreateAndDeleteOnlyBulk
      cannot be cast to class ...Updater
      at BulkDependentResourceInstance.update(BulkDependentResourceReconciler.java:115)
      at AbstractDependentResource.handleUpdate(AbstractDependentResource.java:204)

A non-bulk dependent in the same situation just logs "implement Updater
interface to modify it" and skips the operation.

Fixes this by overriding `creatable()` / `updatable()` in the wrapper and
by using `creatable()` in the create branch. `isCreatable()` /
`isUpdatable()` now delegate to `creatable()` / `updatable()` so the two
pairs cannot drift apart again.

Adds regression tests for a create+delete-only and a delete-only bulk
dependent; both fail without this change.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/dependent/AbstractDependentResource.java
@@ -85,7 +85,7 @@ public abstract class AbstractDependentResource<R, P extends HasMetadata>
   protected ReconcileResult<R> reconcile(P primary, R actualResource, Context<P> context) {
     if (creatable() || updatable()) {
       if (actualResource == null) {
-        if (creatable) {
+        if (creatable()) {
           var desired = getOrComputeDesired(context);
           throwIfNull(desired, primary, "Desired");
           logForOperation("Creating", primary, desired);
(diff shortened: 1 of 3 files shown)


csviri:
LGTM

PR #3522, opened 2026-07-30
fix: keep a strong reference to the processing started latency gauge

csviri:
`eventProcessingStarted` registered the gauge by handing the boxed result
of `getRuntimeMXBean().getUptime()` straight to `registry.gauge(...)`:

    registry.gauge(PROCESSING_STARTED_LATENCY_GAUGE, tags,
        ManagementFactory.getRuntimeMXBean().getUptime());

Micrometer only keeps a weak reference to the gauged object, so it is the
caller's job to hold a strong reference. Nothing here does, which means
the boxed Long is collectable immediately and the gauge reports NaN.
`processing.started.latency` therefore never carried a usable value.

`controllerRegistered` already handles this correctly by keeping its
`AtomicInteger`s in the `gauges` map; this method just did not follow the
same pattern.

Registers the gauge against an `AtomicLong` held in a `longGauges` map
(an `AtomicInteger` would overflow, uptime is in milliseconds) and
updates that holder on subsequent calls, which also stops a duplicate
gauge from being registered every time the event processor starts.

Applies the same fix to the deprecated `MicrometerMetrics`, which has the
identical problem.

Adds regression tests asserting the gauge keeps a non-NaN value across a
GC and that repeated calls reuse a single gauge; the first fails without
this change. Also adds the (already dependency-managed) mockito-core test
dependency to the module, which had no way to stub a Controller before.

Part of #3517

### DIFF
--- a/micrometer-support/pom.xml
+++ b/micrometer-support/pom.xml
@@ -52,6 +52,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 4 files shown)


csviri:
As far I can see, this is not needed for this case.

PR #3523, opened 2026-07-30
fix: create the PollingEventSource timer on start and make it a daemon

csviri:
`PollingEventSource` held its timer in a final field initialised at
construction:

    private final Timer timer = new Timer();

Two problems follow from that.

The event source cannot be restarted. `stop()` calls `timer.cancel()` and
a cancelled `java.util.Timer` cannot be reused, so a subsequent `start()`
fails with `IllegalStateException: Timer already cancelled`. Restart is a
supported lifecycle - `Operator.stop()` / `start()` recreates the thread
pools for exactly this reason, and `TimerEventSource` creates a new
`Timer` inside `start()`.

The timer thread is not a daemon and is created eagerly. Merely
constructing a `PollingEventSource` therefore starts a non-daemon thread
that keeps the JVM from exiting, even if the event source is never
started, and it outlives an operator that is stopped without stopping its
event sources. `TimerEventSource` uses `new Timer(true)`.

The timer is now created in `start()` as a daemon and cleared in `stop()`,
matching `TimerEventSource`.

Adds regression tests for restart and for the daemon flag; the restart one
fails with `IllegalStateException: Timer already cancelled` without this
change.

Note: `PerResourcePollingEventSource` has a related restart limitation
because `stop()` calls `shutdownNow()` on the `ScheduledExecutorService`
from its configuration. That executor is supplied by the caller, so
changing its ownership semantics is a separate discussion and is left
out of this change.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/polling/PollingEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/polling/PollingEventSource.java
@@ -75,11 +75,12 @@ public class PollingEventSource<R, P extends HasMetadata, ID>
 
   @Override
   public void start() throws OperatorException {
+    if (timer != null) {
+      return;
+    }
     super.start();
-    // a cancelled Timer cannot be reused, so a fresh one is created on every start; it is a daemon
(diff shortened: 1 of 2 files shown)


csviri (commented):
LGTM

PR #3524, opened 2026-07-30
fix: read the external resource cache under the event source monitor

csviri:
`ExternalResourceCachingEventSource` mutates its cache from `synchronized`
methods (`handleResources`, `handleDelete`,
`handleRecentResourceCreate/Update`), but the read paths were not
synchronized. The outer map is a `ConcurrentHashMap`; the nested
per-primary maps are plain `HashMap`s that `handleDelete` mutates in
place, so a reconciler thread reading them while a poll or informer
thread writes can observe a corrupted map or throw.

`getSecondaryResources(ResourceID)` additionally looked the primary up
twice:

    var cachedValues = cache.get(primaryID);
    if (cachedValues == null) { return Collections.emptySet(); }
    else { return new HashSet<>(cache.get(primaryID).values()); }

If a concurrent `handleDelete` removes the entry between the two calls,
the second `get` returns null and this throws a NullPointerException.

Adds a `cachedResourcesFor` helper that snapshots the cached resources
while holding the monitor, and routes `getSecondaryResources` plus the
`PerResourcePollingEventSource` and `CachingInboundEventSource` overrides
(and `checkAndRegisterTask`) through it. The helper only copies, so the
potentially slow `ResourceFetcher` calls in those overrides still run
outside the lock and cannot block the informer or poll threads.

`getCache()` still returns a live view for backwards compatibility, but
now documents that iterating the nested maps requires synchronizing on the
event source.

Adds a test asserting `getSecondaryResources` returns a snapshot rather
than a live view.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ExternalResourceCachingEventSource.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/ExternalResourceCachingEventSource.java
@@ -250,12 +250,23 @@ public abstract class ExternalResourceCachingEventSource<R, P extends HasMetadat
   }
 
   public Set<R> getSecondaryResources(ResourceID primaryID) {
+    return cachedResourcesFor(primaryID);
+  }
+
+  /**
+   * Snapshot of the currently cached secondary resources for the target primary. The per-primary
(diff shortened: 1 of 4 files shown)


csviri (commented):
LGTM

xstefank (approved):

PR #3525, opened 2026-07-30
fix: complete the submission when the primary is missing from cache at execution time

csviri:
`ReconcilerExecutor.run` re-reads the primary from the cache and bails out
if it is gone. On the non-triggerOnAllEvents path that bail-out just
returned:

    } else {
      log.debug("Skipping execution; primary resource missing from cache");
      return;
    }

Two consequences.

The resource stays wedged. `submitReconciliationExecution` has already
called `state.setUnderProcessing(true)` and `state.unMarkEventReceived()`,
so returning without `eventProcessingFinished` leaves the state marked as
under processing with no event pending. `isControllerUnderExecution` then
suppresses every future submission for that resource. Usually a DELETED
event drops the whole `ResourceState` and hides this, but if the resource
leaves the informer cache without a delete event reaching the processor
(a narrowed label selector, `changeNamespaces` dropping a namespace, a
re-list whose synthetic delete is filtered by an `onDeleteFilter`) the
resource is never reconciled again.

Metrics drift. The `return` is inside the `try`, so the `finally` still
reported `reconciliationFinished` even though `reconciliationStarted` was
never reached. With `MicrometerMetricsV2` that decrements
`reconciliations.active` without a matching increment, and leaves the
`reconciliations.queue` increment from `reconciliationSubmitted`
outstanding, so both gauges drift monotonically. The same applied to the
"delete event received meanwhile" branch.

Now completes the submission via `eventProcessingFinished` (as the
sibling branch already did) and only reports `reconciliationFinished` when
`reconciliationStarted` was reported.

Adds regression tests for both; they fail without this change.

Note: `reconciliations.queue` is incremented by `reconciliationSubmitted`
but only decremented by `reconciliationStarted`, so a submission that is
abandoned still leaves the queue gauge one too high. Pairing those
properly needs a decision on the `Metrics` contract (decrement in
`reconciliationFinished`, or add an explicit "submission abandoned"
callback) and is deliberately not changed here.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
@@ -543,6 +543,7 @@ public class EventProcessor<P extends HasMetadata> implements EventHandler, Life
       // change thread name for easier debugging
       final var thread = Thread.currentThread();
       final var name = thread.getName();
+      boolean reconciliationStarted = false;
       try {
         // we try to get the most up-to-date resource from cache
         var actualResource = cache.get(resourceID);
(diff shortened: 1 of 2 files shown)


csviri (commented):
LGTM

PR #3526, opened 2026-07-30
fix: always release executors and reset state when stopping the executor manager

csviri:
`ExecutorServiceManager.stop` had three problems, all on the interrupted
path or affecting the scheduled executor.

`scheduledExecutorService` was never shut down. It is created on every
`start()` and exposed through a public accessor, but `stop()` only shut
down the reconcile, workflow and caching executors, so the pool (and any
non-daemon threads a caller created through the accessor) outlived the
operator and leaked again on every restart.

The helper pool leaked when interrupted. `Executors.newFixedThreadPool(3)`
was created inside the `try` and only shut down on the success path, so an
`InterruptedException` from `invokeAll` left three non-daemon threads
behind - in a shutdown path, where they then keep the JVM alive.

`started` was not reset when interrupted. It was only set to false on the
success path, so after an interrupted `stop()` the executors were already
shut down but `start()` would see `started == true` and do nothing. The
operator then looked started while every `execute` on the terminated
reconcile executor failed with `RejectedExecutionException`.

Moves the cleanup into a `finally`, includes the scheduled executor in the
graceful shutdown, and clears both nullable references there.

Adds regression tests for the scheduled executor shutdown and for
restartability; the former fails without this change.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ExecutorServiceManager.java
@@ -161,7 +161,6 @@ public class ExecutorServiceManager {
       // silently leaving the manager with already terminated executors
       parallelExec.shutdownNow();
       workflowExecutor = null;
-      scheduledExecutorService = null;
       started = false;
     }
   }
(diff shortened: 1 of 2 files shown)


csviri (commented):
LGTM

xstefank (approved):

PR #3527, opened 2026-07-30
fix: do not eagerly evaluate the cluster-scoped fallback in KubernetesResourceFetcher

csviri:
`fetchResource` selected between a namespaced and a cluster-scoped lookup
with `Optional.orElse`:

    return resourceId.getNamespace()
        .map(ns -> client.resources(rClass).inNamespace(ns).withName(name).get())
        .orElse(client.resources(rClass).withName(name).get());

`orElse` evaluates its argument unconditionally, so every namespaced
lookup also performed the cluster-scoped GET. For a namespaced resource
type that second request either queries whichever namespace the client
happens to default to or fails outright, and its result is then thrown
away.

This runs on the `BoundedItemStore` cache-miss path
(`refreshMissingStateFromServer`), so it doubles the API server requests
for every bounded-cache refresh of a namespaced resource.

Switches to `orElseGet` so the fallback is only evaluated for
cluster-scoped resource ids.

Part of #3517


csviri (commented):
LGTM

PR #3528, opened 2026-07-30
fix: use the cached desired state when selecting an external secondary resource

csviri:
`AbstractExternalDependentResource.selectTargetSecondaryResource` called
`desired(primary, context)` directly instead of going through
`getOrComputeDesired(context)`.

`getOrComputeDesired` exists precisely so that `desired` is invoked at
most once per reconciliation; its javadoc states that the SDK should call
it exclusively and that `desired` should never be called directly, because
this "supports scenarios where idempotent computation of the desired state
is not feasible".

Calling `desired` directly here means it is computed an extra time on
every reconciliation of an external dependent resource: once to pick the
target secondary and again from `match`/`reconcile` via the cache. That is
wasted work in all cases, and produces an inconsistent desired state for
any implementation whose `desired` is not idempotent - exactly the case
the cache is there to support.

`KubernetesDependentResource` already routes through `getOrComputeDesired`
via `targetSecondaryResourceID`; this brings the external variant in line.

No unit test is added: the code path is only covered by the external
dependent integration tests under `operator-framework`, which need a
cluster.

Part of #3517


csviri (commented):
LGTM

PR #3530, opened 2026-07-30
fix: do not NPE when adding a finalizer to a resource deleted mid-retry

csviri:
`conflictRetryingPatchPrimary` / `conflictRetryingPatch` re-read the
resource from the API server after a 409 or 422 and then re-evaluate the
precondition:

    resource = operation.inNamespace(ns).withName(name).get();

`get()` returns null if the resource was deleted in the meantime, so the
next iteration calls the precondition with null. `removeFinalizer`
anticipates this:

    r -> {
      if (r == null) {
        log.warn("Cannot remove finalizer since resource not exists.");
        return false;
      }
      return r.hasFinalizer(finalizerName);
    }

but `addFinalizer` passes `r -> !r.hasFinalizer(finalizerName)`, which
throws a NullPointerException instead of exiting cleanly.

Gives `addFinalizer` the same null guard, in both `ResourceOperations` and
the deprecated `PrimaryUpdateAndCacheUtils`.

No test is added: reaching the retry path requires stubbing the client to
answer 409/422 and then 404 through the whole fabric8 DSL chain, which the
existing unit tests are not set up for.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/PrimaryUpdateAndCacheUtils.java
@@ -291,7 +291,7 @@ public class PrimaryUpdateAndCacheUtils {
         },
         r -> {
           if (r == null) {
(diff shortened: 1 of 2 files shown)


csviri (commented):
LGTM

PR #3531, opened 2026-07-30
fix: use the null-guarded metrics field in the Controller constructor

csviri:
The constructor deliberately guards against a null `Metrics`:

    this.metrics = Optional.ofNullable(configurationService.getMetrics()).orElse(Metrics.NOOP);

but its last statement bypasses that guard and calls the getter again:

    configurationService.getMetrics().controllerRegistered(this);

so a `ConfigurationService` whose `getMetrics()` returns null fails with a
NullPointerException during controller construction, which is exactly what
the guard three lines up is meant to prevent. `EventProcessor` applies the
same `!= null ? metrics : Metrics.NOOP` guard, so the codebase does treat
a null `Metrics` as reachable even though the interface default returns
`Metrics.NOOP`.

Uses the already-resolved `metrics` field instead.

Part of #3517


csviri (commented):
LGTM

PR #3532, opened 2026-07-30
fix: remove dangling placeholder from event source close warning

csviri:
`log.warn("Error closing {} -> {}", eventSource.name(), e)` supplies two
placeholders but only one non-throwable argument. SLF4J trims a trailing
`Throwable` and reports it as the exception, so the stack trace is logged
correctly, but the second placeholder is left unfilled and the message
renders with a literal brace pair:

    Error closing myEventSource -> {}

Drops the redundant placeholder and names what is being logged.

Note: the superficially similar calls in `InformerManager.stop` and
`InformerWrapper.start` are correct as they stand - they pass more
non-throwable arguments than placeholders, so nothing is left dangling.

Part of #3517


csviri (commented):
LGTM

PR #3533, opened 2026-07-30
fix: assorted robustness and clarity fixes

csviri:
Three unrelated small items found while auditing.

ManagedInformerEventSource.get could throw instead of falling back to the
informer cache. It compared a temporary-cache hit against
`manager().lastSyncResourceVersion(namespace)` without guarding either
failure mode:

* `lastSyncResourceVersion` is null until an informer has completed its
  initial list, and `compareResourceVersions` dereferences it, so this
  threw a NullPointerException.
* `InformerManager.lastSyncResourceVersion` does `getSource(ns)
  .orElseThrow()`, which throws `NoSuchElementException` for a namespace
  that is no longer watched after a dynamic namespace change - a window
  that only closes on the next re-list, when `checkGhostResources` prunes
  the entry.

`TemporaryResourceCache.putResource` already guards the null case; the
comparison is now extracted into a helper that guards both and falls back
to the informer cache, which is the safe answer in either case.

ExpectationResult.name() threw a bare NullPointerException when no
expectation was registered, which is a normal result state produced by
`ExpectationManager.checkExpectation`. It now throws
`IllegalStateException` pointing at `isExpectationPresent()`, and says so
in the javadoc.

ResourceOperations cleanups: drops the `options` parameter of
`desiredForJsonPatch`, which was never read at any of its six call sites,
and rewrites the comment on `create` - it claimed the operation "check if
the resource already exists", which it does not; the actual reason
filtering is safe is that the API server rejects a duplicate create.

Part of #3517

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/reconciler/ResourceOperations.java
@@ -406,7 +406,9 @@ public class ResourceOperations<P extends HasMetadata> {
    * @return the created resource as returned by the API server
    */
   public <R extends HasMetadata> R create(R resource) {
-    // it is safe to do event filtering for create since check if the resource already exists.
+    // filtering the own event is safe here without optimistic locking or a matcher: a create either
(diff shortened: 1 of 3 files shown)


csviri (commented):
LGTM

PR #3543, opened 2026-08-07
test: fix flaky finalizer removal in TriggerReconcilerOnAllEventIT

csviri:
The event count is increased at the beginning of the reconciliation, thus
waiting for it released the test into the middle of a reconciliation that
was still about to remove the finalizer. The subsequent update then raced
with that removal and failed with a conflict.

Wait for the finalizer removal to actually land, and retry the (optimistically
locked) update with a fresh read on conflict.


csviri (commented):
LGTM

PR #3547, opened 2026-08-07
perf: size the workflow result map from Workflow#size

csviri:
AbstractWorkflowExecutor called Workflow#getDependentResourcesByName purely to
read its size. That allocates a HashMap and walks every node to collect the
dependent resources, then discards the map. The executor is constructed on
every reconcile and cleanup of a workflow-based reconciler, so use the existing
Workflow#size instead.

---


csviri (commented):
LGTM

PR #3550, opened 2026-08-07
refactor: drop the unused Options parameter of desiredForJsonPatch

csviri:
ResourceOperations#desiredForJsonPatch takes an Options argument that it never
reads - it only clones the actual resource and applies the operator. All six
jsonPatch call sites thread the value in for nothing, which suggests the method
still honours the option. Remove the parameter.

---


csviri (commented):
LGTM

PR #3554, opened 2026-08-11
test: sample for virtual threads

csviri:
Adds sample (serves as a smoke test) for virtual threads setup of JOSDK.

TODO:
 - check logging / thread local handling

### DIFF
--- a/docs/content/en/docs/documentation/operations/configuration.md
+++ b/docs/content/en/docs/documentation/operations/configuration.md
@@ -23,6 +23,47 @@ Operator operator = new Operator( override -> override
         .withLeaderElectionConfiguration(new LeaderElectionConfiguration("bar", "barNS")));
 ```
 
+## Running on Virtual Threads
+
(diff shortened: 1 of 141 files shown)


docs/content/en/docs/documentation/operations/configuration.md
```diff
@@ -23,6 +23,47 @@ Operator operator = new Operator( override -> override
         .withLeaderElectionConfiguration(new LeaderElectionConfiguration("bar", "barNS")));
 ```
 
+## Running on Virtual Threads
+
+Reconciliations and workflow steps are executed on the executors provided by
+`ConfigurationService`, which are fixed size thread pools by default. On Java 21 and newer, these
+can be replaced with virtual threads, so that a reconciliation waiting on a remote call does not
+occupy a platform thread:
```
xstefank, line 31:
```suggestion
can be replaced with virtual threads. 
```

Each virtual thread always runs on platform thread in the background.

PR #3558, opened 2026-08-12
test: cover informer retry after a CR deserialization problem

### LINKED ISSUE #2688: Add Integration Test for Informer Deserialization issues
In case a CRD is outdated, deserialization cannot happen, see details here:
https://github.com/operator-framework/java-operator-sdk/blob/main/adr/002-Custom-Resource-Deserialization-Problem.md

But the retry seems to be not happening as logged here:
https://github.com/operator-framework/java-operator-sdk/blob/main/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/InformerWrapper.java#L97

We should add an IntegrationTest to check if this is still retried and the actual problem can be fixed in runtime.
This might be changed in fabric8 client.

csviri:
Adds an IT for the situation of ADR 002 where the resource that cannot be
deserialized is already present when the informer lists on startup, the
counterpart of MultiVersionCRDIT, which covers the watch case. With
stopOnInformerErrorDuringStartup set to false the operator starts, but the
informer is not retried: the exception handler installed by the informer pool
declines a retry for deserialization errors and the Reflector of the fabric8
client then completes its stop future, so fixing the problem in the cluster
while the operator runs does not bring the informer back.

The test asserts that current behavior, so it fails once the informer does get
retried; the log of the informer pool no longer promises a periodic retry in
this case.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/AbstractInformerPool.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/source/informer/pool/AbstractInformerPool.java
@@ -170,6 +170,15 @@ public abstract class AbstractInformerPool implements InformerPool {
       if (configurationService.stopOnInformerErrorDuringStartup()) {
         log.error("Informer startup error. Operator will be stopped. Informer: {}", informer, e);
         throw new OperatorException(e);
+      } else if (ExceptionHandler.isDeserializationException(e)) {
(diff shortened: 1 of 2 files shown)


csviri:
This correctly tests the current behavior, although, visiting the topic again, I'm not entirely convinced, if this is the right thing to do, so not to retry on deserialization error. But will stick with this for now, and will re-visit on demand.

xstefank (approved):

PR #3577, opened 2026-08-27
improve: patch for event recorder count to use resource from context

csviri:
This avoids reading the resource again from cluster

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event updates to detect conflicts when multiple operations attempt to modify the same event concurrently.
  * Prevented stale event data from overwriting newer changes.


csviri:
@TQJADE sorry this was an afterthought, it should be a bit more efficient this way.

PR #3582, opened 2026-09-02
fix: ensure release tag points to non-snapshot release commit

### LINKED ISSUE #2304: Improve Release - Proper Tagging
We discovered that during the release the tag that we create contains wrong version (the actual not the target release version), this does not make any functional problems, the code is the proper one, just the project version is set after GitHub Action receives the event, but the tag is already created that time.

Lavanya-N24:
Fixes #2304

### Motivation
GitHub Releases previously pointed to commits containing `-SNAPSHOT` in `pom.xml` because GitHub creates the tag at release publication time before CI can update the version.

### Changes
- **`release.yml`**: Updated branch resolution to strip `-tmp` suffixes so release-branch matching continues to work when releasing via temporary tags.
- **`release-project-in-dir.yml`**:
  - Added `permissions: contents: write` to allow tagging and committing.
  - Updated `pom.xml` to the release version and committed the change to the target branch.
  - Added a defensive check (`git diff --quiet`) to prevent empty commit failures.
  - Created the final release tag on the clean commit and pointed the GitHub Release to it using `gh release edit`.
  - Deleted the temporary `-tmp` tag.

### Testing
Verified via an isolated proof-of-concept repository:
https://github.com/Lavanya-N24/release-tagging-poc

## Summary by CodeRabbit

- **Release Improvements**
  - Improved automated releases for projects located in subdirectories.
  - Release tags are parsed consistently and passed through the release process.
  - Temporary release tags are converted into final version tags pointing to the release commit.
  - Release version updates are automatically committed and pushed.
  - Updated release automation improves reliability and Java environment compatibility.
  - Release workflows now provide clearer handling of release tags and branch-selection failures.


.github/workflows/release-project-in-dir.yml
```diff
@@ -11,20 +11,35 @@ on:
...
 
 jobs:
   publish:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - name: Checkout "${{inputs.version_branch}}" branch
-        uses: actions/checkout@v7
+        uses: actions/checkout@v4
```
csviri, line 27: I guess this is a mistake, v7 should be the version we want to use

.github/workflows/release-project-in-dir.yml
```diff
@@ -11,20 +11,35 @@ on:
...
+          STRIPPED="${RAW_TAG#v}"
+          FINAL_VERSION="${STRIPPED%-tmp}"
+          FINAL_TAG="v${FINAL_VERSION}"
+
+          echo "raw_tag=${RAW_TAG}" >> "$GITHUB_OUTPUT"
+          echo "final_version=${FINAL_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "final_tag=${FINAL_TAG}" >> "$GITHUB_OUTPUT"
 
       - name: Set up Java and Maven
-        uses: actions/setup-java@v6
```
csviri, line 27: Same here this version should not be changed

.github/workflows/release-project-in-dir.yml
```diff
@@ -11,20 +11,35 @@ on:
...
         with:
           ref: "${{inputs.version_branch}}"
+          fetch-depth: 0
+
+      - name: Parse release version
+        id: vars
+        run: |
+          RAW_TAG="${{ github.event.release.tag_name }}"
+          STRIPPED="${RAW_TAG#v}"
+          FINAL_VERSION="${STRIPPED%-tmp}"
```
csviri: This -tmp removal is not needed we can assume that tag will be like `v5.6.0` (for next minor release version)

.github/workflows/release.yml
```diff
@@ -55,16 +58,22 @@ jobs:
...
     needs: prepare-release
+    permissions:
+      contents: write
     uses: ./.github/workflows/release-project-in-dir.yml
     secrets: inherit
     with:
-      version_branch: ${{needs.prepare-release.outputs.version_branch}}
-      project_dir: '.'
+      version_branch: ${{ needs.prepare-release.outputs.version_branch }}
+      release_tag: ${{ needs.prepare-release.outputs.release_tag }}
```
csviri, line 117: We don't declare and use this version in the sub-flow, can we adjust this for that?

.github/workflows/release-project-in-dir.yml
csviri: I belive we don't need the -tmp suffix, we can create the tag like `v5.6.0` and than override that tag. That would be more straightforward during the release so we don't have to remember special naming convention.

```
  git tag -f -a "${RAW_TAG}" -m "Release ${RAW_TAG}"                                                                                                                                                                                               
  git push -f origin "refs/tags/${RAW_TAG}"  
```

.github/workflows/release-project-in-dir.yml
```diff
@@ -36,11 +51,39 @@ jobs:
           gpg-passphrase: MAVEN_GPG_PASSPHRASE
 
       - name: Change version to release version
-        # Assume that RELEASE_VERSION will have form like: "v1.0.1". So we cut the "v"
         run: |
-          ./mvnw ${MAVEN_ARGS} versions:set -DnewVersion="${RELEASE_VERSION:1}" versions:commit -DprocessAllModules
+          ./mvnw ${MAVEN_ARGS} versions:set -DnewVersion="${{ steps.vars.outputs.final_version }}" versions:commit -DprocessAllModules
+
+      - name: Commit and push release version
```
csviri, line 87: Can we have these 2 steps after maven release? that way if maven release fails, and we can easily retry.

.github/workflows/release-project-in-dir.yml
```diff
@@ -36,11 +51,39 @@ jobs:
...
+      - name: Re-tag commit and update GitHub Release
         env:
-          RELEASE_VERSION: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          RAW_TAG="${{ steps.vars.outputs.raw_tag }}"
+          FINAL_TAG="${{ steps.vars.outputs.final_tag }}"
+
+          if [ "$RAW_TAG" != "$FINAL_TAG" ]; then
```
csviri: If we don't have the -temp suffix, we don't have to check this, we simply override the tag as mentioned before, then we don;t have to edit the release, since release just point to a ref name on GitHub

.github/workflows/release.yml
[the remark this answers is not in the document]
csviri: This is valid, see also my coment below

csviri (changes requested):
@Lavanya-N24 added some comments

Lavanya-N24:
Thanks @csviri! I've simplified the workflow per your suggestions:

- Removed the `-tmp` suffix handling entirely; release tags (e.g. `v5.6.0`) are created normally and overridden directly.
- Moved the release commit and tag override steps after `Publish to Apache Maven Central` so deployment failures remain cleanly retryable.
- Overridden the tag using `git tag -f -a "${RELEASE_TAG}"` and `git push -f origin "refs/tags/${RELEASE_TAG}"`, eliminating the need for GitHub release editing.
- Preserved action versions (`actions/checkout@v4`, `actions/setup-java@v4`) and declared `release_tag` in the sub-workflow inputs.

Lavanya-N24:
Hi @csviri, I have updated the branch with your requested changes:
- Removed the `-tmp` suffix logic and switched to direct tag overriding (`git tag -fa` and `git push -f origin refs/tags/...`).
- Moved the release commit and tag override steps to run after Maven Central deploy.
- Reverted the setup actions back to `@v4`.
- Pinned `ad-m/github-push-action` to its immutable commit SHA per security checks.

.github/workflows/release-project-in-dir.yml
csviri: @Lavanya-N24 I still see the v4 here and other places

Lavanya-N24:
Hi @csviri,can you please review once again

.github/workflows/release-project-in-dir.yml
csviri: @Lavanya-N24 these checkout action and other action versions are still wrong, should not be changed

csviri (changes requested):

Lavanya-N24:
Hi @csviri, I've just double-checked the live file via "View file" (not the diff) and confirmed all actions/checkout and actions/setup-java steps are on @v4, and github-push-action is pinned to the SHA — no v7/v6/@master remain anywhere in the current commit. Could you please re-review against the latest commit? Let me know if you're still seeing something different on your end.

.github/workflows/release-project-in-dir.yml
Lavanya-N24: Hi @csviri, done — I moved these 2 steps (commit/push release version + override release tag) out of the publish job into a separate finalize-release job that has needs: publish. So now:

- publish only handles the Maven Central deploy
- finalize-release only runs after publish succeeds, and does the git commit/push + tag override

If the Maven release fails, finalize-release never runs and nothing gets committed/tagged, so we can safely retry publish on its own. Let me know if this matches what you had in mind.

.github/workflows/release.yml
Lavanya-N24: Fixed — release_tag is now declared as a required input in release-project-in-dir.yml and passed through from release.yml (needs.prepare-release.outputs.release_tag), and it's consumed in the version-setting and tagging steps. Let me know if this matches what you had in mind.

csviri:
https://github.com/Lavanya-N24/java-operator-sdk/blob/4aded2f86e0654918317875a6082080ee7ffad8f/.github/workflows/release-project-in-dir.yml#L36

Pls check here the version changes what I mean.

Lavanya-N24:
Hi @csviri, I restored the action versions as requested. actions/checkout is now v7 and actions/setup-java is v6 in the publish and update-working-version jobs. I left the finalize-release versions unchanged.

csviri:
@Lavanya-N24 thank you, will take a look early next week.

csviri:
Hi @Lavanya-N24 could you pls squash you commits and rebase it to the main pls?

Lavanya-N24:
Hi @csviri, done — squashed all commits into one and rebased onto the latest main. Let me know if anything else needs changes.

csviri (approved):
LGTM, thank you @Lavanya-N24

csviri (approved):
LGTM, thank you @Lavanya-N24

PR #3603, opened 2026-09-07
feat: virtual thread support

### LINKED ISSUE #3538: Support for Virtual Threads
Fabric8 client should already support virtual threads, we should add a sample, in case improve our configuration API to support it more nicely. 

Note that thread pools are already customizable, so might work just also now with virtual thread based thread pools.

Virtual threads will probably lead to more efficient threading and higher throughput.

csviri:
Introduces feature flag to use virtual threads instead of standard `ThreadPools`. 
Thus, basically only replaces internal the thread pools, with ones that use virtual threads.  
The concurrency limit also applies for this new mode (using `Semaphore`).
Note that for now we aim for Java 25, and the intention is that only that will be officially supported, since `synchronized` keyword is used extensively in the codebase.

TODO:
- check the client and verify threads set up correnctly

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
@@ -20,7 +20,6 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.function.Consumer;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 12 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/VirtualThreads.java
```diff
@@ -0,0 +1,186 @@
...
+   * the submitting thread, just like queuing it on a fixed size platform thread pool wouldn't.
+   */
+  private static final class BoundedExecutorService extends AbstractExecutorService {
+
+    private final ExecutorService delegate;
+    private final Semaphore permits;
+
+    private BoundedExecutorService(ExecutorService delegate, int maxConcurrency) {
+      this.delegate = delegate;
+      // fair, so that tasks run roughly in submission order as they would on a thread pool
```
xstefank: this seems like a piece of agent communication rather than useful comment

xstefank (approved):
Approving because I know that I will be probably only one against a global virtual thread flag here.

But maybe you want to wait for our next JUG session :) - https://www.meetup.com/brno-java-meetup/events/316579260/?eventOrigin=group_upcoming_events

csviri:
> Approving because I know that I will be probably only one against a global virtual thread flag here.
> 
> But maybe you want to wait for our next JUG session :) - https://www.meetup.com/brno-java-meetup/events/316579260/?eventOrigin=group_upcoming_events

What would you propose as an alternative?

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/VirtualThreads.java
csviri: fixed, thx

gyfora (approved):

operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/config/ConfigurationService.java
```diff
@@ -163,13 +168,17 @@ public <R extends HasMetadata> R clone(R object) {
...
-        .build();
+    final var builder =
+        new KubernetesClientBuilder()
+            .withConfig(
+                new ConfigBuilder(Config.autoConfigure(null))
+                    .withMaxConcurrentRequests(DEFAULT_MAX_CONCURRENT_REQUEST)
+                    .build())
+            .withKubernetesSerialization(new KubernetesSerialization());
+    if (VirtualThreads.shouldUse(useVirtualThreads())) {
+      builder.withTaskExecutorSupplier(VirtualThreads.newKubernetesClientTaskExecutorSupplier());
```
csviri, line 179: @shawkins @manusa I would like to ask for you opinion, we would like to support virtual threads (java 25+) here, and was thinking that if we do we could also use an executor for the client. Does this seems to be a good idea? was the client tested with this approach? Also in general, what do you think? 

thank you!

csviri:
cc @shawkins @manusa if you have the bandwidth please take a look on this PR, thank you!

PR #3604, opened 2026-09-07
feat: configurable aggregation, owner references and naming for the event recorder

### LINKED ISSUE #3601: EventRecorder: default aggregation key, ownerReference support, and naming control

The new `EventRecorder` is exactly the thing we've been hand-rolling in our operator for years, so 5.6.0 was the trigger to try to finally delete our own implementation. We got close, but three things stopped us. Raising them now while the API is still experimental, in case you're open to shaping it a bit further.

1. **The message is part of the event identity by default.** Our health events carry live detail in the message (current pod states and so on), so every message change becomes a new Event object. For this kind of "current status" event that's not great: old states hang around as separate rows in `kubectl get events` until the TTL cleans them up, and during an incident the object count grows with every distinct message. Now, `EventRecord.key()` already solves this nicely, we tried it on a real cluster and with a key set the repeats patch `count`, `lastTimestamp` and `message` in place, giving the classic `(x12 over 3m)` in `kubectl describe`. The catch is that `key()` has to be set on every single record. We have around 40 emission sites, and one forgotten `key()` silently falls back to one-object-per-message.

2. **No ownerReference.** There's no way to tie a recorded event to the resource it's about, so after someone deletes the CR its events stick around until the TTL expires.

3. **No control over the event name.** Names are always `<resource>.<hash>`. Our current events have predictable names (`<resource>-<event-name>`) and it turns out people actually rely on that, we found scripts doing `kubectl get event <name>`. So switching to the recorder as-is would break them.

1. A default aggregation-key strategy at recorder or controller level, something like "aggregate by reason", so `context.eventRecorder().normal(reason, message)` just does the right thing without per-call setup. Per-record `key()` stays as the override.
2. An opt-in flag to set the ownerReference to the regarding object.
3. Let the record (or recorder config) supply the event name or a naming strategy, hash stays the default. A per-reason name gives you the same guarantees the hash provides (no collisions, stable across restarts and replicas), just readable.

Happy to contribute any of these if you like the direction. Also happy to split this into three issues if that works better for you.

Keeping our own implementation, which is what we're doing for now. It works, but it's ~100 lines plus tests that do what `EventRecorder` almost does, and we'd rather be on the SDK. A wrapper that injects `key()` everywhere would only fix the first point and adds a layer we don't want to own.

afalhambra-hivemq:
The commits are self-contained, so feel free to review or cherry-pick them separately.

1. **Default aggregation key**: new `EventKeyStrategy` and a `DefaultEventRecorder.builder(sink)` with a `keyStrategy(...)` option. With `byReason()` the message drops out of the event identity, so status-style events aggregate onto one event (count grows, message gets replaced) without `key()` boilerplate at every call site. A per-record `key()` still wins, and `none()` is today's behavior.
2. **Owner references**: `ownerReference(true)` on the builder, with a per-record override via `EventRecord.Builder.ownedByRegarding(boolean)`. Only set when the object already has a uid.
3. **Naming**: `EventRecord.Builder.name(String)` and a recorder-level `EventNamingStrategy`. Precedence: record name, then strategy, then the identity hash. The name is the aggregation identity and must be a valid RFC 1123 DNS subdomain (the javadoc has the details), long names get truncated at 253 chars, and blank means unset.

Fixes #3601

## Summary by CodeRabbit

* **New Features**
  * Event recording can now be customized with naming and aggregation strategies.
  * Events support explicit names, validation, length limits, and safe fallback naming.
  * Events can aggregate by reason, combining repeated occurrences while tracking counts.
  * Recorded events can optionally include an owner reference to the related resource.
  * Event records support custom names and owner-reference preferences.
  * Added a builder for configuring event recorder behavior.
* **Tests**
  * Added coverage for custom naming, aggregation, name validation, strategy failures, and owner references.

### DIFF
--- a/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/event/EventNamingStrategy.java
+++ b/operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/event/EventNamingStrategy.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright Java Operator SDK Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.javaoperatorsdk.operator.api.event;
+
+import java.util.Optional;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.javaoperatorsdk.operator.api.reconciler.Experimental;
+
+import static io.javaoperatorsdk.operator.api.reconciler.Experimental.API_MIGHT_CHANGE;
+
+/**
+ * Names the event recorded about an object. The name is what the sink looks a recorded event up by,
+ * so it is also the aggregation identity: two records resolving to the same name are counted as
+ * occurrences of one event. A name must therefore be unique among the events it should not
+ * aggregate with, and stable across operator restarts and replicas.
+ *
+ * <p>A name must be a valid RFC 1123 DNS subdomain: at most 253 lowercase alphanumeric characters,
+ * {@code -} or {@code .}, starting and ending with an alphanumeric character. Names longer than the
+ * limit are truncated. A name derived from the object and a fixed lowercase suffix (such as {@code
+ * <object>-status-report}) satisfies all of this by construction.
+ *
+ * <p>An empty result or an invalid name falls back to the default {@code <object>.<identity hash>}
+ * name, rather than the event being lost to the API server rejecting the name.
+ *
+ * <p>Implementations are called from concurrent reconciliations and must be thread safe.
+ */
+@Experimental(API_MIGHT_CHANGE)
+@FunctionalInterface
+public interface EventNamingStrategy {
+
+  Optional<String> nameFor(HasMetadata regarding, EventRecord record);
+
+  /** No custom naming: every event gets the default {@code <object>.<identity hash>} name. */
+  static EventNamingStrategy none() {
+    return (regarding, record) -> Optional.empty();
+  }
+}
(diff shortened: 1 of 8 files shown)


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/api/event/EventNamingStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>An empty result or an invalid name falls back to the default {@code <object>.<identity hash>}
+ * name, rather than the event being lost to the API server rejecting the name.
+ *
+ * <p>Implementations are called from concurrent reconciliations and must be thread safe.
+ */
+@Experimental(API_MIGHT_CHANGE)
+@FunctionalInterface
+public interface EventNamingStrategy {
+
+  Optional<String> nameFor(HasMetadata regarding, EventRecord record);
```
csviri, line 45: kind of a nit but:

Wonder if it would be better just use `String` as result type here. And instead of `none()` allow `null` values in builder. That would require null check. But I think in terms of behavior, if a naming strategy is set, it should always produce a name; and the `Optional<String>` kinda implies the oposite.

Same for `KeyStrategy`.

What do you think @afalhambra-hivemq ?  @xstefank @gaborgsomogyi

csviri: otherwise looks great!

afalhambra-hivemq: The `Optional` is on purpose: it lets a strategy skip events. Empty = "no opinion, fall through to the default", so one controller can name/key its status-style reasons and leave the rest on the hash identity. For example, the new `ConfiguredEventRecorderIT` naming strategy does exactly that, and since the hash naming is internal, falling through is the only way to get it back.

So I think the question here is: may a strategy decline, or must it always produce? I can see the codebase has precedent for both: For example `RetryExecution.nextDelay()` returns Optional (empty = no more retries), while `Creator.create()` returns plain `R` because it can't decline. 
Optional here follows the first pattern; String + nullable builder arg fits the second.

Happy either way. One thing to settle if we go with String approach: what if a strategy returns `null` anyway? The recorder must never fail, so it'd have to quietly fall back to the hash, which sneaks partiality back in through the back door.

csviri: > So I think the question here is: may a strategy decline, or must it always produce? 

Yep, for some reason, the always produce (non nullable) result, is my first intuition. But I'm ok also with the current appraoch.
`RetryExecution.nextDelay()` is Optional, to indicate that we do not want to schedule a next retry.

csviri (approved):
LGTM, thank you @afalhambra-hivemq

csviri:
@afalhambra-hivemq just one more details, could you target pls `next` branch with the PR, that contains the work for next minor version, and since this is not a bug fix, it fits rather there.

csviri:
thank you @afalhambra-hivemq !

PR #3611, opened 2026-09-10
test: deflake ComplexWorkflowIT

csviri:
The test timed out waiting for the workflow to report READY. Its StatefulSets
ran mongo:4.4 with a 1Gi volumeClaimTemplate, so the ready postcondition
(readyReplicas > 0) depended on a ~400MB Docker Hub pull plus PVC provisioning
completing inside the 120s budget. In the failing CI run the first StatefulSet's
status never changed after creation, so `second` never unblocked and the status
stayed NOT_READY.

Swap the workload for registry.k8s.io/pause:3.9 and drop the volume claim
template. The pod has no readiness probe, so it is ready as soon as it runs, and
the image is ~700KB from a registry without anonymous pull rate limits.

Also remove the orphaned copies of these templates under
resources/io/javaoperatorsdk/operator/sample/complexdependent, left behind when
the test classes moved to the workflow package; nothing referenced them.

## Summary by CodeRabbit

- **Tests**
  - Updated workflow fixtures to use generic StatefulSet and Service configurations, including port 80 and a pause-container workload.
  - Removed MongoDB-specific persistence, storage mounts, database ports, and related configuration from the fixtures.
  - Removed obsolete sample MongoDB Service and StatefulSet manifests.
  - Stopped adding the `K8S_NAME` label to the first volume claim template in the test StatefulSet construction.


gaborgsomogyi (approved):

xstefank (commented):
LGTM

PR #3613, opened 2026-09-11
feat: config property to set watched namespaces

csviri:
Adds `josdk.controller.<name>.namespaces` to ConfigLoader, which was
previously the only way of configuring watched namespaces that was not
available from an external configuration source.

The value is a comma-separated list; entries are trimmed and blank ones
ignored. The special values JOSDK_ALL_NAMESPACES and JOSDK_WATCH_CURRENT
are supported but, as everywhere else, only on their own. Validation
happens while building the consumer, so an invalid value fails at wiring
time rather than at reconciliation time.

## Summary by CodeRabbit

* **New Features**
  * Configure watched namespaces through the `josdk.controller.<name>.namespaces` property.
  * Support individual or comma-separated namespaces, with whitespace trimmed and blank entries ignored.
  * Choose between watching all namespaces or only the operator’s current namespace.
  * Namespace settings replace annotation-based defaults and remain adjustable at runtime.
  * Invalid combinations of special values and namespace names are rejected.

* **Documentation**
  * Added guidance on supported values, validation rules, and configuration behavior.

### DIFF
--- a/operator-framework/src/main/java/io/javaoperatorsdk/operator/config/loader/ConfigLoader.java
+++ b/operator-framework/src/main/java/io/javaoperatorsdk/operator/config/loader/ConfigLoader.java
@@ -16,9 +16,12 @@
 package io.javaoperatorsdk.operator.config.loader;
 
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -27,6 +30,7 @@ import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.javaoperatorsdk.operator.api.config.ConfigurationServiceOverrider;
 import io.javaoperatorsdk.operator.api.config.ControllerConfigurationOverrider;
 import io.javaoperatorsdk.operator.api.config.LeaderElectionConfigurationBuilder;
+import io.javaoperatorsdk.operator.api.config.informer.InformerConfiguration;
 import io.javaoperatorsdk.operator.config.loader.provider.AggregatePriorityListConfigProvider;
 import io.javaoperatorsdk.operator.config.loader.provider.EnvVarConfigProvider;
 import io.javaoperatorsdk.operator.config.loader.provider.PropertiesConfigProvider;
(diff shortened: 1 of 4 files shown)


operator-framework/src/main/java/io/javaoperatorsdk/operator/config/loader/ConfigLoader.java
```diff
@@ -297,6 +313,45 @@ Consumer<ControllerConfigurationOverrider<R>> buildRateLimiterConsumer(String pr
...
+   * can only be used on their own. Returns {@code null} when the property is not present.
+   *
+   * @throws IllegalArgumentException if the property is present but does not resolve to a valid set
+   *     of namespaces
+   */
+  private <R extends HasMetadata>
+      Consumer<ControllerConfigurationOverrider<R>> buildNamespacesConsumer(String prefix) {
+    final var key = prefix + NAMESPACES_SUFFIX;
+    final var value = configProvider.getValue(key, String.class);
+    if (value.isEmpty()) {
```
gaborgsomogyi, line 331: Maybe trim as early return?

operator-framework/src/test/java/io/javaoperatorsdk/operator/config/loader/ConfigLoaderTest.java
```diff
@@ -558,6 +560,113 @@ void retryIsIsolatedPerControllerName() {
...
+  // -- watched namespaces -----------------------------------------------------
+
+  private static Set<String> applyAndGetNamespaces(
+      java.util.function.Consumer<
+              io.javaoperatorsdk.operator.api.config.ControllerConfigurationOverrider<
+                  io.fabric8.kubernetes.api.model.ConfigMap>>
+          consumer) {
+    var overrider =
+        io.javaoperatorsdk.operator.api.config.ControllerConfigurationOverrider.override(
+            baseControllerConfig());
```
gaborgsomogyi: import is not possible?

gaborgsomogyi (approved):

operator-framework/src/test/java/io/javaoperatorsdk/operator/config/loader/ConfigLoaderTest.java
csviri: fixed, thx!

operator-framework/src/main/java/io/javaoperatorsdk/operator/config/loader/ConfigLoader.java
csviri: This is an Optional, but added isBlank check after. Thx!!

PR #3614, opened 2026-09-11
test: deflake additionalEventDuringRetryOnDeleteEvent

csviri:
The test released the blocked reconciler right after the update call returned,
but the update event still had to travel back through the informer. When the
reconciliation failed before the event was registered, the framework treated
the failure as a plain retry (consuming the last attempt) instead of instantly
re-triggering because of a superseding event, so only 4 instead of 5
reconciliations happened.

Wait for isNextReconciliationImminent() in the reconciler, which reflects
exactly the state the framework checks after the reconciliation fails and
cannot be unset while it is in progress. Also fix the isWaiting() assertion
that had no terminal assertion and was therefore a no-op.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of pending reconciliation events after resource deletion when processing all events, ensuring reconciliation timing reflects superseding events.
  - Kept retry behavior consistent while using the unified pending-reconciliation check.

- **Documentation**
  - Clarified when the next reconciliation is scheduled, including after successful or failed reconciliations and events received after deletion.

- **Tests**
  - Strengthened integration tests to verify that reconciliation waits for superseding events during retry scenarios.


operator-framework-core/src/main/java/io/javaoperatorsdk/operator/processing/event/EventProcessor.java
```diff
@@ -511,7 +509,19 @@ public synchronized void start() throws OperatorException {
...
+  }
+
+  /**
+   * An event that arrives after a delete event is tracked in a dedicated state, so {@link
+   * ResourceState#eventPresent()} alone does not cover it. Such an event triggers a new
+   * reconciliation right after the current one, both when it succeeds (see {@link
+   * #eventProcessingFinished}) and when it fails (see {@link #handleRetryOnException}), so it has
+   * to be reported as imminent too.
+   */
+  private boolean nextReconciliationImminent(ResourceState state) {
```
gaborgsomogyi: Why not called `isNextReconciliationImminent`? I mean it does the same just like the previous function but on different class, right?

csviri: fixed, thank you!

gaborgsomogyi (approved):

Issue #3624, opened 2026-09-17
comparableResourceVersions is dropped when a dependent resource's informer configuration is copied


afalhambra-hivemq:
`InformerEventSourceConfiguration.Builder.updateFrom` copies 11 fields from the informer configuration but not `comparableResourceVersions`, so it falls back to the default.

`KubernetesDependentResource.createEventSource` is its only caller, which means setting it on a dependent resource has no effect:

```java
@KubernetesDependent(informer = @Informer(comparableResourceVersions = false))
```

It is parsed correctly by `InformerConfiguration.Builder.initFromAnnotation`, ends up in the dependent's configuration, and is then reset to `true` at that hand off. No warning, the setting just does not apply.

Looks like a one line fix, adding `.withComparableResourceVersions(informerConfig.isComparableResourceVersions())` to the chain. Happy to open a PR if you want.

Noticed while working on #3615, where the same method needed a new field added to it.

csviri:
PR merged, thank you!

PR #3625, opened 2026-09-17
fix: keep comparableResourceVersions when copying a dependent resource's informer configuration

afalhambra-hivemq:
`updateFrom` wasn't copying `comparableResourceVersions`, so it fell back to the default and setting it on a dependent resource had no effect. One line, plus a test that fails without it.

Worth noting it's a real behaviour change for anyone who set it to false on a dependent: they now actually get the temporary resource cache disabled, so read-your-own-write is off for that secondary. That's what they asked for, but they haven't been getting it.

Fixes #3624


csviri (approved):
LGTM, thank you!

